PR #4, opened 2015-10-06
Disable tracing when making calls to tcollector

prashantv:
Add methods to ContextBuilder to disable tracing, and use those in the ZipKin trace reporter.

r @junchaowu

### DIFF
--- a/context_builder.go
+++ b/context_builder.go
@@ -38,6 +38,9 @@ type ContextBuilder struct {
 	// CallOptions are TChannel call options for the specific call.
 	CallOptions *CallOptions
 
+	// TracingDisabled disables trace reporting for calls using this context.
+	TracingDisabled bool
+
 	// Hidden fields: we do not want users outside of tchannel to set these.
 	incomingCall IncomingCall
 	span         *Span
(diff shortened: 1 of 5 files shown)


trace/zipin_tracereporter_test.go
```diff
@@ -35,12 +36,49 @@ import (
...
+	ch.Peers().Add(tcollectorCh.PeerInfo().HostPort)
+
+	called := make(chan int)
+	ret := &gen.Response{Ok: true}
+	mockServer.On("Submit", mock.Anything, mock.Anything).Return(ret, nil).Run(func(args mock.Arguments) {
+		called <- 1
+	})
+
+	// Make some calls, and validate that the trace reporter is not called recursively.
+	require.NoError(t, json.Register(tcollectorCh, json.Handlers{"op": func(ctx json.Context, arg map[string]interface{}) (map[string]interface{}, error) {
```
junchaowu: why do you use json

prashantv: Was the simplest to set up an inline server to listen on (just needed any listener)

context_builder.go
```diff
@@ -119,6 +129,10 @@ func (cb *ContextBuilder) Build() (ContextWithHeaders, context.CancelFunc) {
 		timeout = defaultTimeout
 	}
 
+	if cb.TracingDisabled {
+		cb.span.EnableTracing(false)
```
junchaowu: this function look weird. should be enableTracing / disableTracing

prashantv: Yeah looks weird, but it's a single function to set true/false. It's not really worth breaking the public API for it, especially not as part of this change.

prashantv:
Merging to unblock other changes.

PR #7, opened 2015-10-07
Fix typo in test filename: zipin_ -> zipkin

prashantv:
cc @junchaowu


junchaowu:
do we have this naming issue. damn. (shipit)

PR #9, opened 2015-10-09
Adding a post handle callback to the thrift server for resource cleanup

martin-mao:


prashantv:
Created a PR without using a second map:
https://github.com/uber/tchannel-go/pull/14

PR #14, opened 2015-10-13
Add a post-response callback to Thrift server

prashantv:
Uses the Option pattern (http://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html).

This allows us to avoid breaking backwards compatibility, and avoiding another method.

Few things to be careful about:
 - The response struct passed is NOT the same as what is returned by the server. It's the "Result" struct created by Thrift which also contains exceptions etc.
 - The method name does not have the service name prefix.

cc @martin-mao


martin-mao:
LGTM

PR #24, opened 2015-10-16
Optionally, allow SubChannels to isolate their PeerLists

akshayjshah:
(Discussed with @prashantv last night.)

This PR allows subchannels to (optionally) maintain a separate peer list from the parent channel without creating duplicate connections to the same host. It accomplishes this by building an inheritance chain among peer lists such that parent lists are the union of their children's lists.

@prashantv is planning to add some extra tests to make sure that this approach works well with graceful shutdown. I'll leave it up to him whether the tests added to `channel_test.go` are necessary - they test a subset of the cases that `subchannel_test.go` explores, but they do so through a different interface.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -441,7 +441,7 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string, connectionOptio
 // incomingConnectionActive adds a new active connection to our peer list.
 func (ch *Channel) incomingConnectionActive(c *Connection) {
 	c.log.Debugf("Add connection as an active peer for %v", c.remotePeerInfo.HostPort)
-	p := ch.peers.GetOrAdd(c.remotePeerInfo.HostPort)
+	p := ch.rootPeers().GetOrAdd(c.remotePeerInfo.HostPort)
 	p.AddConnection(c)
 
 	ch.mutable.mut.Lock()
--- a/channel.go
+++ b/channel.go
@@ -153,7 +153,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		handlers:          &handlerMap{},
 		subChannels:       &subChannelMap{},
 	}
-	ch.peers = newPeerList(ch)
+	ch.peers = newPeerList(ch).newChild()
 
 	ch.mutable.peerInfo = LocalPeerInfo{
 		PeerInfo: PeerInfo{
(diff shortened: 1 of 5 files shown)


channel.go
```diff
@@ -287,6 +296,11 @@ func (ch *Channel) Peers() *PeerList {
 	return ch.peers
 }
 
+// rootPeers returns the root PeerList for the channel.
```
prashantv: can you add a comment about what rootPeers is here

channel.go
```diff
@@ -427,7 +441,7 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string, connectionOptio
 // incomingConnectionActive adds a new active connection to our peer list.
 func (ch *Channel) incomingConnectionActive(c *Connection) {
 	c.log.Debugf("Add connection as an active peer for %v", c.remotePeerInfo.HostPort)
-	p := ch.peers.GetOrAdd(c.remotePeerInfo.HostPort)
+	p := ch.rootPeers().GetOrAdd(c.remotePeerInfo.HostPort)
```
prashantv: So one behaviour we need to support is:
 - client adds one peer: "hyperbahn-1:21300", and then advertises with Hyperbahn
 - Hyperbahn then connects back to us using 10 connections
 - When client makes calls to Hyperbahn, it should be able to use those 10 connections (without having to manually add peers)

When we get an incoming connection, we need to wait for it to get identified (which is when we find out what service name it is), and probably add it to the right subchannel's peer lists based on the service name.

Let's add a test for that, and see if it works the way we expect.

channel_test.go
```diff
@@ -25,6 +25,8 @@ import (
 	"os"
 	"testing"
 
+	"golang.org/x/net/context"
```
prashantv: nit: golang should go after github imports

peer.go
```diff
@@ -38,24 +38,57 @@ var (
 	peerRng = NewRand(time.Now().UnixNano())
 )
 
+type Connectable interface {
```
prashantv: optionally: if we can break this into a separate commit, we can PR it and submit it separately, since I think there's no behaviour changes as part of it right?

prashantv:
Thanks for making these changes @akshayjshah, the tests are great. I thought of a few more corner cases that we can work through in person.

channel.go
akshayjshah: Done.

channel.go
akshayjshah: Discussed in person. Unfortunately, TChannel doesn't send enough information with the init request to tell which service the connection comes from, so we can't do this. Added a TODO reflecting this.

channel_test.go
akshayjshah: :see_no_evil:

PR #25, opened 2015-10-17
Optionally, allow SubChannels to isolate their PeerLists

akshayjshah:
(Discussed with @prashantv last night.)

This PR allows subchannels to (optionally) maintain a separate peer list from the parent channel without creating duplicate connections to the same host. It accomplishes this by building an inheritance chain among peer lists such that parent lists are the union of their children's lists.


prashantv:
lgtm, thanks @akshayjshah 

feel free to merge

benfleis:
awesome @akshayjshah! glad to see somebody coming through with this :)

PR #27, opened 2015-10-21
Use @generated instead of Autogenerated so Phabricator suppresses diffs

prashantv:
Also add comments in the generated code so go vet stops complaining about the generated code.

cc @martin-mao


martin-mao:
Nice! LGTM!

junchaowu:
do we have to manually add @generated in the future if we generate new code. Can we add it in thrift-go

prashantv:
Yes, when we have our thrift-go compiler we should do the same.

PR #29, opened 2015-10-22
Separate peer connections into inbound and outbound connections

junchaowu:
This patch updates the introspection state as well.
@prashantv

### DIFF
--- a/incoming_test.go
+++ b/incoming_test.go
@@ -42,7 +42,7 @@ func TestPeersIncomingConnection(t *testing.T) {
 	newService := func(svcName string) (*Channel, string) {
 		ch, err := testutils.NewClient(&testutils.ChannelOpts{ServiceName: svcName})
 		require.NoError(t, err, "NewClient failed")
-		require.NoError(t, ch.ListenAndServe(":0"), "ListenAndServe failed")
+		require.NoError(t, ch.ListenAndServe("127.0.0.1:0"), "ListenAndServe failed")
 		channels = append(channels, ch)
 		return ch, ch.PeerInfo().HostPort
 	}
@@ -72,6 +72,20 @@ func TestPeersIncomingConnection(t *testing.T) {
 			_, err := sc.Peers().Get()
 			assert.Equal(t, ErrNoPeers, err,
 				"incoming connections should not be added to non-root peer list")
+		}
+
+		// verify number of peers/connections on the client side
+		opts := &IntrospectionOptions{}
+		serverState := ch.IntrospectState(opts).RootPeers
+		for _, client := range []*Channel{ringpop, hyperbahn} {
+			clientPeerState := client.IntrospectState(opts).RootPeers
+			assert.Equal(t, len(clientPeerState), 1)
+			assert.Equal(t, len(clientPeerState[ch.PeerInfo().HostPort].OutboundConnections), 1)
+			assert.Equal(t, len(clientPeerState[ch.PeerInfo().HostPort].InboundConnections), 0)
+
+			assert.Equal(t, len(serverState), 2)
+			assert.Equal(t, len(serverState[client.PeerInfo().HostPort].InboundConnections), 1)
+			assert.Equal(t, len(serverState[client.PeerInfo().HostPort].OutboundConnections), 0)
 		}
 
 		// In future when connections send a service name, we should be able to
--- a/incoming_test.go
+++ b/incoming_test.go
@@ -79,16 +79,16 @@ func TestPeersIncomingConnection(t *testing.T) {
 		serverState := ch.IntrospectState(opts).RootPeers
 		serverHostPort := ch.PeerInfo().HostPort
 
-		assert.Equal(t, len(serverState), 2, "Incorrect peer number")
+		assert.Equal(t, len(serverState), 2, "Incorrect peer count")
 		for _, client := range []*Channel{ringpop, hyperbahn} {
 			clientPeerState := client.IntrospectState(opts).RootPeers
 			clientHostPort := client.PeerInfo().HostPort
-			assert.Equal(t, len(clientPeerState), 1, "Incorrect peer number")
-			assert.Equal(t, len(clientPeerState[serverHostPort].OutboundConnections), 1, "Incorrect outbound connection number")
-			assert.Equal(t, len(clientPeerState[serverHostPort].InboundConnections), 0, "Incorrect inbound connection number")
+			assert.Equal(t, len(clientPeerState), 1, "Incorrect peer count")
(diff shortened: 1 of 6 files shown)


incoming_test.go
```diff
@@ -74,6 +74,20 @@ func TestPeersIncomingConnection(t *testing.T) {
 				"incoming connections should not be added to non-root peer list")
 		}
 
+		// verify number of peers/connections on the client side
+		opts := &IntrospectionOptions{}
+		serverState := ch.IntrospectState(opts).RootPeers
+		for _, client := range []*Channel{ringpop, hyperbahn} {
+			clientPeerState := client.IntrospectState(opts).RootPeers
+			assert.Equal(t, len(clientPeerState), 1)
```
prashantv: Add messages to all the `assert.Equal` calls

incoming_test.go
```diff
@@ -74,6 +74,20 @@ func TestPeersIncomingConnection(t *testing.T) {
...
+			assert.Equal(t, len(clientPeerState[ch.PeerInfo().HostPort].OutboundConnections), 1)
+			assert.Equal(t, len(clientPeerState[ch.PeerInfo().HostPort].InboundConnections), 0)
+
+			assert.Equal(t, len(serverState), 2)
```
prashantv: You can move this assert to outside of the loop (before or after is OK)

incoming_test.go
```diff
@@ -74,6 +74,20 @@ func TestPeersIncomingConnection(t *testing.T) {
...
+			assert.Equal(t, len(serverState[client.PeerInfo().HostPort].InboundConnections), 1)
```
prashantv: nit: make a `clientHostPort` variable (it looks too similar to `ch.PeerInfo().HostPort` when it's so long). Do the same above, and maybe call that one `serverHostPort := ch.PeerInfo().HostPort`

introspection.go
```diff
@@ -113,7 +114,7 @@ func (l *PeerList) IntrospectState(opts *IntrospectionOptions) map[string]PeerRu
 	l.mut.RLock()
 	for _, peer := range l.peers {
 		peerState := peer.IntrospectState(opts)
-		if len(peerState.Connections) > 0 || opts.IncludeEmptyPeers {
+		if len(peerState.InboundConnections) > 0 || len(peerState.OutboundConnections) > 0 || opts.IncludeEmptyPeers {
```
prashantv:
```go
if len(peerState.InboundConnections) + len(peerState.OutboundConnections) > 0 || opts.IncludeEmptyPeers {
```
(just a little bit shorter and easier to understand)

introspection.go
```diff
@@ -145,15 +146,23 @@ func (p *Peer) IntrospectState(opts *IntrospectionOptions) PeerRuntimeState {
 	p.mut.RLock()
 
 	hostPort := p.hostPort
-	conns := make([]ConnectionRuntimeState, len(p.connections))
-	for i, conn := range p.connections {
-		conns[i] = conn.IntrospectState(opts)
+	inboundConns := make([]ConnectionRuntimeState, len(p.inboundConnections))
```
prashantv: nit: make a local function to do this for you.

```go
func getConnectionRuntimeState(conns []Connections) []ConnectionRuntimeState
```

And then you can just do 
```
inboundConns := getConnectionRuntimeState(p.inboundConnections)
outboundConns := getConnectionRuntimeState(p.outboundConnections)
```

introspection.go
```diff
@@ -92,8 +92,9 @@ type ExchangeRuntimeState struct {
 
 // PeerRuntimeState is the runtime state for a single peer.
 type PeerRuntimeState struct {
-	HostPort    string                   `json:"hostPort"`
-	Connections []ConnectionRuntimeState `json:"connections"`
+	HostPort            string                   `json:"hostPort"`
+	OutboundConnections []ConnectionRuntimeState `json:"outbound connections"`
```
prashantv: json tag is the field name in the json struct, so you shouldn't have spaces.
use
```go
`json:"outboundConnections"`
`json:"inboundConnections"`
```

peer.go
```diff
@@ -219,9 +220,9 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 	return c, nil
 }
 
-// AddConnection adds an active connection to the peer's connection list.
+// AddInboundConnection adds an active inbound connection to the peer's connection list.
 // If a connection is not active, ErrInvalidConnectionState will be returned.
-func (p *Peer) AddConnection(c *Connection) error {
+func (p *Peer) AddInboundConnection(c *Connection) error {
 	switch c.readState() {
 	case connectionActive, connectionStartClose:
```
prashantv: can you add the following TODO for me right after this case:
```go
// TODO(prashantv): Block inbound connections when the connection is not active.
```

utils_for_test.go
```diff
@@ -46,9 +46,9 @@ func OutboundConnection(call *OutboundCall) (*Connection, net.Conn) {
 func GetConnections(ch *Channel) []*Connection {
 	var connections []*Connection
```
prashantv: i just merged a change into master earlier today which removes this function :) can you rebase off latest master, then you won't need any change here.

prashantv:
Looks pretty good, just some minor nits.

introspection.go
junchaowu: I updated all other nits except this one. I don't think it is necessary. The exist code just one line and this is only place to use it. I would keep it as it is.

prashantv: It's not just one line - the function should create the array, and then fill that array with the results of the introspect call. It's not just one line, it's 4 lines of duplication.

utils_for_test.go
```diff
@@ -42,6 +42,17 @@ func OutboundConnection(call *OutboundCall) (*Connection, net.Conn) {
 	return conn, conn.conn
 }
 
+// GetConnections returns all connections for a channel.
```
prashantv: You can delete this function (it no longer exists in master, so no need to update it)

incoming_test.go
```diff
@@ -74,6 +74,23 @@ func TestPeersIncomingConnection(t *testing.T) {
 				"incoming connections should not be added to non-root peer list")
 		}
 
+		// verify number of peers/connections on the client side
+		opts := &IntrospectionOptions{}
+		serverState := ch.IntrospectState(opts).RootPeers
+		serverHostPort := ch.PeerInfo().HostPort
+
+		assert.Equal(t, len(serverState), 2, "Incorrect peer number")
```
prashantv: nit: in all the messages, replace number with count, "Incorrect peer count" "Incorrect outbound connection count" etc.

utils_for_test.go
junchaowu: I am surprised. After rebase it is still here.

prashantv:
lgtm

PR #32, opened 2015-10-22
Add peer selection mechanism

junchaowu:
create peerSelector struct which will choose peers based on different strategy.

### DIFF
--- a/peer_strategies.go
+++ b/peer_strategies.go
@@ -30,13 +30,24 @@ type scoreCalculator interface {
 }
 
 type randCalculator struct {
-	peerRng *rand.Rand
+	rng *rand.Rand
 }
 
 func (r *randCalculator) GetScore(p *Peer) float32 {
-	return peerRng.Float32()
+	return r.rng.Float32()
 }
 
 func newRandCalculator() *randCalculator {
-	return &randCalculator{peerRng: NewRand(time.Now().UnixNano())}
+	return &randCalculator{rng: NewRand(time.Now().UnixNano())}
+}
+
+type preferIncomingCalculator struct {
+}
+
+func newPreferIncomingCalculator() *preferIncomingCalculator {
+	return &preferIncomingCalculator{}
+}
+
+func (r *preferIncomingCalculator) GetScore(p *Peer) float32 {
+	return 0
 }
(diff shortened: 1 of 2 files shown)


peer_strategies.go
```diff
@@ -0,0 +1,57 @@
...
+import "math/rand"
+
+type scoreCalculator interface {
+	GetScore(p *Peer) float32
+}
+
+type randCalculator struct{}
+
+func (r randCalculator) GetScore(p *Peer) float32 {
+	newRand := rand.New(rand.NewSource(99))
```
prashantv: don't create a new rand every time, instead store the `rand` in the `randCalculator` struct, and you should have a `newRandCalculator` function that will create the struct with the new `rand`. take a look at `peerRng`- you'll want to init yours the same way.

peer_strategies.go
```diff
@@ -0,0 +1,57 @@
...
+	return newRand.Float32()
+}
+
+type peerSelector struct {
```
prashantv: do we need a separate struct for this? should this just live inside of `PeerList`? And just have the calculator be a field on peerlist?

peer_strategies.go
```diff
@@ -0,0 +1,42 @@
...
+	"math/rand"
+	"time"
+)
+
+type scoreCalculator interface {
+	GetScore(p *Peer) float32
+}
+
+type randCalculator struct {
+	peerRng *rand.Rand
```
prashantv: call just call this `rng` since it's localized to the struct

peer_strategies.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (r *randCalculator) GetScore(p *Peer) float32 {
+	return peerRng.Float32()
```
prashantv: dealing with floats is less efficient than dealing with ints normally, i think ints might be better

peer_strategies.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func newRandCalculator() *randCalculator {
```
prashantv: move the constructor function to be under the struct definition

prashantv:
lgtm

PR #33, opened 2015-10-23
Separate root peer list from peer list

prashantv:
Prefactor to simplify peer selection using a heap.

r @junchaowu


junchaowu:
:ship:

PR #35, opened 2015-10-26
Implement peer heap for peer selection

junchaowu:
Change peer heap to min heap.
Add randon method to comparator method
Add test for peer heap.

### DIFF
--- a/peer_heap.go
+++ b/peer_heap.go
@@ -20,39 +20,54 @@
 
 package tchannel
 
-import "container/heap"
+import (
+	"container/heap"
+	"math/rand"
+	"time"
+)
 
-// PeerHeap maintains a MAX heap of peers based on the peers' score.
-type PeerHeap []*peerScore
+// PeerHeap maintains a MIN heap of peers based on the peers' score.
+type PeerHeap struct {
+	PeerScores []*peerScore
+	rng        *rand.Rand
+}
 
-func (ph PeerHeap) Len() int { return len(ph) }
+func newPeerHeap() *PeerHeap {
+	return &PeerHeap{PeerScores: nil, rng: NewRand(time.Now().UnixNano())}
+}
+
+func (ph PeerHeap) Len() int { return len(ph.PeerScores) }
 
 func (ph PeerHeap) Less(i, j int) bool {
-	// We want Pop to give us the highest, not lowest, score so we use greater than here.
-	return ph[i].score > ph[j].score
+	// We want Pop to give us the lowest, not highest, score so we use less than here.
+	if ph.PeerScores[i].score == ph.PeerScores[j].score {
+		// return random true or false when scores are the same.
+		return ph.rng.Intn(2) < 1
+	}
+	return ph.PeerScores[i].score < ph.PeerScores[j].score
 }
 
 func (ph PeerHeap) Swap(i, j int) {
-	ph[i], ph[j] = ph[j], ph[i]
-	ph[i].index = i
-	ph[j].index = j
+	ph.PeerScores[i], ph.PeerScores[j] = ph.PeerScores[j], ph.PeerScores[i]
+	ph.PeerScores[i].index = i
+	ph.PeerScores[j].index = j
 }
 
 // Push implements heap Push interface
 func (ph *PeerHeap) Push(x interface{}) {
-	n := len(*ph)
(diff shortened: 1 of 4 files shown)


peer_heap.go
```diff
@@ -20,39 +20,54 @@
...
 
-func (ph PeerHeap) Len() int { return len(ph) }
+func (ph PeerHeap) Len() int { return len(ph.PeerScores) }
 
 func (ph PeerHeap) Less(i, j int) bool {
-	// We want Pop to give us the highest, not lowest, score so we use greater than here.
-	return ph[i].score > ph[j].score
+	// We want Pop to give us the lowest, not highest, score so we use less than here.
+	if ph.PeerScores[i].score == ph.PeerScores[j].score {
+		// return random true or false when scores are the same.
```
prashantv: put a comment about why we're doing this

peer_heap.go
```diff
@@ -20,39 +20,54 @@
...
-	ph[j].index = j
+	ph.PeerScores[i], ph.PeerScores[j] = ph.PeerScores[j], ph.PeerScores[i]
+	ph.PeerScores[i].index = i
+	ph.PeerScores[j].index = j
 }
 
 // Push implements heap Push interface
 func (ph *PeerHeap) Push(x interface{}) {
-	n := len(*ph)
+	n := len((*ph).PeerScores)
```
prashantv: you can just do
```go
len(ph.PeerScores)
```

Same for the below append call.

peer_heap.go
```diff
@@ -64,10 +79,10 @@ func (ph *PeerHeap) pop() *peerScore {
...
 
-func (ph *PeerHeap) push(peer *peerScore) {
-	heap.Push(ph, peer)
+func (ph *PeerHeap) push(peerScore *peerScore) {
+	heap.Push(ph, peerScore)
 }
 
 func (ph *PeerHeap) peek() *peerScore {
-	return (*ph)[0]
+	return (*ph).PeerScores[0]
```
prashantv: should be able to do
```go
return ph.PeerScores[0]
```

peer_heap_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPeerHeap(t *testing.T) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	peerHeap := newPeerHeap()
+	peerScores := make([]*peerScore, 10)
```
prashantv: create a constant in this function for the 10, and use that constant in the loop and here.

peer_heap_test.go
```diff
@@ -0,0 +1,59 @@
...
+	minScore := uint64(10)
+	for i := 0; i < 10; i++ {
+		peerScore := newPeerScore(&Peer{})
+		peerScore.score = uint64(r.Intn(10))
```
prashantv: let's make the range bigger, maybe `r.Intn(numPeer * 5)` or something

peer_heap.go
```diff
@@ -20,39 +20,54 @@
...
 // Pop implements heap Pop interface
 func (ph *PeerHeap) Pop() interface{} {
 	old := *ph
-	n := len(old)
-	item := old[n-1]
+	n := len(old.PeerScores)
+	item := old.PeerScores[n-1]
 	item.index = -1 // for safety
-	*ph = old[0 : n-1]
+	(*ph).PeerScores = old.PeerScores[0 : n-1]
```
prashantv: you can drop the indexes in the slice if they're the default, so this should become:
```go
ph.PeerScores = old.PeerScores[:n-1]
```

peer_heap_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	for i := 0; i < 10; i++ {
+		peerHeap.push(peerScores[i])
+	}
+
+	assert.Equal(t, 10, peerHeap.Len(), "Incorrect peer heap size.")
+	assert.Equal(t, minScore, peerHeap.peek().score, "PeerHeap's peek value is not minium.")
+
+	lastScore := peerHeap.pop().score
+	for i := 0; i < 9; i++ {
```
prashantv: check the length in this loop

prashantv:
lgtm

PR #36, opened 2015-10-26
Prescreen for merge heap branch into master

junchaowu:
@prashantv

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -447,6 +447,13 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string, connectionOptio
 
 // exchangeUpdated updates the peer heap.
 func (ch *Channel) exchangeUpdated(c *Connection) {
+	if c.remotePeerInfo.HostPort == "" {
+		// Hostport is unknown until we get init resp.
+		return
+	}
+
+	p := ch.rootPeers().GetOrAdd(c.remotePeerInfo.HostPort)
+	ch.subChannels.updatePeerHeap(p)
 }
 
 // incomingConnectionActive adds a new active connection to our peer list.
--- a/channel.go
+++ b/channel.go
@@ -348,6 +348,7 @@ func (ch *Channel) serve() {
 		events := connectionEvents{
 			OnActive:           ch.incomingConnectionActive,
 			OnCloseStateChange: ch.connectionCloseStateChange,
+			OnExchangeUpdated:  ch.exchangeUpdated,
 		}
 		if _, err := ch.newInboundConnection(netConn, events, &ch.connectionOptions); err != nil {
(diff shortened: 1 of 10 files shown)


channel.go
```diff
@@ -348,6 +348,7 @@ func (ch *Channel) serve() {
 		events := connectionEvents{
 			OnActive:           ch.incomingConnectionActive,
 			OnCloseStateChange: ch.connectionCloseStateChange,
+			OnExchangeUpdated:  ch.exchangenUpdated,
```
prashantv: `ch.exchangenUpdated` typo

channel.go
```diff
@@ -441,6 +445,17 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string, connectionOptio
...
 
+// exchangenUpdated updates the peer heap.
+func (ch *Channel) exchangenUpdated(c *Connection) {
+	if c.remotePeerInfo.HostPort == "" {
+		// Hostport is unknown until we get init resp.
+		return
+	}
+
+	p := ch.rootPeers().GetOrAdd(c.remotePeerInfo.HostPort)
+	ch.subChannels.updatePeerHeap(p)
```
prashantv: since `updatePeerHeap` only cares about the host:port, should we just pass host:port directly?

connection.go
```diff
@@ -247,12 +250,18 @@ func (ch *Channel) newConnection(conn net.Conn, initialState connectionState, ev
 	}
 	c.inbound.onRemoved = c.checkExchanges
 	c.outbound.onRemoved = c.checkExchanges
+	c.inbound.onAdded = c.onExchangeAdded
```
prashantv: can this direcltly call `c.callOnExchangeChange`? and then we can remove `onExchangeAdded`

connection.go
```diff
@@ -270,6 +279,15 @@ func (c *Connection) callOnCloseStateChange() {
 	}
 }
 
+func (c *Connection) callOnExchangeChange() {
+	// stack := make([]byte, 4096)
```
prashantv: delete debugging code

mex.go
```diff
@@ -177,6 +178,7 @@ func (mexset *messageExchangeSet) newExchange(ctx context.Context, framePool Fra
 
 	mexset.exchanges[mex.msgID] = mex
 	mexset.sendChRefs.Add(1)
+	mexset.onAdded()
```
prashantv: the `mexset.mut` is locked. I don't think we want to do that. If we had a peer scoring strategy that accessed the mex, it would probably dead lock at this point. We should have tests that catch this.

junchaowu:
this is depreciated now.

PR #37, opened 2015-10-26
Fix ch.exchangenUpdated typo

junchaowu:
@prashantv


channel.go
```diff
@@ -446,7 +446,7 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string, connectionOptio
 }
 
 // exchangenUpdated updates the peer heap.
```
prashantv: forgot to fix this

junchaowu: lint doesn't come up. weird.

prashantv:
lgtm

PR #38, opened 2015-10-27
Add InboundConnection test utility

prashantv:
This allows tests to get the remote peer information for a given call.

r @junchaowu


junchaowu:
lgtm

PR #48, opened 2015-11-04
Add prefer incoming strategy

prashantv:
This adds a preferIncoming score calculator which will use higher scores
for any peer with inbound connections.

This is so we use Hyperbahn's incoming connections where possible.

Pushing this now so that clients stop making too many connections to Hyperbahn.
r @junchaowu

### DIFF
--- a/peer_test.go
+++ b/peer_test.go
@@ -22,7 +22,10 @@ package tchannel_test
 
 import (
 	"fmt"
+	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	. "github.com/uber/tchannel-go"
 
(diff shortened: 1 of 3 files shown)


peer_test.go
```diff
@@ -112,3 +115,63 @@ func TestGetPeerAvoidPrevSelected(t *testing.T) {
...
+		}
+		wg.Wait()
+
+		for _, v := range selected {
+			assert.True(t, expected[v], "Peers.Get got unexpected peer: %v", v)
+		}
+	})
+
+	// Clean up allChannels
+	for _, c := range allChannels {
```
breerly: Do you end up putting these in all your tests?

prashantv: Most tests, I'll just use `defer ch.Close()` instead, and that works fine since there's only a few channels being created. This one creates a few more, so I'm closing them in a loop instead. `defer` would probably have been fine here as well though.

PR #79, opened 2015-11-14
Peer selection using peer heap

junchaowu:


mex.go
```diff
@@ -186,6 +187,7 @@ func (mexset *messageExchangeSet) newExchange(ctx context.Context, framePool Fra
 
 	mexset.exchanges[mex.msgID] = mex
 	mexset.sendChRefs.Add(1)
+	mexset.onAdded()
```
prashantv: This is called with the mexset locked, which is not right. This could cause deadlocks, since an exchange update could cause a peer score to update, but the peer score needs the incoming count, which needs a read lock on the mexset which is already locked.

Unlock the `mexset` after the `mexset.exchanges` set, and before the return earlier.

peer.go
```diff
@@ -74,81 +83,74 @@ func (l *PeerList) newSibling() *PeerList {
...
 	if p, ok := l.peersByHostPort[hostPort]; ok {
-		return p
+		return p.Peer
 	}
 
 	p := l.parent.Add(hostPort)
-	atomic.AddUint32(&p.scCount, 1)
-	l.peersByHostPort[hostPort] = p
-	l.peers = append(l.peers, p)
+	ps := newPeerScore(p)
```
prashantv: We should score the peer before we add it.

peer_heap.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// PopPeer pops the top peer of the heap.
+func (ph *PeerHeap) PopPeer() *peerScore {
+	return heap.Pop(ph).(*peerScore)
+}
+
+// PushPeer pushes the new peer into the heap.
+func (ph *PeerHeap) PushPeer(peerScore *peerScore) {
+	atomic.AddUint64(&(ph.order), 1)
```
prashantv: why do we need order? and if we need it, how is it different from the length of the heap?

Maybe this should be called `AddNewPeer` if we end up doing this order stuff here

peer.go
```diff
@@ -74,81 +83,74 @@ func (l *PeerList) newSibling() *PeerList {
...
 
-	if bestPeer != nil {
-		atomic.AddUint64(&bestPeer.chosenCount, 1)
+	if ps == nil {
+		return nil
 	}
-	return bestPeer
+	ps.score++
+	l.peerHeap.PushPeer(ps)
+	atomic.AddUint64(&(ps.chosenCount), 1)
```
prashantv: `&ps.chosenCount` should work without needing parens

peer.go
```diff
@@ -159,11 +161,50 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+	ps, ok := l.peersByHostPort[hostPort]
+	l.RUnlock()
+
+	return ps, ok
+}
+
+// UpdatePeerHeap updates the peer heap.
+func (l *PeerList) UpdatePeerHeap(p *Peer) {
+	if ps, ok := l.exists(p.hostPort); ok {
+		l.Lock()
```
prashantv: don't need to lock while scoring. You need a read lock on `ps` when reading the score but don't need to lock the whole peer list.

peer.go
```diff
@@ -159,11 +161,50 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+	l.RLock()
+	ps, ok := l.peersByHostPort[hostPort]
+	l.RUnlock()
+
+	return ps, ok
+}
+
+// UpdatePeerHeap updates the peer heap.
+func (l *PeerList) UpdatePeerHeap(p *Peer) {
+	if ps, ok := l.exists(p.hostPort); ok {
```
prashantv: remember to reduce nesting where possible.

```go
ps, ok := l.exists(p.hostPort)
if !ok {
  return
}

score := l.scoreCalculator.GetScore(p)
if score == ps.readScore() {
  // no need to update the heap
  continue
}

// update the heap with a lock.
```

peer.go
```diff
@@ -224,6 +268,14 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 	return c, nil
 }
 
+// NumInbound returns the number of inbound connections to this node.
```
prashantv: you can probably revert adding this change and deleting it below.

peer_heap.go
```diff
@@ -0,0 +1,93 @@
...
+
+func newPeerHeap() *PeerHeap {
+	return &PeerHeap{rng: NewRand(time.Now().UnixNano())}
+}
+
+func (ph PeerHeap) Len() int { return len(ph.PeerScores) }
+
+func (ph *PeerHeap) Less(i, j int) bool {
+	if ph.PeerScores[i].score == ph.PeerScores[j].score {
+		return ph.PeerScores[i].order < ph.PeerScores[j].order
```
prashantv: this introduces harmonics -- if you have 200 go clients, they will all have the same hosts in the same order, and will probably make requests to the same nodes in order.

when you make one request at a time, is this enough to avoid selecting the same nodes repeatedly?

peer_strategies.go
```diff
@@ -20,40 +20,49 @@
...
-		rn += math.MaxUint32
+type preferIncomingCalculator struct{}
+
+func (preferIncomingCalculator) GetScore(p *Peer) uint64 {
+	if p.NumInbound() <= 0 {
+		return math.MaxUint64
+	}
+	count := 0
+	for _, c := range p.outboundConnections {
+		count = count + len(c.outbound.exchanges)
```
prashantv: accessing `c.outbound.exchanges` without read locks here and below. run with -race to see if these are detected

peer_test.go
```diff
@@ -22,17 +22,26 @@ package tchannel_test
 
 import (
 	"fmt"
+
 	"runtime"
+
```
prashantv: why are your imports all over the place here?

subchannel.go
```diff
@@ -160,3 +160,11 @@ func (subChMap *subChannelMap) getOrAdd(serviceName string, ch *Channel) *SubCha
 
 	return subChMap.registerNewSubChannel(serviceName, ch)
 }
+
+func (subChMap *subChannelMap) updatePeerHeap(p *Peer) {
+	for _, subCh := range subChMap.subchannels {
```
prashantv: accessing `subchannels` without read lock

peer.go
junchaowu: Right

peer_test.go
junchaowu: what do u mean? this is handled by go-import,

prashantv: There's blank lines between standard library imports which is not right. goimports might be misconfigured or something, not sure but you should check and make sure that the import groups are correct.

Make this import look like:
```go
import (
  "fmt"
  "runtime"
  "sort"
  "sync"
  [... and other standard library imports]

  [all other imports without any blank lines]
)
```

peer_heap.go
junchaowu: we discussed this offline.

connection.go
```diff
@@ -251,12 +254,18 @@ func (ch *Channel) newConnection(conn net.Conn, initialState connectionState, ev
 	c.log = log
 	c.inbound.onRemoved = c.checkExchanges
 	c.outbound.onRemoved = c.checkExchanges
+	c.inbound.onAdded = c.onExchangeAdded
```
prashantv: should we remove the intermediate `onExchangeAdded` and instead directly make this call `callOnExchangeChange`?

mex.go
```diff
@@ -180,12 +179,16 @@ func (mexset *messageExchangeSet) newExchange(ctx context.Context, framePool Fra
 			mexset.log.Warnf("msg id %d used for both active mex %s and new mex %s",
 				mex.msgID, existingMex.msgType, mex.msgType)
 		}
+		mexset.mut.Unlock()
```
prashantv: nit: order this so it looks like:
```go
if [...] {
  [...]
} else {
  [...]
}

mexset.mut.Unlock()
return nil, errDuplicateMex
```

I want the `Unlock` as close to the return

peer.go
```diff
@@ -74,81 +83,76 @@ func (l *PeerList) newSibling() *PeerList {
...
-		return p
+		return p.Peer
 	}
 
 	p := l.parent.Add(hostPort)
-	atomic.AddUint32(&p.scCount, 1)
-	l.peersByHostPort[hostPort] = p
-	l.peers = append(l.peers, p)
+	ps := newPeerScore(p)
+	ps.score = l.scoreCalculator.GetScore(p)
```
prashantv: put the score calculation directly in side of `newPeerScore` rather than giving it some initial score and replacing it immediately

peer.go
```diff
@@ -74,81 +83,76 @@ func (l *PeerList) newSibling() *PeerList {
...
 }
 
 // Get returns a peer from the peer list, or nil if none can be found.
-// Get will avoid peers in prevSelected unless there are no other peers left.
 func (l *PeerList) Get(prevSelected map[string]struct{}) (*Peer, error) {
-	l.RLock()
-
-	if len(l.peers) == 0 {
-		l.RUnlock()
+	l.Lock()
```
prashantv: do we have any numbers that tell us the impact of having an exclusive lock on this compared to a read lock earlier?

peer.go
```diff
@@ -74,81 +83,76 @@ func (l *PeerList) newSibling() *PeerList {
...
 	}
 
-	if bestPeer != nil {
-		atomic.AddUint64(&bestPeer.chosenCount, 1)
+	if ps == nil {
+		return nil
 	}
-	return bestPeer
+	ps.Lock()
+	ps.score++
```
prashantv: We should probably break this logic up -- seems like this should belong in the score calculator

peer.go
```diff
@@ -159,11 +163,62 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+	l.RLock()
+	ps, ok := l.peersByHostPort[hostPort]
+	l.RUnlock()
+
+	return ps, ok
+}
+
+// UpdatePeerHeap updates the peer heap.
+func (l *PeerList) UpdatePeerHeap(p *Peer) {
+
```
prashantv: nit: delete blank line

peer.go
```diff
@@ -159,11 +163,62 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+	ps, ok := l.exists(p.hostPort)
+	if !ok {
+		return
+	}
+
+	score := l.scoreCalculator.GetScore(p)
+	ps.RLock()
```
prashantv: rather than using read locks everywhere, add a function to ps called `readScore()` which will read the score under a read lock.

peer.go
```diff
@@ -159,11 +163,62 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+	if score == ps.score {
+		ps.RUnlock()
+		return
+	}
+	ps.RUnlock()
+
+	l.Lock()
+	l.peerHeap.RemovePeer(ps)
```
prashantv: Instead of removing the peer and readding it, can we update the peer using the `heap.Fix`? it's supposed to be more efficient

peer.go
prashantv: this change is still there even though there's no actual change, all that's happened is that it's moved. please revert

peer_heap.go
```diff
@@ -0,0 +1,95 @@
...
+// PopPeer pops the top peer of the heap.
+func (ph *PeerHeap) PopPeer() *peerScore {
+	return heap.Pop(ph).(*peerScore)
+}
+
+// PushPeer pushes the new peer into the heap.
+func (ph *PeerHeap) PushPeer(peerScore *peerScore) {
+	atomic.AddUint64(&(ph.order), 1)
+	// randRange will affect the deviation of peer's chosenCount
+	randRange := ph.Len()/2 + 1
```
prashantv: why divide by 2?

peer_strategies.go
```diff
@@ -20,40 +20,57 @@
...
 }
 
-func (r *preferIncoming) GetScore(p *Peer) uint64 {
-	rn := uint64(r.rng.Int31())
-	if p.NumInbound() > 0 {
-		rn += math.MaxUint32
+type preferIncomingCalculator struct{}
+
+func (preferIncomingCalculator) GetScore(p *Peer) uint64 {
+	p.mut.RLock()
```
prashantv: avoid read locks and unlocks constantly, use functions which will read the counts under the read lock and give you the count.

in general, I avoid `RLock` or `Lock` on some other object -- only use it on the receiver.

peer_strategies.go
```diff
@@ -20,40 +20,57 @@
...
 	}
-	return rn
+	p.mut.RUnlock()
+
+	return uint64(count)
 }
 
-func newPreferIncoming() *preferIncoming {
-	return &preferIncoming{rng: NewRand(time.Now().UnixNano())}
+func newPreferIncomingCalculator() preferIncomingCalculator {
```
prashantv: Add a comment about the strategy that this uses (prefers incoming, and then chooses based on least pending outbound calls)

peer_test.go
```diff
@@ -23,14 +23,17 @@ package tchannel_test
...
 	"testing"
 	"time"
 
-	. "github.com/uber/tchannel-go"
-
+	"github.com/GaryBoone/GoStats/stats"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	. "github.com/uber/tchannel-go"
```
prashantv: nit: put this import back where it was previously

peer_test.go
```diff
@@ -102,7 +105,6 @@ func TestGetPeerAvoidPrevSelected(t *testing.T) {
 			t.Errorf("Got unexpected error selecting peer: %v", err)
 			continue
 		}
-
```
prashantv: why delete this blank?

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+		if float64(v) < min || float64(v) > max {
+			t.Errorf("Key %v has value %v which is out of range %v-%v", k, v, min, max)
+		}
+	}
+
+	var cc []float64
+	for _, y := range counts {
+		cc = append(cc, float64(y))
+	}
+	fmt.Printf("standard dev %+v\n", stats.StatsPopulationStandardDeviation(cc))
```
prashantv: delete (and remove the stats dependency unless we really need it)

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+// setupServers will create numPeer servers, and register handlers on them.
+func (pt *peerSelectionTest) setupServers(t *testing.T) {
+	pt.servers = make([]*Channel, pt.numPeers)
+
+	// Set up numPeers servers.
+	for i := 0; i < pt.numPeers; i++ {
+		pt.servers[i], _ = pt.NewService(t, "hyperbahn")
+		pt.servers[i].Register(raw.Wrap(newTestHandler(pt.t)), "echo")
+	}
+	fmt.Println("all server nodes setup")
```
prashantv: delete stale print

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+
+func (pt *peerSelectionTest) setupAffinity(t *testing.T) {
+	var wg sync.WaitGroup
+	pt.affinity = make([]*Channel, pt.numAffinity)
+	for i := range pt.affinity {
+		pt.affinity[i] = pt.servers[i]
+	}
+
+	for _, server := range pt.servers {
+		pt.client.Peers().Add(server.PeerInfo().HostPort)
```
prashantv: this code already exists in `setupClient`, you can delete this.

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	pt := &peerSelectionTest{
+		peerTest:      peerTest{t: t},
+		numPeers:      numHyperbahn,
+		numConcurrent: numConcurrent,
+		numAffinity:   int(float64(numHyperbahn) * perAffinity),
+	}
+	defer pt.CleanUp()
+
+	pt.setupServers(t)
+	pt.setupClient(t)
```
prashantv: nit: prefer the order to be
`setupServers` then `setupAffinity` then `setupClient`

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	fmt.Println("affinity nodes connected")
+}
+
+func (pt *peerSelectionTest) setupClient(t *testing.T) {
+	pt.client, _ = pt.NewService(t, "client")
+	pt.client.Register(raw.Wrap(newTestHandler(t)), "echo")
+	for _, server := range pt.servers {
+		pt.client.Peers().Add(server.PeerInfo().HostPort)
+	}
+	time.Sleep(time.Millisecond * 100)
```
prashantv: why do you need a sleep?

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	tickAllClocks := func() {
+		for i := 0; i < numClock; i++ {
+			clocks[i] <- struct{}{}
+		}
+	}
+
+	const numTicks = 10000
+	for i := 0; i < numTicks; i++ {
+		if i%(numTicks/10) == 0 {
+			fmt.Println("upto", i, (100.0 * float64(i) / numTicks))
```
prashantv: let's make this print better
```go
fmt.Printf("Stress test progress: %v\n", 100 * float64(i) / numTicks)
```

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+
+	tickAllClocks := func() {
+		for i := 0; i < numClock; i++ {
+			clocks[i] <- struct{}{}
+		}
+	}
+
+	const numTicks = 10000
+	for i := 0; i < numTicks; i++ {
+		if i%(numTicks/10) == 0 {
```
prashantv: don't print this so often, probably want to make this more like `1000` or maybe `500`

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+			fmt.Println("upto", i, (100.0 * float64(i) / numTicks))
+		}
+		tickAllClocks()
+	}
+
+	time.Sleep(time.Millisecond * 100)
```
prashantv: why do you have a sleep?

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	for i := 0; i < numClock; i++ {
+		close(clocks[i])
+	}
+	wg.Wait()
+}
+
+func PeersHeapPerf(t *testing.T) {
```
prashantv: since this doesn't have a `Test` prefix, it won't even be run.

I would suggest doing something like this:

```go
func TestPeerHeapBalance(t *testing.T) {
  tests := []struct {
    numHyperbahn int
    affinityRatio float64
    numConcurrent int
    hasInboundCall bool
  }{
    {
      numHyperbahn: 1000,
      affinityRatio: 0.1,
      numConcurrent: 5,
      hasInboundCall: true,
    },
    {
      numHyperbahn: 100,
      affinityRatio: 0.1,
      numConcurrent: 1,
      hasInboundCall: true,
    },
    {
      numHyperbahn: 100,
      affinityRatio: 0,
      numConcurrent: 1,
      hasInboundCall: true,
    },
  }

  for _, tt := range tests {
    peerHeapStressTest(t, tt.numHyperbahn, tt.affinityRatio, tt.numConcurrent, tt.hasInboundCall)
  }
}
```

(The actual test cases will need tweaking, but it should be easy to add new cases to run, and we should have a few there)

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	CheckStress(t)
+
+	numHyperbahn := 1000
+	perAffinity := 0.1
+	numConcurrent := 5
+	hasInboundCall := true
+	peersHeapStress(t, numHyperbahn, perAffinity, numConcurrent, hasInboundCall)
+}
+
+func peersHeapStress(t *testing.T, numHyperbahn int, perAffinity float64, numConcurrent int, hasInboundCall bool) {
```
prashantv: rename `perAffinity` to `affinityRatio`. `perAffinity` could mean number of requests per affinity

prashantv: `hasInboundCall` is not used for anything?

peer_test.go
```diff
@@ -217,3 +219,268 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	pt.setupClient(t)
+	pt.setupAffinity(t)
+	pt.runStress()
+	validateStressTest(t, pt.client, pt.numAffinity)
+}
+
+func validateStressTest(t *testing.T, server *Channel, numAffinity int) {
+	state := server.IntrospectState(&IntrospectionOptions{IncludeEmptyPeers: true})
+
+	ones := make(map[string]int)
```
prashantv: `ones`? maybe `counts`?

subchannel.go
```diff
@@ -31,13 +31,16 @@ type SubChannelOption func(*SubChannel)
 
 // Isolated is a SubChannelOption that creates an isolated subchannel.
 func Isolated(s *SubChannel) {
+	s.Lock()
```
prashantv: good catch!

subchannel.go
```diff
@@ -160,3 +163,15 @@ func (subChMap *subChannelMap) getOrAdd(serviceName string, ch *Channel) *SubCha
 
 	return subChMap.registerNewSubChannel(serviceName, ch)
 }
+
+func (subChMap *subChannelMap) updatePeerHeap(p *Peer) {
+	subChMap.mut.RLock()
+	for _, subCh := range subChMap.subchannels {
+		subCh.RLock()
+		if subCh.Isolated() {
```
prashantv: the `subCh.RLock` should be inside of the `Isolated` function.

utils_for_test.go
```diff
@@ -54,3 +54,21 @@ func InboundConnection(call IncomingCall) (*Connection, net.Conn) {
 func NewSpan(traceID uint64, parentID uint64, spanID uint64) Span {
 	return Span{traceID: traceID, parentID: parentID, spanID: spanID, flags: defaultTracingFlags}
 }
+
+func (l *PeerList) GetHeap() *PeerHeap {
+	return l.peerHeap
+}
+
+func GetPendingRequests(p *Peer) int {
```
prashantv: not used anywhere?

prashantv:
The test is looking pretty good, just some cleanups left.

We also need to do performance testing. I can come by and help with that.

connection.go
junchaowu: There is not much difference here. If we have it, it is easy for us to debug and trace the event.

peer.go
junchaowu: I don't have it right now. It starts using lock because the choosePeer will pop up peer because of retry.

peer.go
junchaowu: reverted

peer_heap.go
junchaowu: the randRange will affect the peer's chosenCount range. This is number need to be tweak further. If we pick the ph.Len , it may exceed the 10% range sometimes.

peer_test.go
junchaowu: I tried again. The go-import sort this way.

peer_test.go
junchaowu: it is not often. It only print 10 times.

peer_test.go
junchaowu: Added hasInboundCall attr in the peerselectionTest struct.

peer_test.go
junchaowu: affinity nodes need to know client since they will add client's hostport into their peerlist

peer_test.go
prashantv: Ohh I see, you're doing a divide, this is good.

peer.go
```diff
@@ -159,11 +159,67 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+// exists checks if a hostport exists in the peer list.
+func (l *PeerList) exists(hostPort string) (*peerScore, bool) {
+	l.RLock()
+	ps, ok := l.peersByHostPort[hostPort]
+	l.RUnlock()
+
+	return ps, ok
+}
+
+// UpdatePeerHeap updates the peer heap.
```
prashantv: this function is leaking implementation details of the heap to the caller. I think we should rename this to `PeerUpdated`
and the comment can be like 
```go
// PeerUpdated is called when there is a change that may cause the peer's score to change.
// The new score is calculated, and the peer heap is updated with the new score
```

peer.go
```diff
@@ -159,11 +159,67 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+		return
+	}
+
+	newScore := l.scoreCalculator.GetScore(p)
+	if newScore == ps.readScore() {
+		return
+	}
+
+	l.Lock()
+	ps.setScore(newScore)
```
prashantv: nit: move this out of the lock

peer.go
```diff
@@ -159,11 +159,67 @@ func (l *PeerList) Copy() map[string]*Peer {
...
+func newPeerScore(p *Peer, score uint64) *peerScore {
+	return &peerScore{
+		Peer:  p,
+		score: score,
+		index: -1,
+	}
+}
+
+func (ps *peerScore) readScore() uint64 {
+	ps.RLock()
```
prashantv: defer is a little expensive (https://github.com/prashantv/go-bench/blob/master/mutex_results.md), it adds ~80ns. A simple read will be super fast so the defer is going to slow this method, so remove it.

change this code to do
```go
  ps.RLock()
  score := ps.score
  ps.RUnlock()

  return score
```

peer_test.go
```diff
@@ -217,3 +217,319 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+
+func testDistribution(t testing.TB, counts map[string]int, min, max float64) {
+	for k, v := range counts {
+		if float64(v) < min || float64(v) > max {
+			t.Errorf("Key %v has value %v which is out of range %v-%v", k, v, min, max)
+		}
+	}
+}
+
+type peerSelectionTest struct {
```
prashantv: can you rebase off master? this is mostly in master now right?

prashantv:
lgtm

PR #86, opened 2015-11-20
Add peer selection benchmark test.

junchaowu:
@prashantv

### DIFF
--- a/peer_test.go
+++ b/peer_test.go
@@ -277,8 +277,8 @@ func (pt *peerSelectionTest) setupAffinity(t testing.TB) {
 	wg.Add(pt.numAffinity)
 	// Connect from the affinity nodes to the service.
 	for _, affinity := range pt.affinity {
-		affinity.Peers().Add(hostport)
 		go func(affinity *Channel) {
+			affinity.Peers().Add(hostport)
 			pt.makeCall(affinity.GetSubChannel(serviceName))
 			wg.Done()
 		}(affinity)
--- a/peer_test.go
+++ b/peer_test.go
@@ -31,6 +31,7 @@ import (
 	. "github.com/uber/tchannel-go"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/uber/tchannel-go/raw"
 	"github.com/uber/tchannel-go/testutils"
 )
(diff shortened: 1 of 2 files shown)


peer_test.go
```diff
@@ -217,3 +218,127 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+
+	// Set up numPeers servers.
+	for i := 0; i < pt.numPeers; i++ {
+		pt.servers[i], _ = pt.NewService(t, "hyperbahn")
+		pt.servers[i].Register(raw.Wrap(newTestHandler(pt.t)), "echo")
+	}
+}
+
+func (pt *peerSelectionTest) setupAffinity(t testing.TB) {
+	var wg sync.WaitGroup
```
prashantv: nit: move this down to right before you need to use it (right before `wg.Add`

peer_test.go
```diff
@@ -217,3 +218,127 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	defer cancel()
+	_, _, _, err := raw.CallSC(ctx, sc, "echo", nil, nil)
+	require.NoError(pt.t, err, "raw.Call failed")
+}
+
+func (pt *peerSelectionTest) runStressSimple(b *testing.B) {
+	var wg sync.WaitGroup
+
+	// helper that will make a request every n ticks.
+	reqEveryNTicks := func(n int, sc *SubChannel) {
```
prashantv: name isn't accurate anymore -- this makes `n` requests, and has nothing to do with ticks. make sure you also update the comment

peer_test.go
```diff
@@ -217,3 +218,127 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	}
+
+	wg.Wait()
+}
+
+// Run these commands before run the benchmark.
+// sudo sysctl -w kern.maxfiles=50000
+// ulimit -n 50000
+func BenchmarkSimplePeersHeapPerf(b *testing.B) {
+
```
prashantv: nit: delete blank line

prashantv:
mostly good, just some minor comments

peer_test.go
```diff
@@ -217,3 +218,120 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	for i := 0; i < pt.numConcurrent; i++ {
+		go func(sc *SubChannel) {
+			defer wg.Done()
+			for j := 0; j < b.N; j++ {
+				pt.makeCall(sc)
+			}
+		}(sc)
+	}
+
+	wg.Add(pt.numConcurrent)
```
prashantv: `wg.Add` should be called before you start any goroutines, so move this to right before the for loop

peer_test.go
```diff
@@ -217,3 +218,120 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+	for _, server := range pt.servers {
+		pt.client.Peers().Add(server.PeerInfo().HostPort)
+	}
+}
+
+func (pt *peerSelectionTest) makeCall(sc *SubChannel) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+	_, _, _, err := raw.CallSC(ctx, sc, "echo", nil, nil)
+	require.NoError(pt.t, err, "raw.Call failed")
```
prashantv: make this an `assert.NoError` -- require doesn't work very well from goroutines started by that test (it'll kill that goroutine specific but other test goroutines won't be affected and so the test keeps running)

peer_test.go
```diff
@@ -217,3 +218,120 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+
+	// Connect from the affinity nodes to the service.
+	for _, affinity := range pt.affinity {
+		affinity.Peers().Add(pt.client.PeerInfo().HostPort)
+		go func(affinity *Channel) {
+			pt.makeCall(affinity.GetSubChannel(pt.client.PeerInfo().ServiceName))
+			wg.Done()
+		}(affinity)
+	}
+	wg.Add(pt.numAffinity)
```
prashantv: `Add` should be called before you start any goroutines`.

What I meant by the previous comment was usually you declare the variable closest to where you use it.

You should also cache the client's host port, and avoid doing things before and in the goroutine if you can do just one of those. It should look like this:

```go

clientHP := client.PeerInfo().HostPort

var wg sync.WaitGroup
wg.Add(pt.numAffinity)
for _, affinity := range pt.affinity {
  go func(affinity *Channel) {
    defer wg.Done()

    affinity.Peers().Add(clientHP)
    // makeCall
  }(affinity)
}
```

peer_test.go
```diff
@@ -217,3 +217,122 @@ func TestPeerSelectionPreferIncoming(t *testing.T) {
...
+		pt.affinity[i] = pt.servers[i]
+	}
+
+	hostport := pt.client.PeerInfo().HostPort
+	serviceName := pt.client.PeerInfo().ServiceName
+	var wg sync.WaitGroup
+	wg.Add(pt.numAffinity)
+	// Connect from the affinity nodes to the service.
+	for _, affinity := range pt.affinity {
+		affinity.Peers().Add(hostport)
```
prashantv: why is this outside of the goroutine?

junchaowu: isn't this same to have it in or out of the goroutine?

prashantv: It is the same, which is why you might as well put it in the goroutine. Then the user's context when reading this code is: each loop starts a goroutine for each affinity node, and that affinity node does <blah> work.

Currently, the user's context is:
each loop adds a peer to the affinity, *then* starts a goroutine. Usually that pattern means that something had to be done outside of the goroutine (e.g. for sync reasons), which is not the case here.

It just helps break apart the problem into simpler logical parts -- and in this case, reduces the number or "parts" you have to think about.

prashantv:
lgtm, just one last comment to fix

PR #88, opened 2015-11-23
Close InitialNodesFile in parseConfig

ajrharris:
If `parseConfig` is run too many times, we begin to see `too many open files` errors.

### DIFF
--- a/hyperbahn/client.go
+++ b/hyperbahn/client.go
@@ -97,12 +97,10 @@ func parseConfig(config *Configuration) error {
 		if err != nil {
 			return err
 		}
+		defer f.Close()
 
 		decoder := json.NewDecoder(f)
 		if err := decoder.Decode(&config.InitialNodes); err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


ajrharris:
@prashantv @junchaowu

hyperbahn/client.go
```diff
@@ -102,6 +102,9 @@ func parseConfig(config *Configuration) error {
 		if err := decoder.Decode(&config.InitialNodes); err != nil {
 			return err
 		}
+		if err = f.Close(); err != nil {
```
prashantv: Can you move this to a defer after the file is opened. That way, even if decoding fails, the file will still be closed.
```go
f, err := os.Open(config.InitialNodesFile)
if err != nil {
  return err
}
defer f.Close()
```

prashantv:
Thanks for the fix, just needs a small change.

junchaowu:
thx

PR #89, opened 2015-11-23
Fix "the connection state update doesn't update peer score" bug.

junchaowu:
@prashantv


junchaowu:
break this patch into two pieces.

PR #91, opened 2015-11-24
Fix for new inbound or closed connections not updating peer's score

junchaowu:
@prashantv

### DIFF
--- a/peer_test.go
+++ b/peer_test.go
@@ -320,9 +320,16 @@ type peerSelectionTest struct {
 	// hasInboundCall is the bool flag to tell whether to have inbound calls from affinity nodes
 	hasInboundCall bool
 
-	servers  []*Channel
-	affinity []*Channel
-	client   *Channel
+	servers            []*Channel
+	affinity           []*Channel
+	affinityWithNoCall []*Channel
+	client             *Channel
+}
+
+func (pt *peerSelectionTest) setup(t testing.TB) {
+	pt.setupServers(t)
+	pt.setupClient(t)
+	pt.setupAffinity(t)
 }
 
 // setupServers will create numPeer servers, and register handlers on them.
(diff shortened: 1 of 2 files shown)


peer_test.go
```diff
@@ -355,6 +356,8 @@ func (pt *peerSelectionTest) setupAffinity(t testing.TB) {
 		}(affinity)
 	}
 
+	// use ping to build connection without sending call req.
+	pt.makePing(pt.servers[pt.numAffinity-pt.numAffinityWithNoCall], hostport)
```
prashantv: let's rename `makePing` to `sendPing` (optionally rename `makeCall` to `sendCall`)

You're only making a ping request from a single node right now.

peer_test.go
```diff
@@ -392,7 +402,7 @@ func (pt *peerSelectionTest) runStressSimple(b *testing.B) {
 }
 
 func (pt *peerSelectionTest) runStress() {
-	numClock := pt.numConcurrent + pt.numAffinity
+	numClock := pt.numConcurrent + pt.numAffinity - pt.numAffinityWithNoCall
```
prashantv: we keep repeating this `pt.numAffinity - pt.numAffinityWithNoCall`. I think it might be cleaner (and less error prone) if we treat them as separate lists of affinity nodes. So you have numAffinity = 3, numAffinityWithoutCall = 4, and so there's 7 total affinity nodes. That way you can have more affinities with no call than num affinity too.

peer_test.go
```diff
@@ -502,11 +512,12 @@ func TestPeerHeapPerf(t *testing.T) {
...
-		numPeers:       numserver,
-		numConcurrent:  numConcurrent,
-		hasInboundCall: hasInboundCall,
-		numAffinity:    int(float64(numserver) * affinityRatio),
+		peerTest:              peerTest{t: t},
+		numPeers:              numserver,
+		numConcurrent:         numConcurrent,
+		hasInboundCall:        hasInboundCall,
+		numAffinity:           int(float64(numserver) * affinityRatio),
+		numAffinityWithNoCall: 1,
```
prashantv: Let's make this > 1, like 3 or something.

peer_test.go
```diff
@@ -539,3 +550,30 @@ func validateStressTest(t testing.TB, server *Channel, numAffinity int) {
 	median := counts[len(counts)/2]
 	testDistribution(t, countsByPeer, float64(median)*0.9, float64(median)*1.1)
 }
+
+func (pt *peerSelectionTest) runAfterConnClosed() {
```
prashantv: I don't think this should be part of this test.

Make a separate test that tests only this behaviour. That test doesn't need to be part of the stress test, and should run along with all other tests.

peer_test.go
```diff
@@ -475,18 +500,18 @@ func TestPeerHeapPerf(t *testing.T) {
...
-			numConcurrent:  5,
-			hasInboundCall: true,
-		},
-		{
-			numserver:      1000,
-			affinityRatio:  0.1,
-			numConcurrent:  1,
-			hasInboundCall: true,
-		},
+		// {
```
prashantv: uncomment these

peer_test.go
```diff
@@ -531,11 +554,31 @@ func validateStressTest(t testing.TB, server *Channel, numAffinity int) {
...
 	median := counts[len(counts)/2]
 	testDistribution(t, countsByPeer, float64(median)*0.9, float64(median)*1.1)
 }
+
+func TestPeerSelectionAfterClosed(t *testing.T) {
+	pt := &peerSelectionTest{
+		peerTest:              peerTest{t: t},
+		numPeers:              100,
+		numAffinity:           20,
+		numAffinityWithNoCall: 0,
```
prashantv: nit: delete this, since it's the 0 value anyway

prashantv:
lgtm, just fix the minor comments.

PR #94, opened 2015-11-25
Feature/external templates

thanodnl:
This feature adds an extra optional flag to `thrift-gen` which can be used to pass in external template files to the code generator. The flag for this is `--template <path-to-template-file>`.

The driver behind this extension is for `ringpop-go` to tie into the code generation process. With this extension we are able to define a `[ringpop.thrift-gen](https://github.com/uber/ringpop-go/blob/feature/thrift-forward-adapter/ringpop.thrift-gen)` file in our repository that gets compiled during the code generation phase in the build cycle.

By keeping this file in our own repository we can version control it next to the library being used in the generated code. It gives developers of external libraries the possibility to add new code to be generated outside of the tchannel-go package enabling easier experimentation. It also has the nice benefit that it can be upgraded together with changes to the ringpop library. Making sure that the generated code is always in sync with the version of the library being used.

Multiple projects can benefit from the same code generation phase by having multiple `--template` flags passed in to the program. In an ideal case we would move the tchannel template to `tchan.thrift-gen` in the root of this project, but there is a slight difference in the parameters being passed into the template render engine that it is not a trivial change. So if we would like to accomplish that I would suggest to do so in its own PR.

Please have a look at the changes and fire any feedback you have to me and I will resolve it.

cc @prashantv


thrift/thrift-gen/compile_test.go
```diff
@@ -87,7 +87,7 @@ func runTest(t *testing.T, thriftFile string) error {
 
 	// Generate code from the Thrift file.
 	t.Logf("runTest in %v", tempDir)
-	if err := processFile(true /* generateThrift */, thriftFile, ""); err != nil {
+	if err := processFile(true /* generateThrift */, thriftFile, []string{}, ""); err != nil {
```
prashantv: `nil` is a valid slice, don't need to do `[]string{}`

thrift/thrift-gen/main.go
```diff
@@ -41,14 +41,30 @@ import (
 
 const tchannelThriftImport = "github.com/uber/tchannel-go/thrift"
 
+type stringslice []string
```
prashantv: nit: camcel case, and put `flag` in the name so we know what it's for, `stringSliceFlag`

thrift/thrift-gen/main.go
```diff
@@ -83,15 +106,43 @@ func processFile(generateThrift bool, inputFile string, outputFile string) error
...
+		if err := generateCode(outputFile, goTmpl, packageName(thriftFilename), AST); err != nil {
 			return err
 		}
-		// TODO(prashant): Support multiple files / includes etc?
-		return nil
-	}
 
+		templates, err := getTemplatesFromFiles(templateFiles, AST)
+		if err != nil {
+			return fmt.Errorf("Could not parse thrift-gen file: %v", err)
```
prashantv: errors must always be lower case. the error message is a littel vague, maybe "could not parse template: %v"

thrift/thrift-gen/main.go
```diff
@@ -83,15 +106,43 @@ func processFile(generateThrift bool, inputFile string, outputFile string) error
...
+		}
+
+		for templateFilename, template := range templates {
+			buf := &bytes.Buffer{}
```
prashantv: can we put the logic inside this for into a separate function -- this function is getting a little large.

thrift/thrift-gen/main.go
```diff
@@ -83,15 +106,43 @@ func processFile(generateThrift bool, inputFile string, outputFile string) error
...
+			buf.WriteString("// @generated Code generated by thrift-gen. Do not modify.\n\n")
+
+			if err := template.Execute(buf, &FileBasedTemplateDate{
```
prashantv: this template data is the same regardless of the templateFilename right? it can be created outside of the loop

thrift/thrift-gen/main.go
```diff
@@ -83,15 +106,43 @@ func processFile(generateThrift bool, inputFile string, outputFile string) error
...
+				Package: packageName(thriftFilename),
+				AST:     AST,
+				State:   NewState(AST),
+			}); err != nil {
+				return fmt.Errorf("Could not parse thrift-gen template: %v", err)
```
prashantv: lower case error message, and avoid "thrift-gen" in the error, I think just "could not parse template: %v" is fine.

thrift/thrift-gen/main.go
```diff
@@ -83,15 +106,43 @@ func processFile(generateThrift bool, inputFile string, outputFile string) error
...
+			}
+
+			// post process generated code
+			generatedBytes := buf.Bytes()
+			generatedBytes = cleanGeneratedCode(generatedBytes)
+
+			// write file
+			outputFilename := getOuputFilename(outputFile, thriftFilename, AST, templateFilename)
```
prashantv: there's a lot of code duplication with `generateCode` here, let's DRY this up.

thrift/thrift-gen/main.go
```diff
@@ -149,3 +200,67 @@ func cleanGeneratedCode(generated []byte) []byte {
...
+			return nil, err
+		} else {
+			// add parsed template to templates
+			output[filename] = parsedTemplate
+		}
+	}
+	return output, nil
+}
+
+func outputDirectory(infile string, AST *parser.Thrift) string {
```
prashantv: why does `outputDirectory` needs the `AST`?

thrift/thrift-gen/main.go
```diff
@@ -149,3 +200,67 @@ func cleanGeneratedCode(generated []byte) []byte {
...
+	dir := outputDirectory(outputFile, AST)
+
+	baseThriftFilename := getBasename(thriftFile)
+	baseGenFilename := getBasename(genFile)
+
+	return filepath.Join(dir, baseGenFilename+"-"+baseThriftFilename+".go")
+}
+
+func getBasename(path string) string {
+	_, filename := filepath.Split(path)
```
prashantv: `filepath.Base` might be more appropriate here

thrift/thrift-gen/main.go
```diff
@@ -149,3 +200,67 @@ func cleanGeneratedCode(generated []byte) []byte {
...
+			return strings.ToLower(in)
+		},
+		"goPublicName": goPublicName,
+		"goType": func(t *parser.Type) string {
+			return state.goType(t)
+		},
+	}
+
+	// read template file
+	templateBytes, err := ioutil.ReadFile(filename)
```
prashantv: I thought we decided to do paths relative to the `GOPATH` rather than absolute paths?

prashantv:
Thanks for making this change @thanodnl 

Can we add unit tests for this -- I'd like to see tests that verify that the functionality works, and this will ensure it isn't broken by any future changes or refactorings.

thrift/thrift-gen/main.go
```diff
@@ -83,15 +106,43 @@ func processFile(generateThrift bool, inputFile string, outputFile string) error
...
 
-	goTmpl := parseTemplate()
-	for filename, v := range parsed {
-		if err := generateCode(outputFile, goTmpl, packageName(filename), v); err != nil {
+	goTmpl := parseTemplate() // tchan-template
+	for thriftFilename, AST := range parsed {
+		if err := generateCode(outputFile, goTmpl, packageName(thriftFilename), AST); err != nil {
 			return err
 		}
-		// TODO(prashant): Support multiple files / includes etc?
```
prashantv: This TODO should not be removed, it's still a valid TODO.

thrift/thrift-gen/main.go
thanodnl: Good catch!

This is probably from one of the earlier prototypes where I got the package name from the AST to get the correct path to write to. After some refactoring that wasn't needed anymore, but apparently I didn't remove it.

thrift/thrift-gen/main.go
thanodnl: I was thinking about passing in the absolute path to the file and do the resolution relative to the `GOPATH` in the `Makefile`. However I'm running into some problems with trying that so I might implement it in here. However I think it is also good that one can pass in an absolute path. So the logic would look something like:

1. Resolve `filename` for relative path to `pwd`
2. If file does not exists Iterate over entries in `GOPATH`
  - Check for existence of file
  - Exit iteration if file exists

thanodnl:
I have refactored the application a bit to make it more DRY.

In the process I have added two extra flags to the application:

1. `--outputDir`: Since there are potentially multiple files being outputted it makes more sense to provide the directory where the files should be generated. Due to the behavior of the `thrift` executable it will generate a directory named after the `.thrift` file in the `outputDirectory`
2. `--skipTChannel`: The default behavior of `thrift-gen` will stay the same where it compiles `tchan-<service>.go` for the embedded application. However, if the `--skipTChannel` flag is passed it will not do so. In that case you can pass in `--template github.com/uber/tchannel-go/tchan.thrift-gen`. This is to be both backwards compatible as future proof. Since the `thrift-gen` binary might be not in sync with the used version of `tchannel-go` but the `tchan.thrift-gen` file will be in sync with the version of `tchannel-go` that is used in the application we can more easily change the generated code.

I added tests around this flag and the compilation of the template in `compile_test.go` where it runs twice over all thrift files in the `test_files` directory. Once without templates and without the `skipTChannel` flag. This compiles the embedded template. The second run through all the files it sets the `skipTChannel` flag, and uses the template `github.com/uber/tchannel-go/tchan.thrift-gen`. In both runs it builds the generated code to validate that it is compiled correctly.

@prashantv please have a second look at this PR. I have edited more than I would have liked for this feature, but it is a lot more DRY now, and I already started pulling the tchannel template to its own file. The only reason that I kept the hardcoded template in here is so people can use older versions of tchannel with this version of `thrift-gen` without changing their build scripts. At some point it would be better to remove the hardcoded template from the generator.

prashantv:
Thanks for the changes @thanodnl 

There's some more changes to make, but I think I can take over from here. The changes I want to make are
 - the tchannel template should remain built in so that the binary by itself can be used without any external dependencies for tchannel generation (e.g. the way we use it internally with go-build)
 - I am making changes for includes which has some overlap with requiring outputDirectory, so I'll split that out

I'll need to figure out how to make this work with include support as well

prashantv:
As discussed offline, made some changes and some refactorings and created #104 

Thanks for pushing this @thanodnl

PR #97, opened 2015-11-30
Add method for Hyperbahn Discovery API

dansimau:
This adds Discover(serviceName) to the Go Hyperbahn client.


hyperbahn/discover.go
```diff
@@ -0,0 +1,45 @@
...
+package hyperbahn
+
+import (
+	"time"
+
+	"github.com/uber/tchannel-go/hyperbahn/gen-go/hyperbahn"
+	"github.com/uber/tchannel-go/thrift"
+)
+
+func (c *Client) Discover(serviceName string) ([]*hyperbahn.ServicePeer, error) {
```
dansimau: Should this be returning the raw generated type here (`[]*hyperbahn.ServicePeer`)? Or should the hyperbahn package (Hyperbahn client) define a type?

prashantv: I'd prefer if the hyperbahn package defined a type. That way it hides the implementation details of the call and generated code which may change over time.

hyperbahn/discover.go
```diff
@@ -0,0 +1,45 @@
...
+
+	thriftClient := thrift.NewClient(c.tchan, hyperbahnServiceName, nil)
+	hyperbahnClient := hyperbahn.NewTChanHyperbahnClient(thriftClient)
+
+	ctx, cancel := thrift.NewContext(time.Second)
+	ctx = thrift.WithHeaders(ctx, map[string]string{"user": "anonymous"})
```
prashantv: do you need headers? if not, remove this line

hyperbahn/discover.go
```diff
@@ -0,0 +1,45 @@
...
+	"time"
+
+	"github.com/uber/tchannel-go/hyperbahn/gen-go/hyperbahn"
+	"github.com/uber/tchannel-go/thrift"
+)
+
+func (c *Client) Discover(serviceName string) ([]*hyperbahn.ServicePeer, error) {
+
+	thriftClient := thrift.NewClient(c.tchan, hyperbahnServiceName, nil)
+	hyperbahnClient := hyperbahn.NewTChanHyperbahnClient(thriftClient)
```
prashantv: let's make this once when creating the `Client` and reuse the same `hyperbahnClient`.

prashantv:
Thanks for this change Dan. Can you also add a simple unit test (take a look at how the advertise tests mock Hyperbahn, and `thrift_test` for how you can mock a Thrift service).

dansimau:
Hey, sorry it took so long to get back to this @prashantv. I added some tests by mocking Hyperbahn using testify/mock, as this was relatively quick. Do you think I need to to expand this further?

The tests I added cover my new code, but don't cover the actual Thrift/Hyperbahn mechanics. Does that need to be tested, do you think?

hyperbahn/client.go
```diff
@@ -86,7 +91,10 @@ func NewClient(ch *tchannel.Channel, config Configuration, opts *ClientOptions)
 		addPeer(ch, node)
 	}
 
-	client.client = tjson.NewClient(ch, hyperbahnServiceName, nil)
+	client.jsonClient = tjson.NewClient(ch, hyperbahnServiceName, nil)
+	client.thriftClient = tthrift.NewClient(ch, hyperbahnServiceName, nil)
```
prashantv: the Thrift client isn't used after this, so I'd just make it a local here rather than adding it to `Client`

hyperbahn/discover_test.go
```diff
@@ -0,0 +1,115 @@
...
+	s.client.hyperbahnClient = s.mock
+}
+
+func (s *DiscoverTests) TearDownTest() {
+	s.channel.Close()
+	s.client.Close()
+}
+
+func (s *DiscoverTests) TestDiscoverSuccess() {
+	s.mock.On("Discover", mock.Anything, mock.Anything).Return(&hyperbahn.DiscoveryResult_{Peers: testServicePeers}, nil)
```
prashantv: let's make the second argument Discovery argument specifically look for `test`,
also let's make this line a little easier to read,
```go
s.mock.On("Discover", mock.Anything, "test").
  Return(&hyperbahn.DiscoveryResult_{Peers: testServicePeers}, nil)
```

prashantv:
Thanks for adding the tests Dan. I refactored the code a little -- I figured it'll be useful for others to be able to mock `Discover`, so I updated `mockhyperbahn` (http://godoc.org/github.com/uber/tchannel-go/testutils/mockhyperbahn) to allow mocking the results.

I then updated the tests to use the mock hyperbahn, which simplified them a little (and ends up using the Thrift/TChannel tests, and avoids setting private fields on the Hyperbahn client).

PR #104, opened 2015-12-02
Add support for external templates

prashantv:
Based on #94 with some refactorings.

External templates can be specified via the `-template` flag. This flag can be specified multiple times for multiple templates. Paths can be `GOPATH` relative, so the user does not need to specify absolute paths.

@thanodnl

### DIFF
--- a/thrift/thrift-gen/template.go
+++ b/thrift/thrift-gen/template.go
@@ -40,7 +40,9 @@ type Template struct {
 
 func parseTemplate(contents string) (*template.Template, error) {
 	funcs := map[string]interface{}{
-		"contextType": contextType,
+		"contextType":   contextType,
+		"goPrivateName": goName,
+		"goPublicName":  goPublicName,
 	}
 	return template.New("thrift-gen").Funcs(funcs).Parse(contents)
 }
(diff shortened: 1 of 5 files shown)


thrift/thrift-gen/template.go
```diff
@@ -29,14 +29,42 @@ import (
...
+type Template struct {
+	name     string
+	template *template.Template
+}
+
+func parseTemplate(contents string) (*template.Template, error) {
 	funcs := map[string]interface{}{
-		"contextType": contextType,
+		"contextType":  contextType,
+		"lowercase":    strings.ToLower,
```
thanodnl: I was thinking to maybe change this to `goPrivateName` since that is what it is used for and more in line with `goPublicName`. Shows the intent a bit better in templates.

prashantv: Is that the intent for this function? In that case, we definitely shouldn't use `lowerCase`, but I can put in `goPrivateName` which will actually point to the function `goName` in `names.go`. That way, you can pass in names that are reserved keywords and it'll automatically add a suffix.

PR #106, opened 2015-12-03
The goType function should be passed in to the initial funcs map for …

thanodnl:
…templates using it to compile.

@prashantv I was pretty sure that I tested the binary, but somehow the latest version did not compile my template. During the parsing phase of the template it didn't have a `goType` func defined. I added a stub in the initial function list, and added tests to make sure the `goType` is replaced with the working version of the function.

In the test I used a `.thrift` file that has a User defined struct since that was the first time that I ran into the necessity of using `goType`.

cc @uber/ringpop

### DIFF
--- a/thrift/thrift-gen/compile_test.go
+++ b/thrift/thrift-gen/compile_test.go
@@ -115,6 +115,45 @@ func TestExternalTemplate(t *testing.T) {
 	}
 }
 
+func TestExternalTemplateWithGoType(t *testing.T) {
+	template1 := `package {{ .Package }}
+
+{{ range .AST.Services }}
+// Service {{ .Name }} has {{ len .Methods }} methods.
+{{ range .Methods }}
+// func {{ .Name | goPublicName }} ({{ range .Arguments }}{{ .Type | goType }}, {{ end }}) ({{ .ReturnType | goType }}){{ end }}
+{{ end }}
+	`
+	templateFile := writeTempFile(t, template1)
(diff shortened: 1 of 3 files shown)


thrift/thrift-gen/compile_test.go
```diff
@@ -115,6 +115,45 @@ func TestExternalTemplate(t *testing.T) {
 	}
 }
 
+func TestExternalTemplateWithGoType(t *testing.T) {
```
prashantv: Too much code duplication. Just change the above test to range over the methods and add that to the verification rather than duplicating the whole test.

thrift/thrift-gen/template.go
```diff
@@ -43,6 +43,7 @@ func parseTemplate(contents string) (*template.Template, error) {
 		"contextType":   contextType,
 		"goPrivateName": goName,
 		"goPublicName":  goPublicName,
+		"goType":        func() string { return "" },
```
prashantv: let's make a separate function `dummyGoType` and use that. The `dummyGoType` function should have a comment explaining why it's needed and that it will be replaced.

prashantv:
Good catch!

thanodnl:
I made some changes to have less code duplication. I had to add some arguments and return values to the service definition to make sure that `goType` was called and replaced with a working version.

prashantv:
lgtm ,thanks!

PR #107, opened 2015-12-03
Allow mockery code generation on TChanClient.

thanodnl:
When using `mockery` to generate mocks for interfaces it generates uncompilable code for `TChanClient` due to package name collision on `thrift`. By importing apache's thrift as `athrift` mockery is able to generate compilable code so `TChanClient` can be mocked during tests in [ringpop-go](https://github.com/uber/ringpop-go).

### DIFF
--- a/thrift/interfaces.go
+++ b/thrift/interfaces.go
@@ -20,7 +20,7 @@
 
 package thrift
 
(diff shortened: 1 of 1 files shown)


prashantv:
lgtm

PR #112, opened 2015-12-16
Add a sampling trace reporter

prashantv:
The sampling trace reporter can wrap any other trace reporter and sample traces.

cc @yurishkuro

### DIFF
--- a/trace/sampling_reporter.go
+++ b/trace/sampling_reporter.go
@@ -1,0 +1,59 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


trace/sampling_reporter.go
```diff
@@ -0,0 +1,59 @@
...
+		underlying: reporter,
+	}
+}
+
+type samplingReporter struct {
+	sampleRate float64
+	underlying tc.TraceReporter
+}
+
+func (r *samplingReporter) Report(data tc.TraceData) {
```
yurishkuro: sampling needs to be done when creating a brand new trace ID, not when reporting a span. Once a trace ID is chosen to be sampled, the `flags` field must be propagated to all subordinate requests and cause those spans to be reported as well.

prashantv: Ahh, so it looks like sampling will have to be built in to tchannel itself since it selects the trace ID. Thanks

junchaowu: true also you should be the edge client.

PR #113, opened 2015-12-16
Move trace sampling to when root spans are created

prashantv:
Sampling decisions should be made by TChannel when creating the span,
since the flag is set only at the edge when creating the root span.

`DefaultTraceSampleRate` is 1.0 so it's backwards compatible.

@yurishkuro @junchaowu


tracing.go
```diff
@@ -109,3 +110,14 @@ func (s Span) NewChildSpan() *Span {
...
 	return childSpan
 }
+
+func (s *Span) sampleRootSpan(sampleRate float64) {
+	// Sampling only affects root spans
+	if s.ParentID() != 0 {
+		return
+	}
+
+	if rand.Float64() > sampleRate {
```
yurishkuro: I think you want `< sampleRate`

channel.go
```diff
@@ -43,6 +43,9 @@ var (
 
 const (
 	ephemeralHostPort = "0.0.0.0:0"
+
+	// DefaultTraceSampleRate is the default sampling rate for traces.
+	DefaultTraceSampleRate = 1.0
```
yurishkuro: This is the probability, correct?  Can we make the default to be `0.01` (1pct), as in Node implementation?

prashantv: This is probability. I made it 1.0 so the behavior stays the same as previous versions.

I'll chat with the Python team tomorrow and find out their default.

tracing.go
prashantv: I use > since it's to decide whether to *not* sample.

If the sampleRate is 0.3, then we don't want to sample most of the time, which is when the random float is > 0.3

channel.go
yurishkuro: Yes, it's also 1 in Python, but 0.01 in Node.  There are two sides of the coin:
1. for any service with reasonable traffic, the 100% sampling can easily overwhelm the collectors, 
2. for low-traffic services the `0.01` setting might produce no traces. 

Of those two alternatives it's better to be on the safe side. Most people will go to prod without changing the default.

channel.go
```diff
@@ -162,12 +170,18 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		timeNow = time.Now
 	}
 
+	traceSampleRate := opts.TraceSampleRate
+	if traceSampleRate == 0 {
```
yurishkuro: maybe `== null`?  Setting rate to 0 is an effective way to turn off sampling.

tracing.go
yurishkuro: sg.

channel.go
prashantv: As discussed offline, the public library will stay at 100% so the behaviour doesn't change.

Our internal wrapper libraries can set the default to 1%.

PR #118, opened 2015-12-17
Isolated subchannels should not prefer incoming

prashantv:
Isolated subchannels (usually used by ringpop applications) do not have
a prefernece on peers with incoming vs outgoing connections.
Add a leastPending calculator and update tests to verify that neither
incoming or outgoing are preferred.

@junchaowu


prashantv:
Opening a new pull request against master.

PR #139, opened 2015-12-29
Fix infinite loop in VerifyNoBlockedGoroutines

prashantv:
Leak attempts was not being incremented correctly, and so leaked stacks
could cause the verification to enter an infinite loop.


breerly:
Much cleaner. I'm guessing that adding a test is not practical?

prashantv:
Yeah thought about tests, but since it verifies the global goroutine state, it'd be tricky to validate, and since this is only a test utility, figured it wasn't worth it.

breerly:
:ship: :it:

PR #143, opened 2015-12-29
Fix the use of an atomic int.

nomis52:


peer_heap.go
```diff
@@ -88,13 +88,14 @@ func (ph *PeerHeap) PopPeer() *peerScore {
 
 // PushPeer pushes the new peer into the heap.
 func (ph *PeerHeap) PushPeer(peerScore *peerScore) {
-	atomic.AddUint64(&(ph.order), 1)
+	newOrder := atomic.AddUint64(&(ph.order), 1)
```
prashantv: (unrelated, but since you're updating this line) can you remove the parens around `ph.order`, they should not be required:
http://play.golang.org/p/30H2gFZ2Nv

prashantv:
Good catch! lgtm, feel free to merge once my minor comment is addressed.

PR #144, opened 2015-12-29
Run additional tests.

nomis52:
There was a bug in the makefile where it wouldn't run the tests in the top
level directory.

This also adds support for running the tests on Darwin by setting the correct
Thrift path.


prashantv:
The top level directory bug should be fixed by #130 -- can you rebase and make sure it works fine.

The use of the correct platform specific binary looks good, can merge once you rebase and the conflicts are gone.

nomis52:
Updated.

PR #145, opened 2015-12-29
Peer heap

nomis52:
Depends on #144

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ THRIFT_REL := ./scripts/travis/thrift-release/$(PLATFORM)-$(ARCH)
 
 export PATH := $(realpath $(THRIFT_REL)):$(PATH)
 
-
 # Separate packages that use testutils and don't, since they can have different flags.
 # This is especially useful for timeoutMultiplier and connectionLog
 TESTUTILS_TEST_PKGS := . hyperbahn testutils http json thrift pprof trace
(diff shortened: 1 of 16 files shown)


prashantv:
lgtm, you probably want to merge the other changes and rebase before merging this.

Makefile
```diff
@@ -8,7 +8,12 @@ TEST_PKGS := $(addprefix github.com/uber/tchannel-go,$(TEST_PKGS_NOPREFIX))
 BUILD := ./build
 SRCS := $(foreach pkg,$(PKGS),$(wildcard $(pkg)/*.go))
 export GOPATH = $(GODEPS):$(OLDGOPATH)
-export PATH := $(realpath ./scripts/travis/thrift-release/linux-x86_64):$(PATH)
+
+PLATFORM := $(shell uname -s | tr '[:upper:]' '[:lower:]')
+ARCH := $(shell uname -m)
+THRIFT_REL := ./scripts/travis/thrift-release/$(PLATFORM)-$(ARCH)
```
breerly: This is sweet.

akshayjshah:
@nomis52 Ping :)

Can I help get this ready to land?

prashantv:
As of #269, we've documented that the peer heap is not safe for concurrent use, and expects callers (in this case `PeerList`) to serialize access to the heap. We no longer use atomic operations in the heap, so this PR shouldn't be required anymore.

PR #146, opened 2015-12-29
Add UnmarshallYAML to hyperbahn FailStrategy

chucknthem:
Adding UnmarshallYAML so we can specify a FailStrategy in yaml config files.

### DIFF
--- a/hyperbahn/advertise_test.go
+++ b/hyperbahn/advertise_test.go
@@ -306,7 +306,7 @@ func TestRetryFailure(t *testing.T) {
 	})
 }
 
-func TestCanUnmarshallYAMLFailStrategy(t *testing.T) {
+func TestCanUnmarshalYAMLFailStrategy(t *testing.T) {
 	tests := []struct {
 		strategy string
 		expected FailStrategy
--- a/hyperbahn/advertise_test.go
+++ b/hyperbahn/advertise_test.go
@@ -32,6 +32,8 @@ import (
 	"github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/json"
 	"github.com/uber/tchannel-go/testutils"
+
+	"gopkg.in/yaml.v2"
 )
 
 func TestAdvertiseFailed(t *testing.T) {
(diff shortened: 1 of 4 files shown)


hyperbahn/advertise_test.go
```diff
@@ -304,6 +304,17 @@ func TestRetryFailure(t *testing.T) {
 	})
 }
 
+func TestCanUnmarshallYAMLFailStrategy(t *testing.T) {
```
prashantv: Can you make this test just go through the different strategies and ensure they come as expected using a table test:

we'll need to add yaml to the godeps (there's no vendoring in this repo, just the godeps.json will be updated).

```go
tests := []struct{
  strategy string
  expected FailStrategy
  wantErr bool
}{
  {
    strategy: "fatal",
    expected: FailStrategyFatal,
  },
  {
    strategy: "ignore",
    expected: FailStrategyIgnore,
  },
  {
    strategy: "unknown",
    wantErr: true,
  },
  {
    strategy: "",
    wantErr: true,
  },
}

for _, tt := range tests {
  var fs FailStrategy
  err := yaml.Unmarshal([]byte(tt.strategy), &fs)
  if tt.wantErr {
    assert.Error(t, err, "Unmarshal %v should fail", tt.strategy)
  } else {
    assert.NoError(t, err, "Unmarshal %v shouldn't fail", tt.strategy)
  }
  if err != nil {
    continue
  }
  assert.Equal(t, tt.expected, fs, "FailStrategy mismatch")
}

```

hyperbahn/client.go
```diff
@@ -58,6 +58,24 @@ const (
 
 const hyperbahnServiceName = "hyperbahn"
 
+func (f *FailStrategy) UnmarshalYAML(unmarshal func(v interface{}) error) error {
+	var strategy string
+	if err := unmarshal(&strategy); err != nil {
+		return err
+	}
+	var err error
```
prashantv: nit: instead of creating an err variable, you should just return in the default case
```go
switch strategy {
  case "fatal":
    *f = FailStrategyFatal
  case "ignore":
    *f = FailStrategyIgnore
  default:
    return fmt.Errorf(...)
}
return nil
```

hyperbahn/advertise_test.go
akshayjshah: Nit: "Unmarshal" should have only one L.

prashantv:
lgtm, thanks for the change

PR #147, opened 2015-12-29
Expose remote PeerInfo on IncomingCall

prashantv:
This lets you access the host:port of the caller for an incoming call. You can access this using `CurrentCall(ctx).RemotePeer()`

cc: @nomis52  @shawnburke


nomis52:
LGTM

PR #153, opened 2016-01-04
Embed mutexes where idiomatic

### LINKED ISSUE #96: Embed mutexes rather than having them as a separate field
Right now, some mutexes are embedded, while others typically use `mu` as the name. We should embed all mutexes and be consistent.

akshayjshah:
Standardize on embedding mutexes. Fixes #96.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -119,11 +119,11 @@ type Channel struct {
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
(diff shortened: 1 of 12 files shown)


akshayjshah:
First day, first PR. :feelsgood:

prashantv:
nice, lgtm :ship:

akshayjshah:
:shipit: > :ship:

PR #154, opened 2016-01-05
Wrap net.Listener for better Close semantics

### LINKED ISSUE #141: Close does not wait till the socket is destroyed, so channel could accept new connections
See:
https://github.com/golang/go/issues/13762
and
https://github.com/golang/go/issues/10527

We should wrap the listener in a similar way to what is proposed in the last issue so that clients don't need to work around it like in:
https://github.com/uber/tchannel-go/commit/871a7876b408f2d00e802ac124306dac5674b64d

prashantv:
Fixes #141

### DIFF
--- a/tnet/listener.go
+++ b/tnet/listener.go
@@ -1,0 +1,96 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package tnet
+
+import (
+	"net"
+	"sync"
+)
+
+// Wrap returns a new Listener around the provided net.Listener.
+// The returned Listener has a guarantee that when Close returns, it will no longer
+// accept any new connections.
+// See: https://github.com/uber/tchannel-go/issues/141
+func Wrap(l net.Listener) net.Listener {
+	return &listener{Listener: l, cond: sync.NewCond(&sync.Mutex{})}
+}
+
+// listener wraps a net.Listener and ensures that once Listener.Close returns,
+// the underlying socket has been closed.
+//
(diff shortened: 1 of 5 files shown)


tnet/listener.go
```diff
@@ -0,0 +1,101 @@
...
+// Goroutine 2 returns to the caller, makes a new connection.
+// The new connection is accepted by the socket (since it hasn't been destroyed)
+// Goroutine 1 returns from epoll, and has a new connection that it accepts.
+//
+// To avoid this, we block Goroutine 2 from returning till Accept returns
+// to the user, and they get an error.
+type listener struct {
+	l net.Listener
+
+	// cond is used signal Close when there are no references left to the
```
abhinav: you accidentally some words there

tnet/listener.go
```diff
@@ -0,0 +1,101 @@
...
+//  the socket, but instead decrements a reference, marks the connection as closed
+//  and unblocks epoll.
+// Goroutine 2 returns to the caller, makes a new connection.
+// The new connection is accepted by the socket (since it hasn't been destroyed)
+// Goroutine 1 returns from epoll, and has a new connection that it accepts.
+//
+// To avoid this, we block Goroutine 2 from returning till Accept returns
+// to the user, and they get an error.
+type listener struct {
+	l net.Listener
```
abhinav: I think if you embed the listener, you won't have to implement unnecessary methods like `Addr`.

abhinav:
minor nit and missing docs

lgtm otherwise

tnet/listener.go
prashantv: fixed

tnet/listener.go
prashantv: yep, good idea.

abhinav:
The YAML change seems unrelated to this PR. Did you mean to merge it here?

akshayjshah:
7a0e597 shouldn't be part of this PR, I think.

tnet/listener.go
```diff
@@ -0,0 +1,95 @@
...
+	"sync"
+)
+
+// WrapListener returns a new Listener around the provided net.Listener.
+// The returned Listener has a guarantee that when Close returns, it will no longer
+// accept any new connections.
+// See: https://github.com/uber/tchannel-go/issues/141
+func WrapListener(l net.Listener) net.Listener {
+	return &listener{Listener: l, cond: sync.NewCond(&sync.Mutex{})}
+}
```
akshayjshah: (Moved to main PR instead of commit - sorry, getting used to GitHub again.)

Nit: could this function be called `Wrap`? The type system already tells callers that it's wrapping a `net.Listener`.

tnet/listener.go
```diff
@@ -0,0 +1,95 @@
...
+// Any blocked Accept operations will be unblocked and return errors.
+func (s *listener) Close() error {
+	err := s.Listener.Close()
+	if err == nil {
+		s.cond.L.Lock()
+		for s.refs > 0 {
+			s.cond.Wait()
+		}
+		s.cond.L.Unlock()
+	}
```
akshayjshah: This approach lets us have a single return statement, but it doesn't follow the typical approach of [indenting the error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow). Up to you whether it's worth changing to the usual `if err != nil`.

tnet/listener_test.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func connect(addr string) (localAddr string, remoteAddr string, err error) {
+	conn, err := net.Dial("tcp", addr)
+	if err == nil {
+		localAddr = conn.LocalAddr().String()
+		remoteAddr = conn.RemoteAddr().String()
+		conn.Close()
+	}
+	return localAddr, remoteAddr, err
```
akshayjshah: Why return all three values? The test only uses the error.

tnet/listener.go
prashantv: Yep, updated to `Wrap`

tnet/listener.go
prashantv: Since this logic isn't trivial, updated to do the usual `if err != nil`.

tnet/listener_test.go
prashantv: Leftover from debugging, removed now.

prashantv:
Rebased and addressed comments

akshayjshah:
:shipit:

PR #155, opened 2016-01-05
Pool checksum objects

prashantv:
This brings down the number of allocs in BenchmarkCallsSerial from
127 allocs/op to 115 allocs/op.

@akshayjshah


akshayjshah:
:shipit: Looks good. Added #156 to track the farmhash TODO.

PR #157, opened 2016-01-05
Start timers at earliest possible time

prashantv:
Call timeNow as soon as handleCallReq/beginCall are entered.
Use this time to track when a request was started and for adding
annotations to the spans.
This reduces the number of times we call timeNow.

Inbound requests will now include the time it takes to start a goroutine
to process the incoming request. Before, the metrics emitted only
tracked the time from after the goroutine was started to when it
completed writing the response.

@akshayjshah


akshayjshah:
:shipit:

PR #159, opened 2016-01-06
Ensure reserved bytes are always written as 0

prashantv:
The first reserved byte in the frame header was read, and due to frame pooling, could be written out with a non-zero value.

The 8 byte reserved bytes are never read, so they are always 0.

@akshayjshah 
cc: @Raynos

### DIFF
--- a/testutils/testreader/loop.go
+++ b/testutils/testreader/loop.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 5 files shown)


testutils/testreader/loop.go
```diff
@@ -0,0 +1,43 @@
...
+			r.pos = 0
+		}
+	}
+	return len(p), nil
+}
+
+// Looper returns a reader that will return the bytes in bs as if it was a circular buffer.
+func Looper(bs []byte) io.Reader {
+	return &loopReader{bs, 0}
+}
```
akshayjshah: Nifty!

akshayjshah:
lgtm. Just so I'm clear - this is ensuring that [this reserved padding](http://tchannel.readthedocs.org/en/latest/protocol/#reserved8) is maintained while pooling, right? Want to make sure I'm reading the spec correctly.

prashantv:
Basically -- it was just the one earlier. That one was fine with pooling since we never read into it (so it's always just 0), but [this one](http://tchannel.readthedocs.org/en/latest/protocol/#reserved1) was being read, and so with pooling, could end up being set.

PR #162, opened 2016-01-06
Make FailStrategyIgnore retry forever

### LINKED ISSUE #150: Change FailStrategyIgnore to keep retrying forever
Currently `FailStrategyIgnore` will stop trying to advertise after `maxRegistrationFailures`.

We should update it to keep retrying (with some maximum backoff time), similar to the Node implementation.

akshayjshah:
Fixes #150.

### DIFF
--- a/hyperbahn/advertise.go
+++ b/hyperbahn/advertise.go
@@ -79,16 +79,17 @@ func (c *Client) advertiseLoop() {
 
 		if err := c.sendAdvertise(); err != nil {
 			consecutiveFailures++
-			if consecutiveFailures >= maxAdvertiseFailures {
+			if consecutiveFailures >= maxAdvertiseFailures && c.opts.FailStrategy == FailStrategyFatal {
 				c.opts.Handler.OnError(ErrAdvertiseFailed{Cause: err, WillRetry: false})
-				if c.opts.FailStrategy == FailStrategyFatal {
-					c.tchan.Logger().Fatalf("Hyperbahn client registration failed: %v", err)
-				}
-				return
+				c.tchan.Logger().Fatalf("Hyperbahn client registration failed: %v", err)
 			}
 			c.tchan.Logger().Warnf("Hyperbahn client registration failed (will retry): %v", err)
 			c.opts.Handler.OnError(ErrAdvertiseFailed{Cause: err, WillRetry: true})
-			sleepFor = fuzzInterval(advertiseRetryInterval * time.Duration(1<<consecutiveFailures))
+
+			// Even after many failures, cap backoff.
+			if consecutiveFailures < maxAdvertiseFailures {
+				sleepFor = fuzzInterval(advertiseRetryInterval * time.Duration(1<<consecutiveFailures))
+			}
 		} else {
 			c.opts.Handler.On(Readvertised)
 			sleepFor = c.fuzzedAdvertiseInterval()
--- a/hyperbahn/advertise_test.go
+++ b/hyperbahn/advertise_test.go
@@ -261,48 +261,49 @@ func TestRetryTemporaryFailure(t *testing.T) {
 
 func TestRetryFailure(t *testing.T) {
 	runRetryTest(t, func(r *retryTest) {
-		r.mock.On("On", SendAdvertise).Return().
-			Times(1 /* initial */ + maxAdvertiseFailures /* fail */)
 		r.mock.On("On", Advertised).Return().Once()
+		r.mock.On("On", SendAdvertise).Return().Times(1 + (maxAdvertiseFailures * 2))
+
+		// Since FailStrategyIgnore will keep retrying forever, we need to
+		// signal when we're done testing.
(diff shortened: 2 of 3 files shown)


hyperbahn/advertise.go
```diff
@@ -79,16 +79,17 @@ func (c *Client) advertiseLoop() {
...
+			if consecutiveFailures >= maxAdvertiseFailures && c.opts.FailStrategy == FailStrategyFatal {
 				c.opts.Handler.OnError(ErrAdvertiseFailed{Cause: err, WillRetry: false})
-				if c.opts.FailStrategy == FailStrategyFatal {
-					c.tchan.Logger().Fatalf("Hyperbahn client registration failed: %v", err)
-				}
-				return
+				c.tchan.Logger().Fatalf("Hyperbahn client registration failed: %v", err)
 			}
-			c.tchan.Logger().Warnf("Hyperbahn client registration failed (will retry): %v", err)
+			c.tchan.Logger().Infof("Hyperbahn client registration failed (will retry): %v", err)
```
prashantv: Hmm, should we log warnings if Hyperbahn fails to connect after `maxAdvertiseFailures` and when the fail strategy isn't Fatal?

hyperbahn/advertise_test.go
```diff
@@ -261,48 +262,50 @@ func TestRetryTemporaryFailure(t *testing.T) {
...
+		// signal when we're done testing.
+		doneTesting := make(chan struct{}, 1)
+		// For all but the last failure, we just need to assert that the
+		// OnError handler was called. The callback for the last failure needs
+		// to signal that we're done testing and close the Hyperbahn client (so
+		// that the Advertise call returns).
+		expectationCalled := 0
+		r.mock.On("OnError", ErrAdvertiseFailed{true, advertiseErr}).Return(nil).Times(maxAdvertiseFailures * 2).Run(func(_ mock.Arguments) {
+			expectationCalled++
+			fmt.Println(expectationCalled)
```
prashantv: remove print

hyperbahn/advertise.go
akshayjshah: With the current Sentry defaults, I'd say no...but there's no reason to pull that discussion into this diff. I'll go back to the previous behavior and address error/warning levels separately.

hyperbahn/advertise_test.go
```diff
@@ -261,48 +262,50 @@ func TestRetryTemporaryFailure(t *testing.T) {
...
+		// Hyperbahn, FailStrategyIgnore should keep retrying.
+		for i := 1; i <= maxAdvertiseFailures*2; i++ {
 			r.sleepBlock <- struct{}{}
 			r.setAdvertiseFailure()
 			<-r.reqCh
 
-			s1 := <-r.sleepArgs
-			checkRetryInterval(t, s1, i+1 /* retryNum */)
+			sleptFor := <-r.sleepArgs
+			checkRetryInterval(t, sleptFor, i)
```
prashantv: The `i` should be capped here as well (make sure that the test fails if you cap `i` but don't cap in the production code)

akshayjshah:
@prashantv done.

hyperbahn/advertise_test.go
```diff
@@ -261,48 +261,56 @@ func TestRetryTemporaryFailure(t *testing.T) {
...
+		doneTesting := make(chan struct{}, 1)
+		// For all but the last failure, we just need to assert that the
+		// OnError handler was called. The callback for the last failure needs
+		// to signal that we're done testing and close the Hyperbahn client (so
+		// that the Advertise call returns).
+		expectationCalled := 0
+		r.mock.On("OnError", ErrAdvertiseFailed{true, advertiseErr}).Return(nil).Times(maxAdvertiseFailures * 2).Run(func(_ mock.Arguments) {
+			expectationCalled++
+			if expectationCalled >= maxAdvertiseFailures*2 {
+				doneTesting <- struct{}{}
```
prashantv: I know `Close` was being called twice earlier. Is that expected? Do we want to enforce that we only expect that to happen once by making this line `close(dontTesting)`. That'll cause the test to crash if we close it twice.

prashantv:
lgtm with one minor change

hyperbahn/advertise_test.go
akshayjshah: I'll update to close this channel, since that's a good safety practice regardless.

However, but this isn't the channel that was being closed twice. That was expected behavior: the Hyperbahn client is closed once in the meat of the test (so AdvertiseLoop returns even though we never successfully advertise), and once in the shared test teardown code.

If we don't want to alter the Hyperbahn client's Close method so it's safe to call twice, I can tweak the test code to avoid double-closing the client. I'm a bit reluctant to make the testing code even more complex, though.

prashantv: Oh that's right, I forgot about that.

I think the hyperbahn Client.Close being resilient is a good change, so let's keep that.

Looks good, ship it :+1:

Issue #76, opened 2015-11-13
Update logs to avoid formatting strings


prashantv:
Logs should only use `LogFields` to pass parameters -- the messages should be constant strings.

akshayjshah:
I'll take this on tomorrow - I really want Moe's Sentry buckets to be better.

akshayjshah:
Three-step process:

- [ ] Expand the `Logger` interface to support `Debug`, `Info`, et al.
- [ ] Migrate all logging to use fields instead of string substitution.
- [ ] Remove the `Debugf`, `Infof`, et al. from `Logger`.

PR #164, opened 2016-01-07
Temporarily expand Logger interface

akshayjshah:
This is step 1 of #76: it expands the `Logger` interface to support `Debug`, `Info`, and friends. After migrating all logging calls to use these methods, we'll remove the formatting variants.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,12 @@
-*~
-*.swp
-.DS_Store
 build
 Godeps/_workspace
 thrift-gen-release/
+
+# Cover profiles
+*.out
+
+# Editor and OS detritus
+*~
+*.swp
+.DS_Store
 .idea
--- a/logger.go
+++ b/logger.go
@@ -39,20 +39,35 @@ type Logger interface {
 	// Enabled returns whether the given level is enabled.
 	Enabled(level LogLevel) bool
 
-	// Fatalf logs a message, then exits with os.Exit(1)
+	// Fatalf logs a message, then exits with os.Exit(1).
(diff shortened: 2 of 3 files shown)


prashantv:
lgtm

A random thought, I've been thinking about changing the logging interface to make it easier to implement a simple logger. E.g. if we had

```go
type LoggerBase interface {
  IsEnabled(level LogLevel) bool
  Log(level LogLevel, msg string)
  Logf(level LogLevel, msg string, args... interface{})
  WithFields(fields ...LogFields) LoggerBase
}
```

The implementations would have much less API coverage to fill. Within TChannel, we could generate the full interface by wrapping these functions (we can keep the existing interface and have it return that)
```go
type fullLogger struct {
  base LoggerBase
}

func (l fullLogger) Errorf(msg string, args... interface{}) {
  l.base.Logf(LogLevelError, msg, args...)
}

func (l fullLogger) WithFields(fields... LogFields) fullLogger {  
  return fullLogger{l.base.WithFields(fields...)}
}

[...]
```

What do you think?

akshayjshah:
That sounds good - I'd like to leave it until later, though. Like we discussed today, let's experiment a bit to find a performant, easy-to-implement logging approach that we can someday move into go-common.

akshayjshah:
@prashantv Looking at this issue with fresh eyes, migrating *all* logging will cost a ton of allocations, since we don't know the message's log level until *after* we've already eaten the allocations required by `.WithFields`.

Suggest that we leave debug-level logging as-is and only migrate Info and above (and possibly only Error- and Fatal-level logs, since Sentry bucketing matters most there).

PR #167, opened 2016-01-08
Remove high-priority string-munging logger methods

### LINKED ISSUE #76: Update logs to avoid formatting strings

akshayjshah:
Resolves #76.

Migrates all Sentry-able logging to use fields and a constant message, and then removes `logger.Warnf`, `logger.Errorf`, and `logger.Panicf`.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -370,7 +370,10 @@ func (ch *Channel) serve() {
 				if max := 1 * time.Second; acceptBackoff > max {
 					acceptBackoff = max
 				}
-				ch.log.Warnf("accept error: %v; retrying in %v", err, acceptBackoff)
+				ch.log.WithFields(
+					LogField{"error", err.Error()},
+					LogField{"backoff", acceptBackoff},
+				).Warn("Accept error, will wait and retry.")
 				time.Sleep(acceptBackoff)
 				continue
 			} else {
@@ -378,7 +381,9 @@ func (ch *Channel) serve() {
 				if ch.State() >= ChannelStartClose {
 					return
 				}
-				ch.log.Fatalf("unrecoverable accept error: %v; closing server", err)
+				ch.log.WithFields(
+					LogField{"error", err.Error()},
+				).Fatal("Unrecoverable accept error, closing server.")
 				return
 			}
 		}
@@ -393,7 +398,9 @@ func (ch *Channel) serve() {
 		}
 		if _, err := ch.newInboundConnection(netConn, events); err != nil {
 			// Server is getting overloaded - begin rejecting new connections
-			ch.log.Errorf("could not create new TChannelConnection for incoming conn: %v", err)
+			ch.log.WithFields(
+				LogField{"error", err.Error()},
+			).Error("Couldn't create new TChannelConnection for incoming conn.")
 			netConn.Close()
 			continue
 		}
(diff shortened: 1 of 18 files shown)


channel.go
```diff
@@ -370,7 +370,10 @@ func (ch *Channel) serve() {
 				if max := 1 * time.Second; acceptBackoff > max {
 					acceptBackoff = max
 				}
-				ch.log.Warnf("accept error: %v; retrying in %v", err, acceptBackoff)
+				ch.log.WithFields(
+					LogField{"error", err.Error()},
```
prashantv: should we make a helper in the package to create a field for an error,
```go
func errorField(err error) LogField {
  return LogField{"error", err.Error()}
}
```

That way we'll have a consistent key and we can't miss calling `err.Error()`

channel.go
```diff
@@ -370,7 +370,10 @@ func (ch *Channel) serve() {
...
+					LogField{"backoff", acceptBackoff},
+				).Warn("Accept error, will wait and retry.")
```
prashantv: nit: intentional period at end of all log messages? I'm fine either way, we might want to make sure it's consistent within our code and with other projects

prashantv:
A lot of log sites log `frame.Header`. `Header` has a nice `String()` method, but probably won't marshal to json nicely.

Can you take a look at how it is marshalled, and possibly write a custom JSON marshal that includes at least the message type, message ID and frame size.

channel.go
akshayjshah: Sure, that makes sense.

channel.go
akshayjshah: Was intentional on my part - I think of logs as letters to future me. All Warn, Error, and Fatal logs in tchannel-go should now have terminal punctuation.

I haven't noticed a strong convention in go-common re: initial capitals and sentence-style punctuation - does Google recommend a particular logging style? I'm happy to change if necessary.

akshayjshah:
Added an `ErrField` helper and made `FrameHeader` satisfy `json.Marshaler` interface. @prashantv, happy to change log message formatting if you'd like - having a hard time inferring convention from go-common and ringpop-go.

frame.go
```diff
@@ -74,6 +75,15 @@ func (fh FrameHeader) FrameSize() uint16 {
 
 func (fh FrameHeader) String() string { return fmt.Sprintf("%v[%d]", fh.messageType, fh.ID) }
 
+func (fh FrameHeader) MarshalJSON() ([]byte, error) {
+	s := struct {
+		ID      uint32
```
prashantv: nit: should we give this `json` tags with lower case field names

prashantv:
lgtm, happy with the punctuation at the end, was just wondering whether there were opinions in other places but it doesn't look like it.

akshayjshah:
Sorry - didn't see the lower-case JSON tag suggestion before merging, will
fix in a separate PR.

On Fri, Jan 8, 2016 at 2:52 PM Prashant Varanasi <notifications@github.com>
wrote:

> lgtm, happy with the punctuation at the end, was just wondering whether
> there were opinions in other places but it doesn't look like it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/uber/tchannel-go/pull/167#issuecomment-170149603>.
>
-- 
Akshay Shah
Data Engineering
shah@uber.com
(818) 312-4647

Issue #166, opened 2016-01-08
Release a 1.0 of tchannel-go


prashantv:
We should come up with some sort of release process, and release a 1.0 release.

- [x] Remove TODO.md
- [x] Add CONTRIBUTING.md
- [x] Add LICENSE.md
- [x] Add coveralls support (it should be enabled for our repo)
- [x] Add changelog. Started one in `version` branch, https://github.com/uber/tchannel-go/tree/version
- [x] Update version to `1.0`
- [x] Tag release
- [x] Create a binary release (for thrift-gen, there should be Makefile rules to help with that)

mranney:
I suggest that at this point you just declare 1.0, because we are already using this in production.

akshayjshah:
Point taken, issue updated to reference 1.0

PR #168, opened 2016-01-08
Cleanup before 1.0 release

akshayjshah:
Addresses #166.


version.go
```diff
@@ -23,4 +23,4 @@ package tchannel
 // VersionInfo identifies the version of the TChannel library.
 // Due to lack of proper package management, this version string will
 // be maintained manually.
-const VersionInfo = "0.0.1-dev"
+const VersionInfo = "1.0.0"
```
breerly: From the POV of someone using this library, how do they "install" `1.0.0`? Use the correlating commit hash?

akshayjshah: Typically, yes. Some of the build tools, e.g. Glide, also allow pinning to a tag.

breerly: Oh wow - we should def look into that.

prashantv:
lgtm

akshayjshah:
Okay, this should be ready to go now. Coveralls support is limited - the coverage report only covers the `github.com/uber/tchannel-go` package itself, not any of the sub-packages.

We should port over some of `go-build`'s Makefile fanciness to make linting, vetting, and coverage reporting better. I created #171 to track that task.

prashantv:
Sweet, thanks for adding coverage. Yep, we can cover multiple packages coverage in a later change.

:rocket:

PR #170, opened 2016-01-09
Expose thrift.{Read,Write}Headers

prashantv:
Expose thrift header reading and writing functionality so callers can read/write Thrift encoded arguments without using the thrift.Client.

Since it's exposed, `thrift.ReadHeaders` has been rewritten to avoid consuming the whole `io.Reader`.

This is one of the changes required to allow streaming Thrift to be built as a separate library.

cc @jcorbin

### DIFF
--- a/thrift/headers.go
+++ b/thrift/headers.go
@@ -23,7 +23,6 @@ package thrift
 import (
 	"fmt"
 	"io"
-	"io/ioutil"
 
 	"github.com/uber/tchannel-go/typed"
 )
@@ -56,7 +55,8 @@ func WriteHeaders(w io.Writer, headers map[string]string) error {
 
 	// Safety check to ensure the bytes written calculation is correct.
 	if writeBuffer.BytesWritten() != size {
-		return fmt.Errorf("writeHeaders size calculation wrong, expected to write %v bytes, only wrote %v bytes",
+		return fmt.Errorf(
+			"writeHeaders size calculation wrong, expected to write %v bytes, only wrote %v bytes",
 			size, writeBuffer.BytesWritten())
 	}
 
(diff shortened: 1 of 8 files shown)


thrift/headers.go
```diff
@@ -23,13 +23,17 @@ package thrift
...
 )
 
-// TODO(prashant): Use a small buffer and then flush it when it's full.
-func writeHeaders(w io.Writer, headers map[string]string) error {
+// WriteHeaders writes the given key-value pairs using the following encoding:
+// len~2 (k~4 v~4)~len
+func WriteHeaders(w io.Writer, headers map[string]string) error {
+	// TODO(prashant): Since we are not writing length-prefixed data here,
+	// we can write out to the buffer, and if it fills up, flush it.
+	// Right now, we calculate the size of the required buffer and write it out.
```
akshayjshah: :+1:

thrift/headers_test.go
```diff
@@ -41,20 +42,137 @@ var headers = map[string]string{
...
+
+		// Writes iterate over the map in an undefined order, so we might
+		// get expected, or expected2.
+		if bytes.Equal(tt.encoding, buf.Bytes()) ||
+			bytes.Equal(tt.encoding2, buf.Bytes()) {
+			continue
+		}
+
+		assert.Equal(t, tt.encoding, buf.Bytes(), "Unexpected bytes")
+		assert.Equal(t, tt.encoding2, buf.Bytes(), "Unexpected bytes")
```
akshayjshah: This seems like it's going to be confusing to future me - we only expect one of these assertions to be true, right? Can we at least `t.Log` a message to that effect?

typed/reader_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/uber/tchannel-go/testutils/testreader"
+)
+
+var _MyVar = "1"
+
+const cMyConst = "2"
+const gMyConst = "3"
```
akshayjshah: What are these? They're not referenced anywhere, as far as I can see.

prashantv: Was doing unrelated testing about what prefixes golint/go vet complain about. Totally unrelated and accidentally left it in, good catch. Removed.

thrift/headers_test.go
prashantv: Rewrote the logic so it's a little clearer -- do you think it's better without need for a log now?

akshayjshah:
lgtm :shipit:

PR #172, opened 2016-01-11
Update changelog and version

prashantv:


akshayjshah:
:+1: for the changelog updates.

Can we discuss the version-number-in-code approach? If we need to update that with every diff, this is really going to be a pain.

prashantv:
As discussed offline, keeping it a string for now.

PR #173, opened 2016-01-11
WIP: early share work towards experimental streaming support

jcorbin:
Much stil TODO:
- [ ] break out and regression test the partial read fix to fragmentingReader
- [ ] break out and finish the new ErrorHandlerFunc convenience
- [ ] refactor the `withArg2`/`withArg23` helpers added in client.go to be instead
      methods on `tchannel.OutboundCallResponse`
- [ ] refactor the `withArg2`/`withArg23` helpers added in server.go to be instead
      methods on `tchannel.InboundCall`
- [ ] implement response coalescing under the server watch handler
- [ ] implement error/cancellation support in the server watch handler
- [ ] implement finer grained streaming timeout support: the context timeout
  should apply to only the first frame of a stream, with a separate longer
  timeout for the lifetime of the entire streamed request and/or response
  expressed as a request header (and then maybe modified by the initial
  response frame?)

cc @prashantv @nomis52

### DIFF
--- a/arguments.go
+++ b/arguments.go
@@ -143,3 +143,50 @@ func (w ArgWriteHelper) WriteJSON(data interface{}) error {
 		return e.Encode(data)
 	})
 }
+
+type ArgReaderProvider interface {
+	Arg2Reader() (ArgReader, error)
+	Arg3Reader() (ArgReader, error)
+}
+
+// withArg2 calls the provided function with all of arg2 as a byte buffer and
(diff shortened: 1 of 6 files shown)


prashantv:
Closing stale PR against master, filed #466 to fix this issue (and have a branch with a slightly different fix). However, since it doesn't affect any users it's not a high priority issue right now.

PR #174, opened 2016-01-12
Expose thrift.{Read,Write}Struct 

prashantv:
Expose thrift struct reading and writing so callers can read and write Thrift structs as encoded by TChannel without using thrift.Client.

Ånother required to allow streaming Thrift to be built as a separate library.

cc @jcorbin

### DIFF
--- a/testutils/testwriter/limited.go
+++ b/testutils/testwriter/limited.go
@@ -1,0 +1,50 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


testutils/testwriter/limited.go
```diff
@@ -0,0 +1,50 @@
...
+	return writerFunc(func(p []byte) (int, error) {
+		if n < len(p) {
+			retN := n
+			n = 0
+			return retN, ErrOutOfSpace
+		}
+
+		n -= len(p)
+		return len(p), nil
+	})
```
akshayjshah: Fun - tricks with closures! I wouldn't have thought of this.

akshayjshah:
:shipit:

PR #175, opened 2016-01-13
Cleanup and unify ArgReader and ArgWriter interfaces

jcorbin:
- use ArgReader instead of io.ReadCloser
- define both in the arguments.go file

r @prashantv

### DIFF
--- a/arguments.go
+++ b/arguments.go
@@ -31,6 +31,16 @@ import (
 // OutboundCallResponse and an InboundCall
 type ArgReader io.ReadCloser
 
+// ArgWriter is the interface for the arg2 and arg3 streams on an OutboundCall
+// and an InboundCallResponse
+type ArgWriter interface {
+	io.WriteCloser
+
+	// Flush flushes the currently written bytes without waiting for the frame
+	// to be filled.
+	Flush() error
+}
+
 // ArgReadHelper providers a simpler interface to reading arguments.
 type ArgReadHelper struct {
 	reader ArgReader
--- a/arguments.go
+++ b/arguments.go
@@ -27,15 +27,19 @@ import (
 	"io/ioutil"
 )
 
+// ArgReader is the interface for the arg2 and arg3 streams on an
+// OutboundCallResponse and an InboundCall
+type ArgReader io.ReadCloser
+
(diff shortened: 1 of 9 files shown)


jcorbin:
Towards #173

arguments.go
```diff
@@ -27,15 +27,27 @@ import (
 	"io/ioutil"
 )
 
+type ArgReader interface {
```
prashantv: Can you add comments to the exported interfaces here and below. (`go vet` / `golint` should complain about this too, suggest setting those up to run on save)

prashantv:
cc @akshayjshah 

Can you also update the changelog,
https://github.com/uber/tchannel-go/blob/master/CHANGELOG.md

jcorbin:
Added comments

arguments.go
```diff
@@ -27,15 +27,31 @@ import (
...
+
+// ArgWriter is the interface for the arg2 and arg3 streams on an OutboundCall
+// and an InboundCallResponse
+type ArgWriter interface {
+	io.WriteCloser
+
+	// Flush flushes the currently written bytes without waiting for the frame
+	// to be filled.
+	Flush() error
+}
```
akshayjshah: :+1: for keeping these in the same place.

arguments.go
```diff
@@ -27,15 +27,31 @@ import (
 	"io/ioutil"
 )
 
+// ArgReader is the interface for the arg2 and arg3 streams on an
+// OutboundCallResponse and an InboundCall
+type ArgReader interface {
+	io.ReadCloser
+}
```
akshayjshah: Optionally terser as:

```
type ArgReader io.ReadCloser
```

If you're planning to add methods to this interface in a follow-up PR, though, definitely makes sense to keep it as-is.

akshayjshah:
lgtm

arguments.go
prashantv: +1, we have no plans right now to add more methods, so this seems like a good idea.

prashantv:
lgtm

jcorbin:
Simplified the `ArgReader` type

PR #178, opened 2016-01-13
Fix a few go lint exceptions

jcorbin:
Variable shadowing and missing comments.

r @prashantv

### DIFF
--- a/init_test.go
+++ b/init_test.go
@@ -158,8 +158,7 @@ func TestHandleInitReqNewVersion(t *testing.T) {
 // TestHandleInitRes ensures that a Connection is ready to handle messages immediately
 // after receiving an InitRes.
 func TestHandleInitRes(t *testing.T) {
-	l, err := net.Listen("tcp", ":0")
-	require.NoError(t, err, "net.Listen failed")
+	l := newListener(t)
 	listenerComplete := make(chan struct{})
 
 	go func() {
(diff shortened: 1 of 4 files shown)


init_test.go
```diff
@@ -163,15 +163,15 @@ func TestHandleInitRes(t *testing.T) {
 	listenerComplete := make(chan struct{})
 
 	go func() {
-		conn, err := l.Accept()
-		require.NoError(t, err, "l.Accept failed")
+		conn, err2 := l.Accept()
```
prashantv: Prefer not to name this `err2`, I think instead of making this change, the best option might be to have a helper,
```go

func newListener(t *testing.T) net.Listener {
  l, err := net.Listen("tcp", ":0")
  require.NoError(t, err, "Listen failed")
  return l
}
```

That way we don't create a variable `err` before starting the goroutine, and should avoid the shadowing error.

prashantv:
cc @akshayjshah 

It turns out shadowing is currently experimental and not turned on by default,
https://golang.org/cmd/vet/#hdr-Shadowed_variables

So I guess it isn't 100% accurate. I strongly prefer to keep the error variable named `err` and I'm OK with either turning off the shadowing warning in `go vet` or by adding a helper method to avoid the error in the first place.

jcorbin:
Shifted to newListener.

prashantv:
lgtm

akshayjshah:
:shipit:

PR #183, opened 2016-01-14
Add ErrorHandlerFunc

jcorbin:
A simple corollary of HandlerFunc which allows the wrapped function to return
an error and handles returning a minimal error response.  

r @prashantv @akshayjshah

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -21,6 +21,8 @@
 package tchannel
 
 import (
+	"reflect"
+	"runtime"
 	"sync"
 
 	"golang.org/x/net/context"
@@ -40,6 +42,33 @@ type HandlerFunc func(ctx context.Context, call *InboundCall)
 
 // Handle calls f(ctx, call)
 func (f HandlerFunc) Handle(ctx context.Context, call *InboundCall) { f(ctx, call) }
+
+// An ErrorHandlerFunc is an adapter to allow the use of ordinary functions as
+// Channel handlers, with error handling convenience.  If f is a function with
+// the appropriate signature, then ErrorHandlerFunc(f) is a Handler object that
+// calls f.
+type ErrorHandlerFunc func(ctx context.Context, call *InboundCall) error
+
+// Handle calls f(ctx, call)
+func (f ErrorHandlerFunc) Handle(ctx context.Context, call *InboundCall) {
+	if err := f(ctx, call); err != nil {
+		if GetSystemErrorCode(err) == ErrCodeUnexpected {
+			call.log.WithFields(f.getLogFields()...).WithFields(ErrField(err)).Error("Unexpected handler error")
+		}
+		call.Response().SendSystemError(err)
+	}
+}
+
+func (f ErrorHandlerFunc) getLogFields() LogFields {
+	ptr := reflect.ValueOf(f).Pointer()
+	handlerFunc := runtime.FuncForPC(ptr) // can't be nil
+	fileName, fileLine := handlerFunc.FileLine(ptr)
+	return LogFields{
+		{"handlerFuncName", handlerFunc.Name()},
+		{"handlerFuncFileName", fileName},
+		{"handlerFuncFileLine", fileLine},
+	}
+}
 
 // Manages handlers
 type handlerMap struct {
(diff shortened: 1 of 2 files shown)


jcorbin:
Towards #173

prashantv:
Code looks good, can we add a test that uses the ErrorHandlerFunc. `TestServerBusy` in connection_test is an example that tests errors being returned.

jcorbin:
Updated `TestServerBusy` to use; here's an idea for how we could start pivoting the existing `raw.Wrap` utility:
```
diff --git a/raw/handler.go b/raw/handler.go
index 5e92032..9ff7493 100644
--- a/raw/handler.go
+++ b/raw/handler.go
@@ -87,22 +87,15 @@ func WriteResponse(response *tchannel.InboundCallResponse, resp *Res) error {

 // Wrap wraps a Handler as a tchannel.Handler that can be passed to tchannel.Register.
 func Wrap(handler Handler) tchannel.Handler {
-       return tchannel.HandlerFunc(func(ctx context.Context, call *tchannel.InboundCall) {
+       return tchannel.ErrorHandlerFunc(func(ctx context.Context, call *tchannel.InboundCall) error {
                args, err := ReadArgs(call)
                if err != nil {
-                       handler.OnError(ctx, err)
-                       return
+                       return err
                }
-
                resp, err := handler.Handle(ctx, args)
-               response := call.Response()
                if err != nil {
-                       resp = &Res{
-                               SystemErr: err,
-                       }
-               }
-               if err := WriteResponse(response, resp); err != nil {
-                       handler.OnError(ctx, err)
+                       return err
                }
+               return WriteResponse(response, resp)
        })
 }
```

My thoughts are thus:
- `raw.Wrap` and co seem really heavy handed to me
- they don't recompose or reuse well (e.g. to my streaming case)
- allowing a handler to implement custom error handling for `InboundCall` arg reading or `InboundCallResponse` arg writing seems like a pile of YAGNI to me (if not at least a source of non-standard behavior)
- so that means the raw.Wrap handler interface could simplify down to a single function; could well just be this new `ErrorHandlerFunc` plus use of the upcoming `WithArg23` utility (or even `raw.ReadArgs` if we prefer)

Ideally we'd share as much of this sort of code between all as-es (raw, http, thrift, json, etc).

prashantv:
I'm happy with getting rid of `raw.Wrap` as well.

However, `ErrorHandlerFunc` does force how errors are handled (e.g. logging currently) which I don't think is always appropriate.

E.g. this test actually means that a handler returning `ErrServerBusy` would log an error, which is probably not what you want (especially if you're busy and the error path uses more resources).

The tests should have caught this extra log message, but log verification was broken recently by some interface changes. I'm fixing that in #186 which we should rebase this change off, which should cause the test to fail due to an unexpected log message.

handlers.go
```diff
@@ -33,14 +33,31 @@ type Handler interface {
...
+// An ErrorHandlerFunc is an adapter to allow the use of ordinary functions as
+// Channel handlers, with error handling convenience.  If f is a function with
+// the appropriate signature, then ErrorHandlerFunc(f) is a Handler object that
+// calls f.
+type ErrorHandlerFunc func(ctx context.Context, call *InboundCall) error
+
+// Handle calls f(ctx, call)
+func (f ErrorHandlerFunc) Handle(ctx context.Context, call *InboundCall) {
+	if err := f(ctx, call); err != nil {
+		call.log.WithFields(ErrField(err)).Error("Handler error.")
```
prashantv: here and below, we should add the handlerr name (which should be accessible via `call`) so the error log has more information

jcorbin:
Added error log annotate.

Good note about undue logging, I'll add some filtering after that eventual rebase so that we only log certain system errors; at the very least we want unexpected errors (and especially errors that are only coded as unexpected "by default" by not actually being SystemError s).

prashantv:
Merged other PR, so you should be able to rebase. I think only logging unexpected errors make sense.

jcorbin:
Rebased and changed to only log unexpected errors (and to ignore any error returned by `resp.SendSystemError` as the thrift layer does).

prashantv:
code lgtm.

Would be good to also have a similar test with an unexpected error being returned just to have coverage on that branch. too

jcorbin:
Added test case for unexpected handler error

prashantv:
test looks good, lgtm :ship:

PR #186, opened 2016-01-14
Fix log verification in tests, broken by recent Logger interface changes

prashantv:
@akshayjshah


akshayjshah:
:shipit: mea culpa

Issue #176, opened 2016-01-13
Fix golint errors in tchannel package


prashantv:
```
> golint . | grep -v _string.go
frame.go:78:1: exported method FrameHeader.MarshalJSON should have comment or be unexported
logger.go:76:1: exported function ErrField should have comment or be unexported
peer_heap.go:85:31: exported method PopPeer returns unexported type *tchannel.peerScore, which can be annoying to use
```

We may want to add a list of packages we check (or check all of them, and have ignore rules e.g. for _string.go files)

jcorbin:
All except the `peer_heap.go` one were done by #178.

PR #187, opened 2016-01-15
Unexport internal peer constructs like the peer heap

prashantv:
The peer heap is not part of the public interface, and shouldn't be exported. Fixes peer lint issues in #176

cc: @junchaowu


akshayjshah:
:shipit: looks good.

junchaowu:
lgtm

PR #188, opened 2016-01-15
Add a "make lint" and fix golint/go vet/gofmt failures

### LINKED ISSUE #176: Fix golint errors in tchannel package

prashantv:
Will make this part of Travis in a separate PR to ensure we don't have any regressions.

Fixes #176

### DIFF
--- a/pprof/pprof.go
+++ b/pprof/pprof.go
@@ -43,7 +43,7 @@ func Register(registrar tchannel.Registrar) {
 		req, err := thttp.ReadRequest(call)
 		if err != nil {
 			registrar.Logger().WithFields(
-				tchannel.LogField{"err", err.Error()},
+				tchannel.LogField{Key: "err", Value: err.Error()},
 			).Warn("Failed to read HTTP request.")
 			return
 		}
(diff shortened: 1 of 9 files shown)


pprof/pprof.go
```diff
@@ -43,7 +43,7 @@ func Register(registrar tchannel.Registrar) {
 		req, err := thttp.ReadRequest(call)
 		if err != nil {
 			registrar.Logger().WithFields(
-				tchannel.LogField{"err", err.Error()},
+				tchannel.LogField{Key: "err", Value: err.Error()},
```
akshayjshah: I can't really argue with the linters, but this makes the API pretty awful.

akshayjshah:
:shipit:

jcorbin:
:+1:

pprof/pprof.go
prashantv: +1, might need to change the API to just do `.With` where it takes key/value strings or something.

PR #189, opened 2016-01-15
Avoid host (not just host:pair) on retries

### LINKED ISSUE #182: Retries should try to avoid hosts
Currently, retries will avoid `host:port` but it would be better if the ordering was:
1. Select a peer, and try to avoid `host`
2. If no peers are found, select a peer, try to avoid `host:port`
3. If no peers are found, select any peer with no exclusion filters

prashantv:
Fixes #182


akshayjshah:
:shipit:

jcorbin:
:ok: at least it's an unexported interface so we can change it easily going forward.

PR #190, opened 2016-01-15
Update travis to run "make lint"

### LINKED ISSUE #177: Set up linter to run in Travis
We have some lint errors in TChannel right now, see #176. We should update Travis to run these checks.

prashantv:
Fixes #177 

cc @jcorbin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,10 @@ go:
   - 1.4
   - 1.5
 install: make install_ci
-script: make test_ci
-after_success: make cover_ci
+script:
+ - make test_ci
+ - make cover_ci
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,6 +93,7 @@ cover_ci: cover_profile
 
 FILTER := grep -v -e '_string.go' -e '/gen-go/' -e '/mocks/'
 lint:
+	@rm -rf $(GODEPS)
```
akshayjshah: Do we still need this if we're handling it in the Travis config?

prashantv: Nope, removed.

akshayjshah:
looks necessary to me? not good per se, but :shipit:

PR #192, opened 2016-01-16
Add PeerList.Remove for removing a peer

### LINKED ISSUE #184: Add a method of removing a peer.
We'll need this in order to keep the peer list in sync with  ip:ports obtained from DNS.

Alternatively, we could add DNS resolution & re-resolution to the channel library and avoid the need to manage peers in the application.

prashantv:
Resolves #184 

cc @nomis52

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -37,6 +37,9 @@ var (
 	// ErrNoPeers indicates that there are no peers.
 	ErrNoPeers = errors.New("no peers available")
 
+	// ErrPeerNotFound indicates that the specified peer was not found.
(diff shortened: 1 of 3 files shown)


nomis52:
LGTM

akshayjshah:
:shipit:

junchaowu:
Right now which case Go will remove peer?

prashantv:
See #184. It's useful for services that talk to other TChannel services using DNS, and need to maintain the peer list in sync with the DNS entries.

PR #195, opened 2016-01-21
Refactor raw args read helpers

jcorbin:
- refactored `ReadArgsV2` to not use the arg reader helpers
- modified to return any partially read buffer(s) throughout
- factored out half of `ReadArgsV2` into a new helper that could be used to stream arg3 (or otherwise scan, or progressively read arg3 like the thrift and json modules do)

r @prashantv

### DIFF
--- a/raw/call.go
+++ b/raw/call.go
@@ -30,7 +30,7 @@ import (
 // ErrAppError is returned if the application sets an error response.
 var ErrAppError = errors.New("application error")
 
-// ReadArgsV2 reads arg2 and arg3 from a reader.
+// ReadArgsV2 reads arg2 and arg3 from an ArgReadable.
 func ReadArgsV2(r tchannel.ArgReadable) ([]byte, []byte, error) {
 	var arg2, arg3 []byte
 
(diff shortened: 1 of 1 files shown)


prashantv:
+ @akshayjshah 

Given how similar this is to what was already in `raw`, I'm strongly considering whether this should just go into raw. I'm also not sure if it's actually general-purpose enough to belong in the TChannel package, since even JSON directly decodes arg2 into the JSON headers rather than requiring an intermediate buffer.

jcorbin:
The more I think about this code, the more I'm agreeing with that @prashantv; the only point against is the appeal to naturality of a function of a single interface arg being dispatch on the interface instead.  Also the raw "as" is special enough as it is so low level, but... I think you're right and that we should just have these read functions in the raw package, at least for now barring some larger cleanup (say to split out an explicit arguments sub package).

jcorbin:
Okay, re-told the story in terms of an evolution of `raw/call` instead.

prashantv:
@jcorbin code lgtm. Can we open a new PR against the dev branch, apparently the PR can't be changed to be against a new branch.

prashantv:
Closing since this change is against the `master` branch, and we need to do PRs against the `dev` branch.

PR #197, opened 2016-01-22
Better Hyperbahn client retries

### LINKED ISSUE #196: Improve initial advertise behaviour in Hyperbahn client
We should do the following:
 - Increase timeout to a larger value (e.g. 3 seconds)
 - Set TimeoutPerAttempt to 1 second
 - Do explicit retries on Advertise (maybe 2 or 3, depending on the value chosen for Timeout above)

Otherwise, a single RPC timeout can cause services to fail to startup since Advertise failed.

prashantv:
Fixes #196

### DIFF
--- a/hyperbahn/advertise.go
+++ b/hyperbahn/advertise.go
@@ -100,3 +100,21 @@ func (c *Client) advertiseLoop() {
 		}
 	}
 }
+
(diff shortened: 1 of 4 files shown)


akshayjshah:
lgtm

PR #198, opened 2016-01-22
Set timeout when using net.Dial

### LINKED ISSUE #194: net.Dial does not take into account calling context's timeout.
An outbound call from the user can go on past the timeout since `net.Dial` will block the user's request.

We should also test that timeouts are respected in the `init req` path.

prashantv:
If an outbound call required a new connection, the timeout was ignored
when the connection is made (e.g. in net.Dial).

Use net.DialTimeout and set a default timeout to avoid hanging.

Fixes #194

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -524,3 +524,18 @@ func TestGracefulClose(t *testing.T) {
 		})
 	})
 }
+
+func TestNetDialTimeout(t *testing.T) {
+	// timeoutHostPort uses a blackholed address (RFC 6890) with a port
+	// reserved for documentation. This address should always cause a timeout.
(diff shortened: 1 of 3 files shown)


connection_test.go
```diff
@@ -524,3 +524,18 @@ func TestGracefulClose(t *testing.T) {
 		})
 	})
 }
+
+func TestNetDialTimeout(t *testing.T) {
+	// timeoutHostPort uses a blackholed address (RFC 6890) with a port
```
akshayjshah: :+1: for thorough comments.

akshayjshah:
lgtm

PR #199, opened 2016-01-22
Make TestPeerSelectionRanking deterministic

prashantv:
Seed the random number generators so that the test is deterministic
but still verifies that peer selection ranking is not biased.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ get_thrift:
 
 install:
 	GOPATH=$(GODEPS) go get github.com/tools/godep
-	GOPATH=$(GODEPS) godep restore
+	GOPATH=$(GODEPS) godep restore -v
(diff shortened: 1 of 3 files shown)


akshayjshah:
Change looks good.

Do we need to keep testing against Go 1.4? Perhaps we can drop 1.4 support once 1.6 is released?

akshayjshah:
Hm. In the Makefile, may want to change `godep restore` to `godep restore -v`. Travis is pretty picky - it'll kill the build if we don't write something to standard out every 10m.

prashantv:
Once 1.6 is out and we're using it in production, happy to drop 1.4 support. 

Added `-v` to `godep restore`, not sure what's going on with Travis today, but it's been really slow.

PR #202, opened 2016-01-22
Improve Hyperbahn client logging

prashantv:
Don't log warnings as soon as an advertise fails, only log warnings if we've failed more than 5 times.


akshayjshah:
Wat! No tests for logging!!?

lgtm.

PR #203, opened 2016-01-22
Add golint to travis

prashantv:
Travis wasn't really linting due to lack of golint
It was only running go vet and gofmt.


akshayjshah:
lgtm

PR #204, opened 2016-01-27
Split handleCallReq

jcorbin:


jcorbin:
cc @prashantv @akshayjshah this is my idea for how we can add relaying to tchannel-go:
- implement relaying within existing message exchange code path
- change handleCallReq to conditionally create a relaying message exchange...
- ...instead of the current graph of inbound call + message exchange + outbound call + etc

inbound.go
```diff
@@ -42,12 +46,19 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
...
 		panic(fmt.Errorf("unknown connection state for call req: %v", state))
 	}
 
+	return c.handleInboundCall(frame)
+}
+
+// handleInboundCall handles an incoming call request, registering a message
+// exchange to receive further fragments for that call, and dispatching it in
+// another goroutine
+func (c *Connection) handleInboundCall(frame *Frame) bool {
```
akshayjshah: Breaking this function up results in compiler errors:

```
Testing packages:
# github.com/uber/tchannel-go
./inbound.go:94: undefined: now
./inbound.go:113: undefined: now
FAIL	github.com/uber/tchannel-go/stats [build failed]
```

Looks like this helper will need to take `now` as a param.

prashantv:
We should think about whether relaying should be done inside of the tchannel library, or whether it should be done separately. While working on a relay prototype (https://github.com/uber/tchannel-go/tree/go-relay), I ended up having a completely separate path for the relay code, so wasn't sure if we should make the library support it, or build the relay separately.

akshayjshah:
This PR grew out of me asking Josh how the Node library does Hyperbahn relaying, and whether there's an obvious parallel in Go.

I'd also prefer to handle relaying as a separate concern, outside the main TChannel library. With a few small changes to TChannel itself, I think we can implement this as a wrapper. I'll open some PRs/diffs and we can decide which direction we like more.

akshayjshah:
We've made enough progress on the relaying branch that we can close this now.

PR #205, opened 2016-01-29
Support setting routing delegate

akshayjshah:
Modify the ContextBuilder to support setting the [`rd` transport header](http://tchannel.readthedocs.org/en/latest/protocol/#transport-header-rd-routing-delegate).

/cc @abhinav @danqing @prashantv

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -79,7 +79,7 @@ func TestRoutingDelegatePropagates(t *testing.T) {
 		defer cancel()
 		_, arg3, _, err := raw.Call(ctx, ch, peerInfo.HostPort, peerInfo.ServiceName, "test", nil, nil)
 		assert.NoError(t, err, "Call failed")
-		assert.Equal(t, []byte(""), arg3, "Expected no routing delegate header")
+		assert.Equal(t, "", string(arg3), "Expected no routing delegate header")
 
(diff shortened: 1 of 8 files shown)


context_test.go
```diff
@@ -66,6 +66,29 @@ func TestNewContextTimeoutZero(t *testing.T) {
...
+			return &raw.Res{
+				Arg3: []byte(CurrentCall(ctx).RoutingDelegate()),
+			}, nil
+		})
+
+		ctx, cancel := NewContextBuilder(time.Second).Build()
+		defer cancel()
+		_, arg3, _, err := raw.Call(ctx, ch, peerInfo.HostPort, peerInfo.ServiceName, "test", nil, nil)
+		assert.NoError(t, err, "Call failed")
+		assert.Equal(t, []byte(""), arg3, "Expected no routing delegate header")
```
prashantv: nit: second test compares the values as strings, this compares them as byte slices. suggest using strings for both.

prashantv:
lgtm

context_test.go
akshayjshah: Yep, sounds good! Will change & land.

abhinav:
:+1:

danqing:
Thanks @akshayjshah !

Issue #207, opened 2016-01-29
Fix flaky tests


akshayjshah:
Merging PRs occasionally gets challenging, since the test suite is pretty flaky on Travis. Currently problematic flaky tests:

From https://travis-ci.org/uber/tchannel-go/jobs/119372829:
```
--- FAIL: TestCloseSendError-2 (0.21s)
	<autogenerated>:31: [Warn]: Connection error. [{service testService} {process testService-45798} {hostPort 127.0.0.1:45798} {connID 38} {localAddr 127.0.0.1:45798} {remoteAddr 127.0.0.1:40668} {site write frames} {error write tcp 127.0.0.1:40668: connection reset by peer}]
```

From https://travis-ci.org/uber/tchannel-go/jobs/124337096#L3:
```
--- FAIL: TestPeersIncomingConnection-2 (0.02s)
	Error Trace:	test_server.go:212
			test_server.go:174
			test_server.go:134
			test_server.go:90
			verify_utils_test.go:63
			incoming_test.go:88
[...]

PR #208, opened 2016-02-01
Skip TestNetDialTimeout if network is unavailable

prashantv:
See failure in #207. This also happens if the machine has no active network interfaces (e.g. no networks connected), so we should skip the test in these scenarios.


akshayjshah:
lgtm

PR #209, opened 2016-02-01
Fix golint in Travis

prashantv:
The lint package was getting the library instead of the binary, and we were deleting the Godeps workspace which removed the binary.

Make sure "golint" runs on Travis, and fix minor lint issue.

cc @akshayjshah


akshayjshah:
:crying_cat_face: Wow...yep, this seems like it'll help.

breerly:
:+1:

PR #210, opened 2016-02-01
Fix flaky TestReadTimeout, add support for log field filters

prashantv:
Part of #207 

Log filters can now specify field filters as well.

### DIFF
--- a/testutils/channel_opts.go
+++ b/testutils/channel_opts.go
@@ -66,6 +66,9 @@ type LogFilter struct {
 	// Count is the maximum number of allowed warn+ logs matching
 	// Filter before errors are raised.
 	Count uint
+
+	// FieldFilters specifies expected substring matches for fields.
+	FieldFilters map[string]string
 }
 
 // SetServiceName sets ServiceName.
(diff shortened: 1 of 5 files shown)


testutils/channel_opts.go
```diff
@@ -124,10 +127,16 @@ func (o *ChannelOpts) DisableLogVerification() *ChannelOpts {
 
 // AddLogFilter sets an allowed filter for warning/error logs and sets
 // the maximum number of times that log can occur.
-func (o *ChannelOpts) AddLogFilter(filter string, maxCount uint) *ChannelOpts {
+func (o *ChannelOpts) AddLogFilter(filter string, maxCount uint, fields ...string) *ChannelOpts {
```
akshayjshah: I don't love the implicit requirement for an even number of fields, but taking `LogFields` would be really verbose.

akshayjshah:
lgtm

testutils/channel_opts.go
prashantv: Yeah neither. I really wanted to use varargs, and this seems to be the easiest way to avoid verbose call sites. Since it's only a test utility and not part of a public API, I left it in.

PR #211, opened 2016-02-01
Timeout test fixes

prashantv:


connection.go
```diff
@@ -768,13 +768,13 @@ func (c *Connection) checkExchanges() {
...
 		if updated == connectionClosed {
-			go func() {
+			go func(connID uint32) {
 				// We cannot close sendCh until we are sure that there are no other goroutines
 				// that may try to write to sendCh.
 				c.inbound.waitForSendCh()
 				c.outbound.waitForSendCh()
 				close(c.sendCh)
-			}()
+			}(c.connID)
```
akshayjshah: I'm a little confused by this change. Are we just trying to get the connection ID onto the stack for debugging? If that's the case, let's leave a comment.

prashantv: Yep purely for debugging. Added a comment at the top.

akshayjshah: :+1:

akshayjshah:
lgtm

PR #212, opened 2016-02-02
Track all channels created in a process

prashantv:
Expose these channels through introspection for easier debugging

In a future change, we can report errors if a user uses the same channel name twice and hasn't explicitly asked to allow duplicate channels.

### DIFF
--- a/all_channels.go
+++ b/all_channels.go
@@ -1,0 +1,90 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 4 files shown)


all_channels.go
```diff
@@ -0,0 +1,90 @@
...
+	channels := channelMap.existing[ch.ServiceName()]
+	for i, v := range channels {
+		if v != ch {
+			continue
+		}
+
+		// Replace current index with the last element, and truncate channels.
+		channels[i] = channels[len(channels)-1]
+		channels = channels[:len(channels)-1]
+		break
```
akshayjshah: Clever - this is a nice idiom to keep in my back pocket.

introspection.go
```diff
@@ -128,11 +140,41 @@ func (ch *Channel) IntrospectState(opts *IntrospectionOptions) *RuntimeState {
...
+			otherChannelStates = append(otherChannelStates, otherChan.IntrospectOtherState(opts))
+		}
+		states[svc] = otherChannelStates
+	}
+
+	return states
+}
+
+// IntrospectOtherState returns OtherChannelState for a channel.
+func (ch *Channel) IntrospectOtherState(opts *IntrospectionOptions) OtherChannelState {
```
akshayjshah: This name is a bit confusing - `ch.IntrospectOtherState(opts)` reports on the channel's *own* state.

What do you think about this renaming:

`OtherChannelState` -> `ChannelState`
`IntrospectOtherState` -> `ReportState`

Then `IntrospectOthers` is clearly asking each other channel to report its own state.

prashantv: Unfortunately `ChannelState` is already used for the current channel's state enum (closed, active etc). I've renamed it to `ChannelInfo` and `ReportInfo`.

akshayjshah:
lgtm

all_channels_test.go
```diff
@@ -0,0 +1,62 @@
...
+package tchannel
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAllChannelsRegistered(t *testing.T) {
```
breerly: Looks a bit like this should actually be 4 tests, or a table-driven test.

prashantv: This one's tricky because they're not independent tests -- I'm also testing that closing the channels actually affects the state. Because of that, it's really one larger test rather than independent tests

PR #213, opened 2016-02-03
Make removeExchange defensive to avoid panics

prashantv:
Currently, if `removeExchange` is called multiple times for the same `msgID`, it will decrement the `sendCh` wait group multiple times, and this will end up causing a panic.

Update `removeExchange` to be more defensive, and instead log an error if it detects that it's been called multiple times. This way, services will be more resilient and it'll be easier to understand these errors from logs.


akshayjshah:
lgtm

PR #214, opened 2016-02-03
Guard mex shutdown to ensure shutdown is only invoked once

prashantv:
Message exchanges assume unary calls right now. So for an incoming call, the assumption is that the request is read, and then the response is written. Any errors during read will cause a mex shutdown, as will any errors during write.

In the streaming case, writes and reads happen concurrently, and so the reader and writer may both hit an error (the same error in the case of something like timeout or cancelled), and will both try to shutdown the message exchange.

Since the reader and writer are independent, possibly running in separate goroutines, and interacting with different objects (different reader vs writer), synchronizing error state would be complex and slow down the code. Instead, we can protect the message exchange with an atomic value to ensure that only one shutdown will go through.

### DIFF
--- a/mex.go
+++ b/mex.go
@@ -23,6 +23,7 @@ package tchannel
 import (
 	"errors"
 	"sync"
+	"sync/atomic"
 
 	"github.com/uber/tchannel-go/typed"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


akshayjshah:
lgtm too

PR #216, opened 2016-02-04
Repro for a stuck RPC during a streaming call

prashantv:
See commit messages for more information on why this happens. There is no good fix for this issue, so opening this to get some ideas.

Possible ideas:
 - Increase buffer size to something larger to avoid this
   - Doesn't really fix the issue since the remote side could send enough frames to fill the buffer
   - Larger channel means more memory usage for an exchange that will rarely be used
   - We could pool `recvCh` to reduce the impact of ^

- Resizable buffer for `recvCh`
  - Will have to implement our own resizable channel (more code, more likely to have bugs and will need a lot of testing)
  - Complicates the standard RPC flow and may have a performance impact
  - Unbounded memory pressure could have other problems

- Fast path vs slow path (have a separate path for streaming calls)
   - Similar to above, but could avoid impacting the standard RPC flow

- Clients should not make RPC calls in their streaming read handler
   - No performance impact, or code complexity impact in library
   - Simplest fix, but pushes some complexity to the user

cc @akshayjshah

### DIFF
--- a/mex.go
+++ b/mex.go
@@ -40,7 +40,7 @@ const (
 	messageExchangeSetOutbound = "outbound"
 
 	// mexChannelBufferSize is the size of the message exchange channel buffer.
-	mexChannelBufferSize = 2
+	mexChannelBufferSize = 10
 )
 
 // A messageExchange tracks this Connections's side of a message exchange with a
--- a/stream_test.go
+++ b/stream_test.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/uber/tchannel-go/raw"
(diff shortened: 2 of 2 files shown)


akshayjshah:
Interesting - this is a pretty intricate bug.

I'm in favor of the last option, at least for now. As far as I can see, you're right - there's no really nice solution here. Given that, and the fact that we're not officially supporting streaming yet, I'd vote for the fewest possible number of changes (even if that pushes complexity onto streaming users).

At some point, we'll need to rethink how we'd like to support streaming. Since that's likely to require some significant changes and might require a 2.0 release, let's batch it up with any changes needed for YARPC, UNS, and Muttley.

prashantv:
Closing for now, since clients can work around this by not making blocking RPC calls inside of the streaming read handler.

PR #217, opened 2016-02-04
Add relaying

akshayjshah:
This is mostly the same as @prashantv's relay branch, with a few small changes and a few more tests.

I *think* that this should let Muttley write peer selection logic that's shared between the HTTP and TChannel code, wrap it up into a `RelayHosts`, and inject it into a `tchannel.Channel`.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -152,6 +152,7 @@ type Connection struct {
 	nextMessageID   uint32
 	events          connectionEvents
 	commonStatsTags map[string]string
+	relay           *Relayer
 
 	// closeNetworkCalled is used to avoid errors from being logged
 	// when this side closes a connection.
@@ -264,6 +265,9 @@ func (ch *Channel) newConnection(conn net.Conn, initialState connectionState, ev
 	c.inbound.onAdded = c.onExchangeAdded
 	c.outbound.onAdded = c.onExchangeAdded
 
+	if ch.relayHosts != nil {
+		c.relay = NewRelayer(ch, c)
+	}
 	go c.readFrames(connID)
 	go c.writeFrames(connID)
 	return c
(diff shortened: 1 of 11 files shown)


relay.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"errors"
+	"sync"
+)
+
+// RelayHosts allows external wrappers to inject peer selection logic for
+// relaying.
+type RelayHosts interface {
+	// Get returns the host:port of the best peer for the group.
+	Get(group string) string
```
prashantv: is `group` different from `service`/`serviceName`?

connection.go
```diff
@@ -671,41 +675,86 @@ func (c *Connection) readFrames(_ uint32) {
...
 		if releaseFrame {
 			c.framePool.Release(frame)
 		}
 	}
 }
 
+func (c *Connection) readFrameRelay(frame *Frame) bool {
+	releaseFrame := true
+
+	// call req and call res messages may not want the frame released immediately.
```
prashantv: can delete stale comment

connection.go
```diff
@@ -671,41 +675,86 @@ func (c *Connection) readFrames(_ uint32) {
...
-			releaseFrame = c.handleError(frame)
-		default:
-			// TODO(mmihic): Log and close connection with protocol error
-			c.log.WithFields(
-				LogField{"header", frame.Header},
-				LogField{"remotePeer", c.remotePeerInfo},
-			).Error("Received unexpected frame.")
+		var releaseFrame bool
+		if c.relay == nil {
+			releaseFrame = c.readFrameNoRelay(frame)
```
prashantv: maybe `handleFrameNoRelay` / `handleFrameRelay`?

connection.go
```diff
@@ -671,41 +675,86 @@ func (c *Connection) readFrames(_ uint32) {
...
+	switch frame.Header.messageType {
+	case messageTypeCallReq, messageTypeCallReqContinue, messageTypeCallRes, messageTypeCallResContinue:
+		if err := c.relay.Relay(frame); err != nil {
+			c.log.WithFields(
+				LogField{"header", frame.Header},
+				LogField{"remotePeer", c.remotePeerInfo},
+			).Error("Failed to relay frame.")
+		}
+		releaseFrame = false
+	case messageTypeInitReq:
```
prashantv: what do you think about having a `default` that just calls `handleFrameNoRelay` here instead of duplicating these calls

frame.go
```diff
@@ -167,6 +167,20 @@ func (f *Frame) SizedPayload() []byte {
 	return f.Payload[:f.Header.PayloadSize()]
 }
 
+// messageType returns the message type.
+func (f *Frame) messageType() messageType {
+	return f.Header.messageType
+}
+
+// Service returns the name of the destination service.
+func (f *Frame) Service() string {
```
prashantv: this was a quick hack I did, let's make it a little bit safer and easier to read by doing,
```go
const (
  serviceLenIndex = 1 /* flags */ + 4 /* ttl */ + 25 /* tracing */
  serviceIndex = serviceLenIndex + 1
)

if f.messageType() != messageTypeCallReq {
  panic("[...]")
}
serviceLen := f.Payload[serviceLenIndex])
return string(f.Payload[serviceIndex:serviceIndex + serviceLen])
```

frame_test.go
```diff
@@ -135,3 +134,27 @@ func TestReservedBytes(t *testing.T) {
...
+	f := NewFrame(100)
+	fh := fakeHeader()
+	f.Header = fh
+	fh.write(typed.NewWriteBuffer(f.headerBuffer))
+
+	payload := typed.NewWriteBuffer(f.Payload)
+	payload.WriteSingleByte(0)
+	payload.WriteUint32(42)
+	payload.WriteBytes(make([]byte, 25))
+	payload.WriteLen8String("bankmoji")
```
prashantv: instead of creating a payload from scratch, you might want to create a `callReq` and then serialize that to a frame.
e.g.
https://github.com/uber/tchannel-go/blob/dev/init_test.go#L36

relay.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// A Relayer forwards frames.
+type Relayer struct {
+	sync.RWMutex
+	metrics StatsReporter
+	hosts   RelayHosts
+	items   map[uint32]relayItem
+
+	// Immutable
```
prashantv: is this still accurate? i imagine that most things in here are immutable

relay.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Hosts returns the RelayHosts guiding peer selection.
+func (r *Relayer) Hosts() RelayHosts {
+	return r.hosts
+}
+
+// Relay forwards a frame.
+func (r *Relayer) Relay(f *Frame) error {
+	if f.messageType() != messageTypeCallReq {
```
prashantv: can you put a TODO in to remove a relay when we get a `messageTypeCallRes` or `callResContinue` without a continuation bit. Right now, the relayer never removes items I think. It's tricky since it's the remote side that will have to remove items.

We'll also want a TODO to remove relay items on timeout

relay_test.go
```diff
@@ -0,0 +1,48 @@
...
+	server := testutils.NewServer(t, testutils.NewOpts().SetServiceName("test"))
+	defer server.Close()
+	server.Register(raw.Wrap(newTestHandler(t)), "echo")
+	relay.RelayHosts().(*SimpleRelayHosts).Add("test", server.PeerInfo().HostPort)
+
+	client := testutils.NewClient(t, nil)
+	client.Peers().Add(relay.PeerInfo().HostPort)
+	defer client.Close()
+
+	var wg sync.WaitGroup
```
prashantv: this test seems weird -- it doesn't use `wg` and it defers code to run after the test function ends

connection.go
akshayjshah: :+1: good point.

frame.go
akshayjshah: Done.

relay.go
akshayjshah: I'm honestly not quite sure what you meant by "immutable" here - these all seem like bags of mutable state to me.

relay.go
akshayjshah: :+1:

relay_test.go
akshayjshah: Yep, this is completely wrong. Fixing.

akshayjshah:
@prashantv How's this looking now?

connection.go
```diff
@@ -671,41 +675,67 @@ func (c *Connection) readFrames(_ uint32) {
...
-
 		if releaseFrame {
 			c.framePool.Release(frame)
 		}
 	}
 }
 
+func (c *Connection) handleFrameRelay(frame *Frame) bool {
+	switch frame.Header.messageType {
+	case messageTypeCallReq, messageTypeCallReqContinue, messageTypeCallRes, messageTypeCallResContinue:
```
prashantv: let's put a TODO for error frames

relay_test.go
```diff
@@ -0,0 +1,52 @@
...
+		t.Log(string(arg3))
+		require.NoError(t, err, "Relayed call failed.")
+		assert.Equal(t, "fake-header", string(arg2), "Header was mangled during relay.")
+		assert.Equal(t, "fake-body", string(arg3), "Body was mangled during relay.")
+
+		wg.Done()
+	}(client)
+
+	wg.Wait()
+	client.Close()
```
prashantv: optionally: can defer the Close after you create next after they are created.

prashantv:
lgtm

PR #222, opened 2016-02-08
Reduce allocations when reading or writing Thrift structs

prashantv:
Thrift has 2 levels of transports,
 - `TTransport` which is just the `Read` and `Write`
 - `TRichTransport` which is `TTransport` as well as `ReadByte`, `WriteByte`, and `WriteString`.

The protocols use a `TRichTransport`, so when you pass a `TTransport`, it wraps the transport using the `RichTransport` class:
https://godoc.org/github.com/apache/thrift/lib/go/thrift#RichTransport

The `RichTransport` implementations for `ReadByte` and `WriteByte` did allocations on every single call, and since every field requires these functions, it ended up causing a lot of allocations.

To reduce allocations, we implement `TRichTransport` in our internal `readerWriterTransport`, where we have our own `ReadByte` and `WriteByte` implementations that do not do allocations.

The added benchmark results go from:
```
BenchmarkWriteStruct-8   2000000               581 ns/op              96 B/op          6 allocs/op
BenchmarkReadStruct-8    2000000               633 ns/op              96 B/op          6 allocs/op
```
to
```
BenchmarkWriteStruct-8   5000000               351 ns/op               0 B/op          0 allocs/op
BenchmarkReadStruct-8    5000000               394 ns/op              16 B/op          1 allocs/op
```

cc @nomis52 @abhinav

### DIFF
--- a/thrift/transport.go
+++ b/thrift/transport.go
@@ -21,6 +21,7 @@
 package thrift
 
 import (
+	"errors"
 	"io"
 	"sync"
 
@@ -31,7 +32,11 @@ import (
 type readWriterTransport struct {
 	io.Writer
 	io.Reader
+	readBuf  [1]byte
+	writeBuf [1]byte
 }
+
+var errNoBytesRead = errors.New("no bytes read")
 
 func (t *readWriterTransport) Open() error {
 	return nil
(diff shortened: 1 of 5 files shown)


thrift/transport.go
```diff
@@ -49,6 +54,37 @@ func (t *readWriterTransport) Close() error {
...
+		}
+	}
+
+	if err == io.EOF && n > 0 {
+		err = nil
+	}
+	return v[0], err
+}
+
+func (t *readWriterTransport) WriteByte(b byte) error {
```
nomis52: Is this only ever called by one thread? I couldn't find any comments on thread safety.

nomis52:
LGTM

thrift/transport.go
```diff
@@ -49,6 +54,37 @@ func (t *readWriterTransport) Close() error {
...
 }
 
+func (t *readWriterTransport) ReadByte() (byte, error) {
+	v := t.readBuf[0:1]
+
+	var n int
+	var err error
+
+	for {
+		n, err = t.Read(v)
```
abhinav: Is it actually possible for the transport to read zero bytes and not return an error?

prashantv: Since we wrap any `Reader`, it is possible according to the `Reader` interface:
https://golang.org/pkg/io/#Reader

Specifically,
```
Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) == 0. Callers should treat a return of 0 and nil as indicating that nothing happened; in particular it does not indicate EOF.
```
So implementations are discouraged from doing that, but callers should ignore 0 and nil.

thrift/transport.go
prashantv: Yep, since the transport wraps a `Writer`, and you can only have one goroutine writing to a `Writer` at a time, the transport calls are also serialized.

abhinav:
:+1:

blampe:
:sheep:

PR #225, opened 2016-02-16
Release v1.0.3

prashantv:
Merges #208, #209, #210, #211, #212, #213, #214, #219, #220, #222 and #224 from dev into master for a 1.0.3 release.


breerly:
Has this branch-based approach to releasing been working out ok?

version.go
```diff
@@ -23,4 +23,4 @@ package tchannel
 // VersionInfo identifies the version of the TChannel library.
 // Due to lack of proper package management, this version string will
 // be maintained manually.
-const VersionInfo = "1.0.2"
+const VersionInfo = "1.0.3"
```
breerly: How does installing a specific version work? Does glide support that?

prashantv:
Yeah, I think it's been great. We set the default branch to "dev" so our pull request flow etc remains the same. Locally, you have to remember to use the "dev" branch when rebasing, but that's not too hard.

Since we only merge to master on a release, when users `go get` or `git pull` tchannel-go, they only get the released versions. This has a couple of benefits:
 * Users are only using stable released versions
 * When looking at Godeps, the tag will be set to a released version and there's only a few commit hashes to look for.

abhinav:
:+1:

version.go
prashantv: So once I merge this in, I tag the release with "v1.0.3" and glide supports tags to get specific versions.

PR #229, opened 2016-02-19
Hyperbahn client: Move time.Sleep stub into client-specific option

prashantv:
This ensures that:
 - tests cannot impact either each other by stubbing the same sleep
   function
 - external packages can stub the time.Sleep call

### DIFF
--- a/hyperbahn/advertise.go
+++ b/hyperbahn/advertise.go
@@ -41,9 +41,6 @@ const (
 	advertiseRetryInterval = 1 * time.Second
 )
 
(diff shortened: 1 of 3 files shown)


akshayjshah:
lgtm.

Since we're going through the trouble to thread this stub through so many different layers, does it make sense to pass through a more full-featured time stub (e.g., github.com/andres-erbsen/clock)?

prashantv:
Yeah I was thinking about that, but I want to avoid third party dependencies where possible.

PR #230, opened 2016-02-23
Stream unblock readers

aravindvs:
@prashantv  - As we discussed I have a WIP for unblocking readers with an error channel rather than a "bool" channel.. The "outbound" side seems pretty good and the example that @samarabbas added now passes cleanly, if we start both the server and the client and ctrl+C the server (examples/server_crash/).. I am not 100% sure if we need to add more to the "inbound" side..


aravindvs:
@prashantv - btw if we crash the server before we crash the client, we still are blocked because looks like the message exchange is already removed from underneath and we fail with the message exchange not found..

prashantv:
@aravindvs is it possible to rebase this change off the `dev` branch with no streaming changes -- otherwise we won't be able to land this to the `dev` branch. It'll also help with reviewing too.

aravindvs:
@prashantv - will do that and will rebase off of dev.. btw with my latest change, I am now able to successfully run the unit test as well with a longer timeout..

prashantv:
#234 contains the fixes from this PR, closing this PR.

PR #231, opened 2016-02-23
Unblock on close

aravindvs:
@prashantv  - Now we should be against dev..

If we get a connection error, we need to make sure all the exchanges are stopped completely that way we can unblock any application callers.

This patch adds an error channel on an message exchange and makes sure we send the appropriate error code on this channel if we see a connection error on both the inbound and the outbound side.

Also update the unit test to use a longer context timeout and with this change we can run this one cleanly with no issues  (@samarabbas) .

### DIFF
--- a/close_test.go
+++ b/close_test.go
@@ -414,7 +414,6 @@ func TestCloseOneSide(t *testing.T) {
 // TestCloseSendError tests that system errors are not attempted to be sent when
 // a connection is closed, and ensures there's no race conditions such as the error
 // frame being added to the channel just as it is closed.
-// TODO(prashant): This test is waiting for timeout, but socket close shouldn't wait for timeout.
 func TestCloseSendError(t *testing.T) {
 	closed := uint32(0)
 	counter := uint32(0)
@@ -433,14 +432,16 @@ func TestCloseSendError(t *testing.T) {
 	clientCh := testutils.NewClient(t, nil)
 
 	// Create a connection that will be shared.
-	require.NoError(t, testutils.Ping(clientCh, serverCh), "Ping from client to server failed")
+	// Use PingLong here so that we have a context with a bigger timeout
+	require.NoError(t, testutils.PingLong(clientCh, serverCh), "Ping from client to server failed")
 
 	var wg sync.WaitGroup
 	for i := 0; i < 100; i++ {
 		wg.Add(1)
 		go func() {
 			time.Sleep(time.Duration(rand.Intn(1000)) * time.Microsecond)
-			err := testutils.CallEcho(clientCh, serverCh, nil)
+			// Use CallEchoLong here so that we have a context with a bigger timeout
+			err := testutils.CallEchoLong(clientCh, serverCh, nil)
 			if err != nil && atomic.LoadUint32(&closed) == 0 {
 				t.Errorf("Call failed: %v", err)
 			}
--- a/close_test.go
+++ b/close_test.go
@@ -432,16 +432,14 @@ func TestCloseSendError(t *testing.T) {
 	clientCh := testutils.NewClient(t, nil)
 
 	// Create a connection that will be shared.
-	// Use PingLong here so that we have a context with a bigger timeout
-	require.NoError(t, testutils.PingLong(clientCh, serverCh), "Ping from client to server failed")
+	require.NoError(t, testutils.Ping(clientCh, serverCh), "Ping from client to server failed")
 
 	var wg sync.WaitGroup
 	for i := 0; i < 100; i++ {
 		wg.Add(1)
 		go func() {
 			time.Sleep(time.Duration(rand.Intn(1000)) * time.Microsecond)
-			// Use CallEchoLong here so that we have a context with a bigger timeout
-			err := testutils.CallEchoLong(clientCh, serverCh, nil)
+			err := testutils.CallEcho(clientCh, serverCh, nil)
 			if err != nil && atomic.LoadUint32(&closed) == 0 {
 				t.Errorf("Call failed: %v", err)
 			}
--- a/inbound.go
+++ b/inbound.go
@@ -75,7 +75,7 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 
 	// Close may have been called between the time we checked the state and us creating the exchange.
 	if c.readState() != connectionActive {
-		mex.shutdown()
+		mex.shutdown(nil)
 		return true
 	}
 
@@ -219,10 +219,9 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
 		// If we see an error on the exchange, then we should exit immediately
 		case connErr := <-call.mex.errCh:
 			if c.log.Enabled(LogLevelDebug) {
-				c.log.Debugf("Got an error on the exchange: %v (unblock the application)", connErr)
+				c.log.Debugf("Wait for timeout/cancellation interrupted by error: %v", connErr)
 			}
-			call.Response().SendSystemError(connErr)
-			releaseFrame()
+			return
 		}
 	}()
 
@@ -408,6 +407,6 @@ func (response *InboundCallResponse) doneSending() {
 
 	// The message exchange is still open if there are no errors, call shutdown.
 	if response.err == nil {
-		response.mex.shutdown()
+		response.mex.shutdown(nil)
 	}
 }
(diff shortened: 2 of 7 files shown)


aravindvs:
looks like there is a race which i need to fix..

aravindvs:
Adding @thuningxu..

close_test.go
```diff
@@ -433,14 +432,16 @@ func TestCloseSendError(t *testing.T) {
 	clientCh := testutils.NewClient(t, nil)
 
 	// Create a connection that will be shared.
-	require.NoError(t, testutils.Ping(clientCh, serverCh), "Ping from client to server failed")
+	// Use PingLong here so that we have a context with a bigger timeout
```
prashantv: let's remove `PingLong` and `EchoLong` away from testutils and just have one context created with a long timeout here and call them directly. I'd like to avoid `testutils` having methods that are only used by a single call site.

close_test.go
```diff
@@ -433,14 +432,16 @@ func TestCloseSendError(t *testing.T) {
...
+	require.NoError(t, testutils.PingLong(clientCh, serverCh), "Ping from client to server failed")
```
prashantv: you don't care about a long timeout here, since it's a call that returns immediately anyway. This should be reverted back to the `testutils.Ping`

inbound.go
```diff
@@ -211,8 +211,18 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
...
-			call.mex.inboundTimeout()
+		select {
+		case <-call.mex.ctx.Done():
+			if call.mex.ctx.Err() == context.DeadlineExceeded {
+				call.mex.inboundTimeout()
+			}
+		// If we see an error on the exchange, then we should exit immediately
+		case connErr := <-call.mex.errCh:
+			if c.log.Enabled(LogLevelDebug) {
+				c.log.Debugf("Got an error on the exchange: %v (unblock the application)", connErr)
```
prashantv: The current log message is too vague and doesn't make it clear where the error happened.

Change debug log to:
```go
c.log.Debugf("Wait for timeout/cancellation interrupted by error: %v", connErr)
```

inbound.go
```diff
@@ -211,8 +211,18 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
...
+			}
+			call.Response().SendSystemError(connErr)
```
prashantv: If the call was cancelled because of a connection error, does it make sense to send a system error?

inbound.go
```diff
@@ -211,8 +211,18 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
...
+			releaseFrame()
```
prashantv: Is this safe to do? The frame may have already been read and released?

mex.go
```diff
@@ -72,6 +73,9 @@ func (mex *messageExchange) forwardPeerFrame(frame *Frame) error {
 		// Note: One slow reader processing a large request could stall the connection.
 		// If we see this, we need to increase the recvCh buffer size.
 		return GetContextError(mex.ctx.Err())
+	case connErr := <-mex.errCh:
+		// @aravindv: This is to unblock read because we saw a connection error
```
prashantv: Suggest changing this to:
```go
// Unblock readers/writers on connection error and forward the error.
```

we don't tend to have notes include the author's names in the codebase.

mex.go
```diff
@@ -94,6 +98,9 @@ func (mex *messageExchange) recvPeerFrame() (*Frame, error) {
 		return frame, nil
 	case <-mex.ctx.Done():
 		return nil, GetContextError(mex.ctx.Err())
+	case connErr := <-mex.errCh:
+		// @aravindv: This is to unblock read because we saw a connection error
```
prashantv: same as above, update comment.

mex.go
```diff
@@ -326,3 +340,23 @@ func (mexset *messageExchangeSet) forwardPeerFrame(frame *Frame) error {
...
+// stopExchanges goes through all the exhanges and makes sure nobody is waiting
+// because the connection is already being run down
+func (mexset *messageExchangeSet) stopExchanges(err error) error {
+	if mexset.log.Enabled(LogLevelDebug) {
+		mexset.log.Debugf("stopping exchangeset %s", mexset.name)
+	}
+
+	// We need to stop all exchanges on this set
+	mexset.RLock()
+	if mexset.log.Enabled(LogLevelDebug) {
```
prashantv: do not log inside of an rlock

mex.go
```diff
@@ -326,3 +340,23 @@ func (mexset *messageExchangeSet) forwardPeerFrame(frame *Frame) error {
...
+		mexset.log.Debugf("stopping num exchanges: %v", len(mexset.exchanges))
+	}
+	for _, mex := range mexset.exchanges {
+		mex.errCh <- err
```
prashantv: let's move all channel operations outside of the lock

mex.go
```diff
@@ -326,3 +340,23 @@ func (mexset *messageExchangeSet) forwardPeerFrame(frame *Frame) error {
...
+// because the connection is already being run down
+func (mexset *messageExchangeSet) stopExchanges(err error) error {
+	if mexset.log.Enabled(LogLevelDebug) {
+		mexset.log.Debugf("stopping exchangeset %s", mexset.name)
+	}
+
+	// We need to stop all exchanges on this set
+	mexset.RLock()
+	if mexset.log.Enabled(LogLevelDebug) {
+		mexset.log.Debugf("stopping num exchanges: %v", len(mexset.exchanges))
```
prashantv: Avoid 2 logs in the same method, have a single log, and it should be something along the lines of:

```go
mexset.log.Debugf("stopping %v exchanges due to error: %v", numExchanges, err)
```
the name of the mexset should be part of `mexset.log` as a field so you don't need to duplicate it

prashantv:
@aravindvs This change fixes the specific connection error issue, but I want to make sure that the design is correct for the other case -- where on an outbound call, if one side fails (e.g. writing to the stream fails), then the reader should be unblocked too.

Similarly, in an outbound streaming call, if the reader gets an EOF, then the writer should also error out with "response already ended" (you can't maintain a request body for longer than the response body).

I want the same error channel to be applicable to both.

aravindvs:
thanks @prashantv - I will go through the feedback and will address those..

aravindvs:
@prashantv - if i make the error channel to be inside the connection object and update it during the connection close, then it probably will address all scenarios right?

prashantv:
@aravindvs Error channel inside of the connection object will be more complicated since:
 - You'll need to broadcast an error when a connection fails, and channels don't really broadcast events
 - For the case when a single stream needs to close the other side, an error channel in the connection can't be reused -- you'd need something specific on a per-exchange basis.

I think the error channel in the mex like you have is the right design. I think the only thing to figure out is how `stopExchanges` should interact with the existing methods on exchanges.

inbound.go
aravindvs: if i don't send a response here, looks like the call seems to be still alive and is blocked (from the application side).. 
the test we have is based on the streaming branch where we have 2 processes server and client (not the per-process thing we discussed earlier today), crash the client and the server is expected to get an error but it doesn't (if i remove sending the response here)..

mex.go
```diff
@@ -199,6 +208,7 @@ func (mexset *messageExchangeSet) newExchange(ctx context.Context, framePool Fra
 		msgID:     msgID,
 		ctx:       ctx,
 		recvCh:    make(chan *Frame, bufferSize),
+		errCh:     make(chan error, 1),
```
prashantv: Just realized a serious issue with this approach -- if you have both a Writer and a Reader blocked on the `errCh`, and then an error is sent because the connection failed, only *one* of them will be unblocked. We want all readers on `errCh` unblocked (possible reader, writer, as well as the timeout/cancellation wait in inbound).

inbound.go
prashantv: Hm, that needs more debugging then -- sending a system error after a connection is marked as failed is confusing and doesn't really make sense. I'd like to make sure we understand the root problem rather than patching it to make it work.

prashantv:
Used a different approach in #234 where a channel close is used to broadcast an error, but the error is stored separately as a field.

@aravindvs take a look at that PR (which is not 100% yet, some tests to fix still).

prashantv:
Closing as #234 is a superset of this change.

PR #232, opened 2016-02-25
Add go 1.6 and tip to Travis

### LINKED ISSUE #200: Add tip to the build matrix for Travis
Currently, we're only testing against go1.4 and go1.5. We should add `tip` as well to make sure that we're ready for 1.6.

I think there were some test issues last time I tried to do this.

prashantv:
Had to make a few changes to get this working:
 - Fix go vet issues that are new as of 1.6
 - Only run gofmt on 1.5, since gofmt has one small change in 1.6

Fixes #200

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: go
 go:
   - 1.4
   - 1.5
+  - 1.6
+  - tip
 install: make install_ci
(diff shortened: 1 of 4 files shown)


akshayjshah:
lgtm

Issue #151, opened 2016-01-04
Frame pool leaking goroutine through time.Sleep


prashantv:
Seeing failures on travis sometimes due to leaking goroutines in state sleep.

The stack trace shows that the NewSyncFramePool is calling time.Sleep -- which doesn't make sense since there's no calls to sleep in the New function.

```
--- FAIL: TestCloseSingleChannel (0.02s)
	verify.go:121: Found leaked goroutine: Goroutine 191 in state sleep:
		goroutine 191 [sleep]:
		time.Sleep(0x186a0)
			/tmp/workdir/go/src/runtime/time.go:59 +0xf9
		github.com/uber/tchannel-go.NewSyncFramePool.func1(0x0, 0x0)
			/home/travis/gopath/src/github.com/uber/tchannel-go/frame_pool.go:52 +0xab
		sync.(*Pool).Get(0xc820014b00, 0x0, 0x0)
			/tmp/workdir/go/src/sync/pool.go:96 +0x46
		github.com/uber/tchannel-go.syncFramePool.Get(0xc820014b00, 0x8)
			/home/travis/gopath/src/github.com/uber/tchannel-go/frame_pool.go:57 +0x2f
		github.com/uber/tchannel-go.(*syncFramePool).Get(0xc82002e048, 0xc820104600)
			<autogenerated>:205 +0xb4
		github.com/uber/tchannel-go.(*Connection).readFrames(0xc8206ebe00, 0x20)
			/home/travis/gopath/src/github.com/uber/tchannel-go/connection.go:653 +0x74
		created by github.com/uber/tchannel-go.(*Channel).newConnection
			/home/travis/gopath/src/github.com/uber/tchannel-go/connection.go:267 +0x135a
	verify.go:127: Expect at most 2 goroutines, found more:
		[Goroutine 1 in state chan receive:
		goroutine 1 [chan receive]:
		testing.RunTests(0xaea5b0, 0xc8f9a0, 0x6b, 0x6b, 0xcb5e01)
			/tmp/workdir/go/src/testing/testing.go:562 +0xafa
		testing.(*M).Run(0xc82034cf10, 0xc820013d90)
			/tmp/workdir/go/src/testing/testing.go:494 +0xe5
		main.main()
			github.com/uber/tchannel-go/_test/_testmain.go:286 +0x210
		
		 Goroutine 17 in state syscall, locked to thread:
		goroutine 17 [syscall, locked to thread]:
		runtime.goexit()
[...]

PR #234, opened 2016-02-29
Unblock calls on connection errors or if the stream has ended

prashantv:
There are 2 main fixes in this PR:
 * Make sure calls readers/writers are unblocked when a connection is closed (due to EOF or network error)
 * If an outbound stream receives EOF on the response stream, make sure the request stream gets an error

This is similar to #231 but fixes some more corner cases.

cc: @aravindvs @nomis52

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -99,6 +99,13 @@ func (h *testHandler) OnError(ctx context.Context, err error) {
 	h.t.Errorf("testHandler got error: %v stack:\n%s", err, stack)
 }
 
+func writeFlushStr(w ArgWriter, d string) error {
+	if _, err := io.WriteString(w, d); err != nil {
+		return err
+	}
+	return w.Flush()
+}
+
 func TestRoundTrip(t *testing.T) {
 	WithVerifiedServer(t, nil, func(ch *Channel, hostPort string) {
 		handler := newTestHandler(t)
@@ -497,6 +504,40 @@ func TestWriteErrorAfterTimeout(t *testing.T) {
 		<-done
 	})
 	goroutines.VerifyNoLeaks(t, nil)
+}
+
+func TestWriteAfterConnectionError(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	// Closing network connections can lead to warnings in many places.
+	opts := testutils.NewOpts().DisableLogVerification()
+	WithVerifiedServer(t, opts, func(ch *Channel, hostPort string) {
+		testutils.RegisterEcho(ch, nil)
+
+		call, err := ch.BeginCall(ctx, hostPort, ch.ServiceName(), "echo", nil)
+		require.NoError(t, err, "Call failed")
+
+		w, err := call.Arg2Writer()
+		require.NoError(t, err, "Arg2Writer failed")
+		require.NoError(t, writeFlushStr(w, "initial"), "write initial failed")
+
+		// Now close the underlying network connection, writes should fail.
+		_, conn := OutboundConnection(call)
+		conn.Close()
+
+		// Writes should start failing pretty soon.
+		var writeErr error
+		for i := 0; i < 100; i++ {
+			if writeErr = writeFlushStr(w, "f"); writeErr != nil {
+				break
+			}
+			time.Sleep(time.Millisecond)
+		}
+		if assert.Error(t, writeErr, "Writes should fail after a connection is closed") {
+			assert.Equal(t, ErrCodeNetwork, GetSystemErrorCode(writeErr), "write should fail due to network error")
+		}
+	})
 }
 
 func TestReadTimeout(t *testing.T) {
--- a/inbound.go
+++ b/inbound.go
@@ -75,7 +75,7 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 
 	// Close may have been called between the time we checked the state and us creating the exchange.
 	if c.readState() != connectionActive {
-		mex.shutdown(nil)
+		mex.shutdown()
 		return true
 	}
 
@@ -377,6 +377,6 @@ func (response *InboundCallResponse) doneSending() {
 
 	// The message exchange is still open if there are no errors, call shutdown.
 	if response.err == nil {
-		response.mex.shutdown(nil)
+		response.mex.shutdown()
 	}
 }
--- a/inbound.go
+++ b/inbound.go
@@ -75,7 +75,7 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 
 	// Close may have been called between the time we checked the state and us creating the exchange.
 	if c.readState() != connectionActive {
-		mex.shutdown()
+		mex.shutdown(nil)
 		return true
 	}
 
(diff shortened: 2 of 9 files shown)


inbound.go
```diff
@@ -211,8 +211,15 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
...
-		if <-call.mex.ctx.Done(); call.mex.ctx.Err() == context.DeadlineExceeded {
-			call.mex.inboundTimeout()
+		select {
+		case <-call.mex.ctx.Done():
+			if call.mex.ctx.Err() == context.DeadlineExceeded {
+				call.mex.inboundTimeout()
+			}
+		case <-call.mex.errCh.c:
+			if c.log.Enabled(LogLevelDebug) {
+				c.log.Debugf("Wait for timeout/cancellation interrupted by error: %v", call.mex.errCh.err)
```
akshayjshah: I can't wait to get rid of this guard nonsense.

nomis52:
I'm excited to see how this performs.

mex.go
```diff
@@ -72,28 +132,62 @@ func (mex *messageExchange) forwardPeerFrame(frame *Frame) error {
...
-			return nil, errUnexpectedFrameType
+		if err := mex.checkFrame(frame); err != nil {
+			return nil, err
 		}
 		return frame, nil
 	case <-mex.ctx.Done():
 		return nil, GetContextError(mex.ctx.Err())
+	case <-mex.errCh.c:
+		// Select will randomly choose a case, but we want to prioritize
+		// receiving a frame over errCh. Try a non-blockign read.
```
akshayjshah: nit: spelling for "non-blocking"

mex.go
```diff
@@ -326,3 +447,47 @@ func (mexset *messageExchangeSet) forwardPeerFrame(frame *Frame) error {
 
 	return nil
 }
+
+// copyExchanges returns a copy of the exchanges if the exchange is active.
```
akshayjshah: Might be helpful to note in the comment that the mexset should be locked.

mex.go
```diff
@@ -43,28 +46,85 @@ const (
...
+	if err == nil {
+		panic("cannot Notify with no error")
+	}
+
+	// There may be some sort of race where we try to notify the mex twice.
+	select {
+	case <-e.c:
+		return fmt.Errorf("cannot broadcast error: %v, already have: %v", err, e.err)
+	default:
+	}
```
akshayjshah: Does this guard actually protect us from races? It seems possible for concurrent callers to pass through the select statement together and reach line 72 at the same time, potentially over-writing each other.

Perhaps a double-checked lock would be more robust?

mex.go
```diff
@@ -43,28 +46,85 @@ const (
...
 func (mex *messageExchange) forwardPeerFrame(frame *Frame) error {
+	// We want a very specific priority here:
+	// 1. Timeouts/cancellation (mex.ctx errors)
+	// 2. Whether recvCh has buffer space (non-blocking select over mex.recvCh)
+	// 3. Other mex errors (mex.errCh)
+	// Which is why we check the context error only (instead of mex.checkError).
+	// In the mex.errCh case, we do a non-blocking write to recvCh to prioritize it.
 	if err := mex.ctx.Err(); err != nil {
-		return GetContextError(err)
+		return GetContextError(mex.ctx.Err())
```
akshayjshah: The comment above doesn't make it clear to me why we can't `return GetContextError(err)` here. Is this change required for some reason? The context docs specifically mention that successive calls to `Err()` should return the same value.

mex.go
```diff
@@ -72,28 +132,62 @@ func (mex *messageExchange) forwardPeerFrame(frame *Frame) error {
...
 func (mex *messageExchange) recvPeerFrame() (*Frame, error) {
+	// We have to check frames/errors in a very specific order here:
+	// 1. Timeouts/cancellation (mex.ctx errors)
+	// 2. Any pending frames (non-blocking select over mex.recvCh)
+	// 3. Other mex errors (mex.errCh)
+	// Which is why we check the context error only (instead of mex.checkError)e
+	// In the mex.errCh case, we do a non-blocking read from recvCh to prioritize it.
 	if err := mex.ctx.Err(); err != nil {
-		return nil, GetContextError(err)
+		return nil, GetContextError(mex.ctx.Err())
```
akshayjshah: Same as above - why do we need this change?

mex.go
```diff
@@ -199,25 +315,30 @@ func (mexset *messageExchangeSet) newExchange(ctx context.Context, framePool Fra
...
+			mexset.log.WithFields(
+				LogField{"msgID", mex.msgID},
+				LogField{"msgType", mex.msgType},
+			).Warn("Attempted to create new mex after mexset shutdown")
+		} else if addErr == errDuplicateMex {
+			mexset.log.WithFields(
+				LogField{"msgID", mex.msgID},
+				LogField{"newType", mex.msgType},
+			).Warn("Duplicate msg ID for active and new mex.")
+		}
```
akshayjshah: Should we also log any unexpected errors here?

mex.go
```diff
@@ -175,6 +275,7 @@ type messageExchangeSet struct {
 	// maps are mutable, and are protected by the mutex.
 	exchanges        map[uint32]*messageExchange
 	timeoutExchanges map[uint32]struct{}
+	shutdown         bool
```
akshayjshah: The placement suggests to me that we're protecting `shutdown` with the mutex too, but it doesn't look like we synchronize reads at all.

akshayjshah: never mind

mex.go
```diff
@@ -326,3 +447,47 @@ func (mexset *messageExchangeSet) forwardPeerFrame(frame *Frame) error {
...
+		return
+	}
+
+	for _, mex := range exchanges {
+		// When there's a connection failure, we want to notify blocked callers that the
+		// call will fail, but we don't want to shutdown the exchange as only the
+		// arg reader/writer should shutdown the exchange. Otherwise, our guarantee
+		// on sendChRefs that there's no references to sendCh is violated since
+		// readers/writers could still have a reference to sendCh even though
+		// we shutdown the exchange and called Done on sendChRefs.
```
akshayjshah: blinds

mex.go
prashantv: It didn't protect against races very well, it was mostly to ensure we didn't call it twice. But better to do it properly, so updated to use an atomic int (it's less code, and more readable than locks IMO)

mex.go
prashantv: nope, just lots of variations of things led to me leaving this bad code in. Updating.

mex.go
prashantv: fixed, was just copy + paste from the `.Done` branch.

mex.go
prashantv: fixed typo

mex.go
prashantv: So the callers of this already log (specifically `inbound.go` and `connection.go`), these logs are here because we want to add more structured information that the caller won't know.

mex.go
prashantv: :+1:

prashantv:
Updated

mex.go
akshayjshah: Gotcha, makes sense.

connection_test.go
```diff
@@ -499,6 +506,38 @@ func TestWriteErrorAfterTimeout(t *testing.T) {
...
+
+		// Writes should start failing pretty soon.
+		var writeErr error
+		for i := 0; i < 100; i++ {
+			if writeErr = writeFlushStr(w, "f"); writeErr != nil {
+				break
+			}
+			time.Sleep(time.Millisecond)
+		}
+		assert.Error(t, writeErr, "Writes should fail after a connection is closed")
```
prashantv: Updated to check that tchannel error code is `ErrCodeNetwork`. I don't want to get too specific on exact error message, since you could get different errors like `connection reset by peer`, `use of closed network connection` etc.

prashantv:
Travis is failing due to a known issue (#151). Restart in Travis is not working atm, so merging without green build.

PR #235, opened 2016-02-29
SubChannel.SetHandler: override SubChannel Handlers

abhinav:
Added `SubChannel.SetHandler` which makes it possible to set a catch-all
Handler for all requests to a SubChannel. Using `SetHandler()` makes
`Register()` useless and will cause a panic.

As part of this change:

- Removed redundant `handlerMap` from Channel. Always use `SubChannel`'s handler
  instead.
- Changed `handlerMap` to be a map of method name to handler
  (`map[string]Handler`) rather than map of service name to map of method name
  to handler (`map[string]map[string]Handler`). `SubChannel` dispatch based on service name
  happens in `channelHandler` now.
- Made `readableFragment.done()` idempotent.
- Ensure that the read last fragment is released when
  `InboundCallResponse.SendSystemError()` is called.

### DIFF
--- a/subchannel_test.go
+++ b/subchannel_test.go
@@ -127,8 +127,7 @@ func TestSetHandler(t *testing.T) {
 	}
 
 	ch := testutils.NewServer(t, testutils.NewOpts().
-		AddLogFilter("Couldn't find handler", 1, "serviceName", "svc2", "method", "bar"),
-	)
+		AddLogFilter("Couldn't find handler", 1, "serviceName", "svc2", "method", "bar"))
 
 	// Catch-all handler for the main channel that accepts foo, bar, and baz,
 	// and a single registered handler for a different subchannel.
@@ -155,11 +154,8 @@ func TestSetHandler(t *testing.T) {
 
 	for _, tt := range tests {
 		c := client.GetSubChannel(tt.Service)
-
 		ctx, _ := tchannel.NewContext(time.Second)
-		call, err := c.BeginCall(ctx, tt.Method, nil)
-		require.NoError(t, err, "BeginCall failed")
-		_, data, _, err := raw.WriteArgs(call, nil, []byte("irrelevant"))
+		_, data, _, err := raw.CallSC(ctx, c, tt.Method, nil, []byte("irrelevant"))
 
 		if tt.ShouldFail {
 			require.Error(t, err)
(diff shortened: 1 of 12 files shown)


abhinav:
@prashantv @akshayjshah

prashantv:
High-level looks good, but can we split this up into 2 commits:
 - `handlers` and `handlerMap` changes
 - writing up the onDone for the fragment

And then the last thing that'd be useful:
 - update introspection to include information on handlers (if it's a map, mention that it's a map and show what's in it, otherwise, just report that it's a custom handler (and maybe get the type of the custom handler using reflection).

abhinav:
:+1: splitting commits

abhinav:
updated split

prashantv:
reviewed by commit and left comments in the commits.

subchannel_test.go
```diff
@@ -104,3 +109,92 @@ func TestAddReusesPeers(t *testing.T) {
...
+		return tchannel.HandlerFunc(func(ctx context.Context, call *tchannel.InboundCall) {
+			method := call.MethodString()
+			assert.Contains(t, allowedMethods, method, "unexpected call to %q", method)
+			err := raw.WriteResponse(call.Response(), &raw.Res{Arg3: []byte(method)})
+			require.NoError(t, err)
+		})
+	}
+
+	ch := testutils.NewServer(t, testutils.NewOpts().
+		AddLogFilter("Couldn't find handler", 1, "serviceName", "svc2", "method", "bar"),
```
prashantv: nit: trailing comma here is a little strange, usually i just put the trailing `)` on this line too.

subchannel_test.go
```diff
@@ -104,3 +109,92 @@ func TestAddReusesPeers(t *testing.T) {
...
+
+		{"svc2", "foo", false},
+		{"svc2", "bar", true},
+	}
+
+	for _, tt := range tests {
+		c := client.GetSubChannel(tt.Service)
+
+		ctx, _ := tchannel.NewContext(time.Second)
+		call, err := c.BeginCall(ctx, tt.Method, nil)
```
prashantv: you can use `raw.Call` instead of begin call + write args

prashantv:
lgtm

subchannel_test.go
abhinav: :+1:

subchannel_test.go
abhinav: :+1:

PR #240, opened 2016-03-01
Detect if a handler returns a non-nil error type with nil value

### LINKED ISSUE #233: If a Thrift handler returns error with non-nil type but nil value, we send an Empty Result struct over the wire
The generated code for dealing with exceptions looks like this:
```go
err := s.handler.Simple(ctx)
if err != nil {
	switch v := err.(type) {
	case *SimpleErr:
		res.SimpleErr = v
	default:
		return false, nil, err
	}
}
```

Since error interfaces are really 2 values, and comparing to `nil` checks both these values (see https://golang.org/doc/faq#nil_error), we go into the error block. However the casted `v` is `nil` at the point we set it into the result struct. This leads to us sending a result struct no fields set.

We need to improve documentation to make it easier for users to figure this out when they see exceptions from the caller. We can also change the generated code to add an extra check, and return an unknown error instead:

```go
switch v := err.(type) {
  case *SimpleErr:
    if v == nil {
      return false, nil, fmt.Errorf("handler returned non-nil error where type is SimpleErr, but value is nil")
    }
    res.SimpleErr = v
  default:
    return false, nil, err
  }
}
```

cc @norberthu @akshayjshah

prashantv:
@abhinav: fyi, you might want to keep this in mind for yarpc/thriftrw.

Fixes #233

### DIFF
--- a/examples/keyvalue/gen-go/keyvalue/tchan-keyvalue.go
+++ b/examples/keyvalue/gen-go/keyvalue/tchan-keyvalue.go
@@ -120,6 +120,9 @@ func (s *tchanAdminServer) handleClearAll(ctx thrift.Context, protocol athrift.T
 	if err != nil {
 		switch v := err.(type) {
 		case *NotAuthorized:
+			if v == nil {
+				return false, nil, fmt.Errorf("Handler for notAuthorized returned non-nil error type *NotAuthorized but nil value")
+			}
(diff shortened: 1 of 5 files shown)


abhinav:
:+1:

akshayjshah:
:lgtm:

PR #242, opened 2016-03-01
Separate stats test to use cases for success/error

prashantv:
The tests could interfere since the previous inbound call could not have
updated the metrics yet before it receives a new call, which affects the
latency reported.

Addresses #207


akshayjshah:
:shipit:

PR #245, opened 2016-03-02
Clean up relay items when req/res is complete

akshayjshah:
Still needs tests - open just to make sure that this is what @prashantv has in mind.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -38,6 +38,9 @@ const (
 
 	// MaxFramePayloadSize is the maximum size of the payload for a single frame
 	MaxFramePayloadSize = MaxFrameSize - FrameHeaderSize
+
+	_serviceLenIndex  = 1 /* flags */ + 4 /* ttl */ + 25 /* tracing */
+	_serviceNameIndex = _serviceLenIndex + 1
 )
 
 // FrameHeader is the header for a frame, containing the MessageType and size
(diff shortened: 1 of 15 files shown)


frame.go
```diff
@@ -185,6 +186,16 @@ func (f *Frame) Service() string {
 	return string(f.Payload[_serviceNameIndex : _serviceNameIndex+l])
 }
 
+// IsLast indicates whether the frame has a continuation or streaming bit set.
+func (f *Frame) IsLast() bool {
+	if mt := f.messageType(); mt != messageTypeCallReq && mt != messageTypeCallReqContinue && mt != messageTypeCallRes && mt != messageTypeCallResContinue {
```
prashantv: a switch might be nicer here:
```go
switch f.messageType() {
case messageTypeCallReq, messageTypeCallReqContinue [...]:
  // continue
default:
  return false
}
```

frame.go
```diff
@@ -185,6 +186,16 @@ func (f *Frame) Service() string {
...
+		// This message type can't be continued or streamed.
+		return false
+	}
+	flags := f.Payload[_flagsIndex]
+	return (flags == hasMoreFragmentsFlag || flags == isStreamingFlag)
```
prashantv: i think we only care about the `hasMoreFragmentsFlag`. The `streaming` flag is only ever set on call req/call res to say different timeout semantics apply, but streaming calls have to correctly specify `hasMoreFragmentsFlag`

You want to do a bitwise check here too, not just check ` == `
```go
return flags & hasMoreFragmentsFlag != 0
```

PR #250, opened 2016-03-05
Improve GetConnection to avoid allocating a temporary slice

prashantv:
`GetConnection` was creating a temporary slice with all active connections, then picking one of these randomly. Instead, we can iterate through the connections starting at a random index and pick the first active connection we find.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -289,28 +289,52 @@ func (p *Peer) HostPort() string {
 	return p.hostPort
 }
 
-// getActive returns a list of active connections.
-// TODO(prashant): Should we clear inactive connections?
-func (p *Peer) getActive() []*Connection {
-	var active []*Connection
-	p.runWithConnections(func(c *Connection) {
-		if c.IsActive() {
-			active = append(active, c)
-		}
-	})
-	return active
+// getConn treats inbound and outbound connections as a single virtual list
+// that can be indexed. The peer must be read-locked.
+func (p *Peer) getConn(i int) *Connection {
+	inboundLen := len(p.inboundConnections)
+	if i < inboundLen {
+		return p.inboundConnections[i]
+	}
+
+	return p.outboundConnections[i-inboundLen]
 }
 
-func randConn(conns []*Connection) *Connection {
-	return conns[peerRng.Intn(len(conns))]
+func (p *Peer) getActiveConnLocked() (*Connection, bool) {
+	allConns := len(p.inboundConnections) + len(p.outboundConnections)
+	if allConns == 0 {
+		return nil, false
+	}
+
+	// We cycle through the connection list, starting at a random point
+	// to avoid always choosing the same connection.
+	startOffset := peerRng.Intn(allConns)
+	for i := 0; i < allConns; i++ {
+		connIndex := (i + startOffset) % allConns
+		if conn := p.getConn(connIndex); conn.IsActive() {
+			return conn, true
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


peer.go
```diff
@@ -289,28 +289,52 @@ func (p *Peer) HostPort() string {
...
+func (p *Peer) getConn(i int) *Connection {
+	inboundLen := len(p.inboundConnections)
+	if i < inboundLen {
+		return p.inboundConnections[i]
+	}
+
+	return p.outboundConnections[i-inboundLen]
+}
+
+func (p *Peer) getActiveConnLocked() (*Connection, bool) {
```
akshayjshah: Do we need both return values here, or can callers just do a nil check on the Connection?

peer.go
```diff
@@ -381,7 +405,7 @@ func (p *Peer) AddOutboundConnection(c *Connection) error {
 }
 
 // checkInboundConnection will check whether the changed connection is an inbound
-// connection, and will remove any closed connections.
+// connection, and will remove any gclosed connections.
```
akshayjshah: Looks like this was an unintentional change.

peer_bench_test.go
```diff
@@ -0,0 +1,60 @@
...
+	peer := s1.Peers().GetOrAdd(s2.PeerInfo().HostPort)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		peer.GetConnection(ctx)
+	}
+}
+
+func BenchmarkGetConnection0In1Out(b *testing.B) { benchmarkGetConnection(b, 0, 1) }
+func BenchmarkGetConnection1In0Out(b *testing.B) { benchmarkGetConnection(b, 0, 1) }
+func BenchmarkGetConnection5In5Out(b *testing.B) { benchmarkGetConnection(b, 0, 1) }
```
akshayjshah: Oversight: these benchmarks all have the same params.

peer.go
prashantv: One return is enough, but I think it's clearer that something can fail when it returns an ok bool as well as the return item. Otherwise, it's not clear whether the call can return nil or not.

this makes it harder to ignore the `nil` case since you can't do something like `getActiveConn().Close()`

peer.go
prashantv: fixed

peer_bench_test.go
prashantv: good catch, fixed.

prashantv:
Updated, the benchcmp output,
```
> benchcmp bench1.txt bench2.txt
benchmark                           old ns/op     new ns/op     delta
BenchmarkGetConnection0In1Out-8     131           86.2          -34.20%
BenchmarkGetConnection1In0Out-8     122           84.3          -30.90%
BenchmarkGetConnection5In5Out-8     485           93.8          -80.66%

benchmark                           old allocs     new allocs     delta
BenchmarkGetConnection0In1Out-8     1              0              -100.00%
BenchmarkGetConnection1In0Out-8     1              0              -100.00%
BenchmarkGetConnection5In5Out-8     4              0              -100.00%

benchmark                           old bytes     new bytes     delta
BenchmarkGetConnection0In1Out-8     8             0             -100.00%
BenchmarkGetConnection1In0Out-8     8             0             -100.00%
BenchmarkGetConnection5In5Out-8     120           0             -100.00%```

akshayjshah:
:shipit: Looks good - perf improvement is great!

PR #251, opened 2016-03-05
Clean up RelayItems

akshayjshah:
Add some logic to `Frame` and `Relayer` to clean up relay items when the call finishes.

This diff also adds a test simulating a crashed/not-yet-started peer. That test triggers a few race conditions and resource leaks, so it's currently disabled. Our next task is to fix the races and leaks and re-enable the test.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -39,6 +39,7 @@ const (
 	// MaxFramePayloadSize is the maximum size of the payload for a single frame
 	MaxFramePayloadSize = MaxFrameSize - FrameHeaderSize
 
+	_flagsIndex       = 0
 	_serviceLenIndex  = 1 /* flags */ + 4 /* ttl */ + 25 /* tracing */
 	_serviceNameIndex = _serviceLenIndex + 1
 )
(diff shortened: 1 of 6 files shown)


frame.go
```diff
@@ -185,6 +186,18 @@ func (f *Frame) Service() string {
 	return string(f.Payload[_serviceNameIndex : _serviceNameIndex+l])
 }
 
+// isLast indicates whether the frame has the continuation bit set.
+func (f *Frame) isLast() bool {
+	switch f.messageType() {
+	case messageTypeCallReq, messageTypeCallRes, messageTypeCallResContinue, messageTypeCallReqContinue:
+		flags := f.Payload[_flagsIndex]
+		return (flags&hasMoreFragmentsFlag != hasMoreFragmentsFlag)
```
prashantv: nit: don't need parens here

also, normally for bitsets, i just check ` !- 0` or ` == 0`, little bit simpler

frame.go
```diff
@@ -185,6 +186,18 @@ func (f *Frame) Service() string {
...
+	default:
+		// This message type can't be continued or streamed.
```
prashantv: nit: don't mention streaming here

prashantv:
lgtm

akshayjshah:
Nits addressed.

PR #253, opened 2016-03-08
Remove ephemeral peers with closed outbound connections

prashantv:


peer.go
```diff
@@ -399,41 +397,43 @@ func (p *Peer) AddOutboundConnection(c *Connection) error {
...
 }
 
-// checkInboundConnection will check whether the changed connection is an inbound
-// connection, and will remove any closed connections.
-func (p *Peer) checkInboundConnection(changed *Connection) (updated bool, isInbound bool) {
-	newConns := p.inboundConnections[:0]
-	for _, c := range p.inboundConnections {
+// removeClosed will check remove the changed connection if it exists
+// in the specified connections list.
+func (p *Peer) removeClosed(connsPtr *[]*Connection, changed *Connection) (updated bool, found bool) {
```
akshayjshah: Is there any value in protecting ourselves from a nil `connsPtr`? Seems like

```
if connsPtr == nil {
  return false, false
}
```

is easy enough.

peer.go
```diff
@@ -399,41 +397,43 @@ func (p *Peer) AddOutboundConnection(c *Connection) error {
...
-		}
+		updated = true
 	}
 	if updated {
-		p.inboundConnections = newConns
+		*connsPtr = newConns
 	}
 
-	return updated, isInbound
+	return updated, found
```
akshayjshah: I think I'm missing something here - don't we need to append the non-matching conns to `newConns` somewhere?

prashantv: I have no idea how the tests are even passing

peer.go
prashantv: Not too worried about this, if it's nil, I'd be fine with crashing, since it's a pretty bad code bug.

akshayjshah: Also, would this be better named `removeConnection`? I don't see any reason that `changed` needs to be a closed connection.

prashantv: yep, that's true. renaming to `removeConnection`

prashantv:
Fixed and renamed `removeConnection`, totally rewrote it actually to be simpler. Updated now, PTAL

akshayjshah:
:shipit:

PR #255, opened 2016-03-09
Update CHANGELOG for 1.0.4

prashantv:


akshayjshah:
lgtm, since the leaked goroutine obviously wasn't caused by this change.

prashantv:
I wish I could repro leaked goroutines locally, I tried running tests like 1000 times with no luck :\

PR #257, opened 2016-03-10
Clean up RelayItems & handle connection close

akshayjshah:
Count in-progress RPCs on each side of a relayer, and make some attempt to avoid closing connections when RPCs are in progress. This is just step 1 - there are lots of situations that this simple code won't handle.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -775,6 +775,7 @@ func (c *Connection) writeFrames(_ uint32) {
 // checkExchanges is called whenever an exchange is removed, and when Close is called.
 func (c *Connection) checkExchanges() {
 	c.callOnExchangeChange()
+
 	moveState := func(fromState, toState connectionState) bool {
 		err := c.withStateLock(func() error {
 			if c.state != fromState {
@@ -788,6 +789,11 @@ func (c *Connection) checkExchanges() {
 
 	var updated connectionState
 	if c.readState() == connectionStartClose {
+		if c.relay != nil {
+			if !c.relay.canClose() {
+				return
+			}
+		}
 		if c.inbound.count() == 0 && moveState(connectionStartClose, connectionInboundClosed) {
 			updated = connectionInboundClosed
 		}
--- a/connection.go
+++ b/connection.go
@@ -789,10 +789,8 @@ func (c *Connection) checkExchanges() {
 
 	var updated connectionState
 	if c.readState() == connectionStartClose {
-		if c.relay != nil {
-			if !c.relay.canClose() {
-				return
-			}
+		if !c.relay.canClose() {
+			return
 		}
 		if c.inbound.count() == 0 && moveState(connectionStartClose, connectionInboundClosed) {
 			updated = connectionInboundClosed
--- a/relay.go
+++ b/relay.go
@@ -52,21 +52,17 @@ func (r *Relayer) Hosts() RelayHosts {
 func (r *Relayer) Relay(f *Frame) error {
 	// TODO: remove relay items on timeout.
 	if f.messageType() != messageTypeCallReq {
-		r.RLock()
-		item, ok := r.items[f.Header.ID]
-		r.RUnlock()
-		if !ok {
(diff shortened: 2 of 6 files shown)


akshayjshah:
There's a data race here - will address later tonight or first thing tomorrow.

relay.go
```diff
@@ -50,23 +52,26 @@ func (r *Relayer) Hosts() RelayHosts {
...
+	return r.handleCallReq(f)
+}
+
+// Receive accepts a relayed frame.
+func (r *Relayer) Receive(f *Frame) {
+	r.RLock()
+	_, ok := r.items[f.Header.ID]
+	r.RUnlock()
+
+	r.conn.sendCh <- f
```
prashantv: do we need to send this frame if it's not in the `items`?

relay.go
```diff
@@ -99,15 +104,24 @@ func (r *Relayer) Relay(f *Frame) error {
...
+	item, ok := r.items[f.Header.ID]
+	r.RUnlock()
+	if !ok {
+		return errors.New("non-callReq for inactive ID")
+	}
+	f.Header.ID = item.remapID
+	item.destination.Receive(f)
+
+	if finishesCall(f) {
+		r.removeRelayItem(f.Header.ID)
```
prashantv: the `ID` has been modified at this point, so you're removing the wrong ID at this point

relay.go
akshayjshah: Probably not? I wasn't quite sure what to do with `ping req` and `ping res` frames - should we be creating relay items for those too?

relay.go
akshayjshah: Good catch.

relay.go
prashantv: For now, drop the frame (and log, since it shouldn't happen).

Ping req/res also need relay items (since they need ID remapping), let's do that as a separate change though.

relay.go
```diff
@@ -50,26 +52,36 @@ func (r *Relayer) Hosts() RelayHosts {
...
+	r.conn.sendCh <- f
+	if finishesCall(f) {
+		r.removeRelayItem(f.Header.ID)
+	}
+}
+
+func (r *Relayer) handleCallReq(f *Frame) error {
+	r.RLock()
 	if _, ok := r.items[f.Header.ID]; ok {
+		r.RUnlock()
```
prashantv: while I normally try to scope variables in the `if`, I'd prefer to avoid multiple `Unlock` calls. Suggestions on how to fix:
1. 
```go
r.RLock()
_, ok := r.items[f.Header.ID]
r.RUnlock()
if !ok {
  [...]
}
```

2. Break out into a function `func (r *Relayer) hasRelay(id uint32) bool` which gets the `RLock`

connection.go
```diff
@@ -788,6 +789,11 @@ func (c *Connection) checkExchanges() {
 
 	var updated connectionState
 	if c.readState() == connectionStartClose {
+		if c.relay != nil {
```
prashantv: optional: can collapse into a single if and reduce nesting
```go
if c.relay != nil && !c.relay.canClose() {
  return
}
```

prashantv: Or just make `canClose` return `true` for `nil`

relay_test.go
```diff
@@ -83,3 +84,23 @@ func DisabledTestRelayHandlesCrashedPeers(t *testing.T) {
...
+func TestRelayConnectionCloseDrainsRelayItems(t *testing.T) {
+	withRelayedEcho(t, func(relay, server, client *Channel) {
+		ctx, cancel := NewContext(time.Second)
+		defer cancel()
+
+		clientHP := client.PeerInfo().HostPort
+		testutils.RegisterEcho(server, func() {
+			// Handler should close the connection to the relay.
+			conn, err := relay.Peers().GetOrAdd(clientHP).GetConnection(ctx)
+			require.NoError(t, err, "TODO")
```
prashantv: nit: update error message

PR #258, opened 2016-03-11
Fail lint when FIXMEs are left behind

akshayjshah:
We typically use TODOs to track future work. While we're hacking,
though, it's helpful to cut some corners (error messages, comments, test
handling, etc). Now, we can mark those bits of code with a FIXME and be
sure that the branch won't be merged until they're fixed.


prashantv:
lgtm

Issue #152, opened 2016-01-04
Use separate types for atomics


prashantv:
To avoid issues like #143, we should create separate types for atomics to avoid accidentally using them without the atomic operations. E.g.

```go
type AtomicInt32 struct { int32 }

func (i *AtomicInt32) Increment() int32 {
  return i.Add(1)    
}

func (i *AtomicInt32) Add(n int32) {
  return atomic.AddInt32(&i.int32, n)
}

func (i *AtomicInt32) Load() int32() {
  return i.Add(0)
}
```

Since the type will be a struct, it will avoid accidentally incrementing using `++` or passing the value as an argument to a function that expects an `int32` without loading the value atomically.

cc @nomis52

akshayjshah:
@prashantv assuming you'd prefer these to be structs instead of type aliases so we can't mistakenly add to them, right?

prashantv:
Yep, want to make sure you can't use `i++` if `i` is atomic

prashantv:
Fixed by #259

prashantv:
Actually, we should update tchannel-go to use the new package. @akshayjshah

akshayjshah:
Yep, was just taking this in two steps.

PR #259, opened 2016-03-11
Add an atom package with atomic types

akshayjshah:
Rather than using `sync/atomic` directly, create a small library of atomic types. These types don't add any extra functionality, but they prevent unintentional non-atomic use.

If there were any more atomic types, I'd consider templating this code; with only four types and the slight differences between signed and unsigned atomics, I don't think it's worth the added complexity.

@prashantv if this looks okay to you, I'll merge and open a separate PR migrating all of TChannel to these types.

### DIFF
--- a/atom/atom.go
+++ b/atom/atom.go
@@ -1,0 +1,177 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package atom provides simple wrappers around numerics to enforce atomic
+// access.
+package atom
+
+import "sync/atomic"
+
+// Int32 is an atomic wrapper around an int32.
+type Int32 struct{ int32 }
+
+// Load atomically loads the wrapped value.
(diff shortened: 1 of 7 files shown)


atom/stress_test.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+const _parallelism = 4
+const _iterations = 1000
+
+func TestStressInt32(t *testing.T) {
+	defer runtime.GOMAXPROCS(runtime.GOMAXPROCS(_parallelism))
+	for i := 0; i < _parallelism; i++ {
+		go func() {
+			atom := &Int32{0}
```
prashantv: each goroutine has a unique atom, so not sure if the test would catch incorrect use?

akshayjshah:
Fixes #152

atom/atom.go
```diff
@@ -0,0 +1,177 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package atom provides simple wrappers around numerics to enforce atomic
+// access.
+package atom
```
prashantv: everytime I see `atom`, I think of atoms in erlang. if we do name this `atomic` will it be an issue -- since we wouldn't need the standard library when using this package anyway

prashantv:
lgtm, maybe we should just call this atomic

atom/atom.go
Qix-: I would recommend against it, personally. `atomic` is way too commonly used when referencing libraries/namespaces that deal with [atomic operations](https://en.wikipedia.org/wiki/Linearizability). Admittedly I'm new to Go, but generally naming user packages after programming constructs/theory is a Bad Idea&trade;.

"Atom" is better imo, simply because the term is only really used when referring to the Atom editor or (as you mentioned) Erlang's atoms - which appear to simply be a fancy name for constants.

Plus, `atomic` means less to me (personally) as an editor related thing than `atom` does.

atom/atom.go
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
Qix-: It's 2016 :)

atom/atom.go
prashantv: @Qix- This package provides those exact atomic operations that the link you posted describes. It provides types on which you can perform atomic operations.

The reason against naming it `atomic` is because the standard library has a `sync/atomic`, but the point of this package is to wrap `sync/atomic` and avoid use of that package directly. This package is intended to be used internally in tchannel, not for end users. Within tchannel, I don't think we'll want to use this package *and* `sync/atomic`.

Qix-: > but the point of this package is to wrap `sync/atomic` and avoid use of that package directly.

@prashantv Fair enough :+1:

atom/atom.go
akshayjshah: AFAIK, that doesn't matter - this is part of a creative work that was copyrighted in 2015, so that's when the clock started.

E.g., if Walt Disney adds a new brushstroke to Mickey Mouse, that doesn't change the copyright date for Mickey.

akshayjshah: Of course, I'm not a lawyer :)

PR #260, opened 2016-03-17
Get args

IvanMalison:
This PR adds GetArgs to the TChanServer interface which makes it easy to wrap TChanServer instances and intercept the arguments they are receiving for each of their methods in a uniform way.


prashantv:
Hi @IvanMalison, can you give examples of how you expect this to be used?

I'd like to brainstorm to come up with solutions that:
 - don't involve use of empty interfaces, at least not in the generated code
 - similar to above, not have to do type cases.
 - no perf impact
 - Ideally none, but minimal changes to the generated code

If there needs more changes to the generated code, then thrift-gen supports templates, so extra code can be generated if the user wants, and the normal generated code should have just the minimal hook.

Before we get into all this though, I think it's important to understand the use cases properly. I'd prefer to discuss those on an issue rather than on this PR.

prashantv:
We came up with an alternate proposal in #261. Closing PR.

PR #263, opened 2016-03-22

yurishkuro:


retry.go
```diff
@@ -139,7 +139,11 @@ var requestStatePool = sync.Pool{
 }
 
 func getRetryOptions(ctx context.Context) *RetryOptions {
-	opts := getTChannelParams(ctx).retryOptions
+	params := getTChannelParams(ctx)
```
prashantv: prefer this as:
```go
params := getTChannelParams(ctx)
if params == nil {
  return defaultRetryOptions
}

opts := params.retryOptions
if opts == nil {
  return defaultRetryOptions
}
[...]
```

context_header.go
```diff
@@ -36,31 +36,56 @@ type ContextWithHeaders interface {
 	SetResponseHeaders(map[string]string)
 }
 
+type headersContextKey int
```
prashantv: we should be able to reuse `contextKey`
https://github.com/uber/tchannel-go/blob/master/context.go#L31

context_header.go
```diff
@@ -36,31 +36,56 @@ type ContextWithHeaders interface {
...
+}
+
+// headersContainer stores the headers, and is itself stored in the context under `headersKey`
+type headersContainer struct {
 	reqHeaders  map[string]string
 	respHeaders map[string]string
 }
 
+func (c *headerCtx) headers() *headersContainer {
+	if h, ok := c.Value(headersKey).(*headersContainer); ok {
```
prashantv: since the default value is `nil` if it's not found, I'm ok with simplifying this to
```go
return c.Value(headersKey).(*headersContainer)
```

context_header.go
```diff
@@ -36,31 +36,56 @@ type ContextWithHeaders interface {
...
+	return nil
 }
 
 // SetResponseHeaders sets the response headers.
 func (c *headerCtx) SetResponseHeaders(headers map[string]string) {
-	c.respHeaders = headers
+	if h := c.headers(); h != nil {
+		h.respHeaders = headers
+	}
+	// TODO what if we don't find the container?
```
prashantv: this is a tricky one, since it's a serious programming error (they're using a context that tchannel didn't give them or they threw away their context), I'd be ok with a panic here.

prashantv:
Thanks for the change @yurishkuro 

Can you rebase off dev to avoid the merge commits in the history.

Also, let's add some tests to make sure that you can "convert" to a `net.context.Context` and then convert back to a `ContextWithHeaders` and it all works fine. We need some method to convert the `context.Context` to `tchannel.ContextWithHeaders` (or `thrift.Context`)

retry.go
yurishkuro: it didn't seem DRY to me as it falls back on default twice

context_header.go
yurishkuro: I saw it, but the existing constant was explicitly initialized to `1`, so there is a chance someone might crease a clash. But if we can move the new constant to the same file and use the `iota` pattern, then it's safe.

yurishkuro:
@prashantv  Yeah, I'll rebase and squish the commits before merging. But I am also making other changes needed to bridge the Span, not just the context. Or do you prefer that to be in separate PR?

BTW, any idea what's wrong with the build?

```
GOPATH=/home/travis/gopath/src/github.com/uber/tchannel-go/Godeps/_workspace go get github.com/tools/godep
GOPATH=/home/travis/gopath/src/github.com/uber/tchannel-go/Godeps/_workspace go get github.com/golang/lint/golint
package github.com/golang/lint/golint
	imports go/types: unrecognized import path "go/types"
package github.com/golang/lint/golint
	imports golang.org/x/tools/go/gcimporter15
	imports golang.org/x/tools/go/gcimporter15
	imports golang.org/x/tools/go/gcimporter15: no buildable Go source files in /home/travis/gopath/src/github.com/uber/tchannel-go/Godeps/_workspace/src/golang.org/x/tools/go/gcimporter15
make: *** [install] Error 1
```

yurishkuro:
replaced by https://github.com/uber/tchannel-go/pull/267, merging to `master` instead of `dev`

PR #264, opened 2016-03-22
Run linters only on Go 1.5.x and 1.6.x

akshayjshah:
With some Make-fu, we can run the linters only if we're using Go 1.5.x or Go 1.6.x. This allows us to continue running the test suite on Go 1.4 and tip.

This diff also expands our lint filters to hide errors in dependencies, which should let us remove the hack that deletes vendored code before linting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,18 +95,18 @@ cover_ci: cover_profile
 	goveralls -coverprofile=$(BUILD)/coverage.out -service=travis-ci
 
 
-FILTER := grep -v -e '_string.go' -e '/gen-go/' -e '/mocks/'
+FILTER := grep -v -e '_string.go' -e '/gen-go/' -e '/mocks/' -e 'Godeps/' -e 'vendor/'
 lint:
 ifneq ($(filter $(LINTABLE_MINOR_VERSIONS),$(GO_MINOR_VERSION)),)
 	@echo "Linters are enabled on Go version" $(GO_VERSION)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -93,17 +100,23 @@ cover_ci: cover_profile
...
+ifneq ($(filter $(LINTABLE_MINOR_VERSIONS),$(GO_MINOR_VERSION)),)
+	@echo "Linters are enabled on Go version" $(GO_VERSION)
 	@echo "Running golint"
 	-golint ./... | $(FILTER) | tee lint.log
 	@echo "Running go vet"
-	-go tool vet $(PKGS) 2>&1 | tee -a lint.log
+	-go tool vet $(PKGS) 2>&1 | $(FILTER) | tee -a lint.log
 	@echo "Checking gofmt"
-	-[ $(GO_VERSION) != "go1.5" ] || gofmt -d . | tee -a lint.log
+	-[ $(GO_MINOR_VERSION) != $(FMTABLE_MINOR_VERSION) ] || gofmt -d . | tee -a lint.log
```
prashantv: do you think we should use a makefile `if` for this as well and echo that we're not fmt'ing?

akshayjshah: Probably, yes :)

akshayjshah:
Triumph! Take that, Richard Stallman.

prashantv:
lgtm

PR #265, opened 2016-03-22
Never select peers with no active connections

robskillington:
cc @prashantv @akshayjshah


prashantv:
I think this change will break existing TChannel services that rely on a random peer being selected if there are no existing connections. The current flow for most service is:
 - Add Hyperbahn peers using the `hosts.json` file
 - Hyperbahn client calls `Advertise`, which will connect to a random outgoing node
 - After receiving the Advertise, Hyperbahn will then make incoming connections to the TChannel service

So we need to select unconnected peers if there's no connections. The score calculator already has the following priority order:
 - Peers with incoming connections
 - Peers with outgoing connections
 - Unconnected peers

See: https://github.com/uber/tchannel-go/blob/dev/peer_strategies.go#L68

Can you explain the use case you're trying to make work, I think it belongs in creating a custom `ScoreCalculator`, but want to make sure.

akshayjshah:
`ScoreCalculator` is in the middle of a small refactoring to make atomic ops safer; you may want to wait until #269 lands before introducing a new calculator.

robskillington:
@prashantv so that logic means no hosts can be down.

We have the case where sometimes our configs/peers are slightly out of date with what is in production and we want to never ever load balance to hosts that are just always down.  With the current logic a small percentage of requests will always go to those down nodes.

We need to find a way around this.

prashantv:
Why does it imply that hosts can never be down? The client library does automatic retries, and it'll avoid retrying hosts that have previously been tried.

robskillington:
Workaround for what we were seeing is removing peers with previously closed connections now included in 1.0.4 with merge commit: https://github.com/uber/tchannel-go/commit/ec7bdc597ed55688bb9a934dcf6da7e1a340577a

PR #267, opened 2016-03-22

yurishkuro:


prashantv:
Merge to dev is actually what we want. We do PRs against dev, and then we bump version and release to master every couple of weeks or so.

context_test.go
```diff
@@ -119,3 +119,26 @@ func TestCurrentCallWithNilResult(t *testing.T) {
...
+		Build()
+	assert.EqualValues(t, "header value", t_ctx2.Headers()["header key"])
+	assert.EqualValues(t, "some value", t_ctx2.Value("some key"), "inherited from parent ctx")
+
+	t_ctx2.SetResponseHeaders(map[string]string{"resp key": "resp value"})
+	assert.EqualValues(t, "resp value", t_ctx2.ResponseHeaders()["resp key"])
+
+	ctx = t_ctx2 // test as regular context
+	assert.EqualValues(t, "some value", ctx.Value("some key"))
+}
```
prashantv: couple of things I want to see tested.

 - given an existing context with "header1", when you do `NewContextBuilder(..).SetParentCtx(existingCtx).AddHeader"header2", "v2").Guild()`, the new context should have both "header 1" and "header 2".
 - Similarly, I think it's worth adding a method `Wrap` which takes a `net.Context` and returns `ContextWithHeaders` perhaps?

yurishkuro: Interesting point for (1), I think it's doable. The assumption is that if you `AddHeader` with an existing key, you override the previous value, correct?

For (2), if you think it's useful, I can add it. It will be equivalent to `WrapWithHeaders(parent, nil)`.

prashantv: Yep, existing keys should override, but previous keys should be saved by default.

For 2, the main difference is how  `WrapWithheaders(parent, nil)` will handle existing headers. I would imagine `Wrap(parent)` to always keep the parent headers, while `WrapWithHeaders(parent, nil)` to replace any existing headers in the parent with `nil`.

yurishkuro: That semantics for (2) is a bit problematic, because ContextBuilder is calling WrapWithHeaders, and I am defining its parent context like this:

```go
	// ParentContext to build the new context from. If empty, context.Background() is used.
	// If ParentContext is of type ContextWithHeaders, then parent headers will be carried
	// over to the child context, unless overwritten via `AddHeader()`.
	ParentContext context.Context
```

PR #268, opened 2016-03-22

yurishkuro:


context_header.go
```diff
@@ -38,29 +38,48 @@ type ContextWithHeaders interface {
...
+
+// headersContainer stores the headers, and is itself stored in the context under `contextKeyHeaders`
+type headersContainer struct {
 	reqHeaders  map[string]string
 	respHeaders map[string]string
 }
 
+func (c *headerCtx) headers() *headersContainer {
+	return c.Value(contextKeyHeaders).(*headersContainer)
+}
```
yurishkuro: This didn't work
```
panic: interface conversion: interface {} is nil, not *tchannel.headersContainer
```
changing back to a checked conversion.

context_header.go
```diff
@@ -38,29 +38,65 @@ type ContextWithHeaders interface {
...
+			mergedHeaders := make(map[string]string, len(headers) + pl)
+			for k, v := range parent_h.reqHeaders {
+				mergedHeaders[k] = v
+			}
+			for k, v := range headers {
+				mergedHeaders[k] = v
+			}
+			h.reqHeaders = mergedHeaders
+		}
+	}
```
yurishkuro: @prashantv new semantics you requested, parent headers are merged with new ones.

prashantv: Sorry for the back and forth, but I don't think this is the way we want to solve this issue since it becomes really hard to override all headers (e.g. you want to drop the existing ones and add new ones).

So we have a few use cases, and I have code for what I think each of those should look like:

 *  Service gets incoming call, wants to carry over headers, and add a new header
```go
ctx := NewContextBuilder(...)
  .SetParentContext(ctx)
  .AddHeader("extra", "value")
  .Build()
```

Headers are merged because of use of "AddHeader". If the user uses "SetHeaders" that should replace. Unless there's a "merge" in the name, I think previous headers should be replaced.

 * Service gets incoming call, and wants to propagate them all
```go
// ctx can be directly used
```

 * Service gets incoming call, wants to drop all existing headers, and set a new bunch of headers
```go
ctx = WrapWIthHeaders(ctx, map[string]string{"new": "headers"})
```


(we should have separate tests for these use cases and make sure it works fine)

yurishkuro: Suggestion:

Assume parent headers are `{a: b, k: v}`

(1) keep parent headers and add new ones:
```
NewContextBuilder().SetParentContext(parentCtx).SetHeaders({x: y, a: c}).Build()
```
Result: `{a: c. k: v, x: y}`, i.e. merged new headers into parent headers, overriding key `a`

(2) keep parent headers unchanged:
```
NewContextBuilder().SetParentContext(parentCtx).SetHeaders({x: y, a: c}).Build()
```

(3) ignore parent headers - presumably less frequent use case, so done by clearing parent first
```
parentWithoutHeaders := WrapWithHeaders(parent, nil)
child := NewContextBuilder().SetParentContext(parentWithoutHeaders).SetHeaders({x: y, a: c}).Build()
```

yurishkuro: Note: no change to Set / Add semantics from what exists today, they only affect headers accumulated in the Builder, not what happens with parent headers.

prashantv: I'm still against `Set` not actually setting, but merging with previous headers. My expectation is:

`SetParentContext(parentCtx).SetHeaders(headers)` overwrites all previous headers.

`Set` shouldn't be merging, that's what `Add` does. From my perspective:
`SetParentContext(parentCtx)` is the same as `NewContextBuilder(..).SetSpan(..).SetHeaders(parentCtx.Headers)`. There's no special "parent context mode" in the builder, it's just inheriting a lot of things by default when you set the parent context.

yurishkuro: > There's no special "parent context mode" in the builder

Actually, there is, because the parent context is used to create the child context. So, for example, all `Value`s from the parent context become available on the child. But ok, I can make the act of calling `SetHeaders` trigger an "ignore parent headers" mode internally.

One thing I missed is that we should probably also inherit a span from the parent context, unless `SetExternalSpan` is called.

prashantv: +1 for the spans. Let's also add unit tests for everything to make sure it works the way we expect in all the cases.

context_test.go
```diff
@@ -119,3 +119,86 @@ func TestCurrentCallWithNilResult(t *testing.T) {
...
+		SetParentContext(ctx2).
+		Build()
+	assert.Equal(t, &span, CurrentSpan(ctx3), "span inherited from parent")
+
+	ctx4, _ := NewContextBuilder(time.Second).
+		SetParentContext(ctx2).
+		SetExternalSpan(3, 2, 1, true).
+		Build()
+	span4 := NewSpan(3, 2, 1)
+	assert.Equal(t, &span4, CurrentSpan(ctx4), "span inherited from parent")
```
prashantv: nit: message needs to be updated

utils_for_test.go
```diff
@@ -69,6 +69,6 @@ func InboundConnection(call IncomingCall) (*Connection, net.Conn) {
 }
 
 // NewSpan returns a Span for testing.
-func NewSpan(traceID uint64, parentID uint64, spanID uint64) Span {
-	return Span{traceID: traceID, parentID: parentID, spanID: spanID, flags: defaultTracingFlags}
+func NewSpan(traceID uint64, spanID uint64, parentID uint64) Span {
```
prashantv: any reason for this change? i kind of like the ordering of trace -> parent -> span (global -> local).

context_test.go
```diff
@@ -119,3 +119,86 @@ func TestCurrentCallWithNilResult(t *testing.T) {
...
+	ctx1, _ := NewContextBuilder(time.Second).
+		SetParentContext(ctx).
+		AddHeader("header key", "header value").
+		Build()
+	return ctx1
+}
+
+func TestContextBuilderParentContextNoHeaders(t *testing.T) {
+	ctx := getParentContext(t)
+	assert.EqualValues(t, "header value", ctx.Headers()["header key"])
```
prashantv: here and below, compare the entire map, and use `assert.Equal`, shouldn't need `EqualValues` (that's only for type mismatches).

```go
assert.Equal(t, map[string]string{"header key": "header value"}, ctx.Headers())
```

prashantv:
Thanks for adding the tests. The code looks good, just some minor test cleanup:
 - use `assert.Equal` over `EqualValues`
 - compare entire map rather than individual values.

utils_for_test.go
yurishkuro: Don't have a strong opinion. In the Jaeger's http header the order is `trace:span:parent:flags`. In fact, many people argue that it is not even necessary to propagate `parent` across RPC, it was enough to record it out of band.

prashantv:
lgtm, feel free to merge. Thanks for making this change!

PR #269, opened 2016-03-22
Use our wrappers for all atomic operations

### LINKED ISSUE #152: Use separate types for atomics

akshayjshah:
This PR makes it harder for us to unwittingly introduce races. Rather than using `sync/atomic` directly, which requires remembering which variables are atomic and which aren't, use `tchannel-go/atomic` for all atomic operations.

Fixes #152.

### DIFF
--- a/close_test.go
+++ b/close_test.go
@@ -23,11 +23,11 @@ package tchannel_test
 import (
 	"math/rand"
 	"sync"
-	"sync/atomic"
 	"testing"
 	"time"
 
 	. "github.com/uber/tchannel-go"
+	"github.com/uber/tchannel-go/atomic"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -415,15 +415,17 @@ func TestCloseOneSide(t *testing.T) {
 // a connection is closed, and ensures there's no race conditions such as the error
 // frame being added to the channel just as it is closed.
 func TestCloseSendError(t *testing.T) {
-	closed := uint32(0)
-	counter := uint32(0)
+	var (
+		closed  atomic.Uint32
+		counter atomic.Uint32
+	)
 
 	serverCh := testutils.NewServer(t, nil)
 	testutils.RegisterEcho(serverCh, func() {
-		if atomic.AddUint32(&counter, 1) > 10 {
+		if counter.Inc() > 10 {
 			// Close the server in a goroutine to possibly trigger more race conditions.
 			go func() {
-				atomic.AddUint32(&closed, 1)
+				closed.Inc()
 				serverCh.Close()
 			}()
 		}
@@ -440,7 +442,7 @@ func TestCloseSendError(t *testing.T) {
 		go func() {
 			time.Sleep(time.Duration(rand.Intn(1000)) * time.Microsecond)
 			err := testutils.CallEcho(clientCh, serverCh, nil)
-			if err != nil && atomic.LoadUint32(&closed) == 0 {
+			if err != nil && closed.Load() == 0 {
 				t.Errorf("Call failed: %v", err)
 			}
 			wg.Done()
--- a/connection.go
+++ b/connection.go
@@ -28,10 +28,11 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
+	"github.com/uber/tchannel-go/atomic"
 	"github.com/uber/tchannel-go/typed"
+
 	"golang.org/x/net/context"
 )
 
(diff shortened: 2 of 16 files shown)


close_test.go
```diff
@@ -415,15 +415,15 @@ func TestCloseOneSide(t *testing.T) {
 // a connection is closed, and ensures there's no race conditions such as the error
 // frame being added to the channel just as it is closed.
 func TestCloseSendError(t *testing.T) {
-	closed := uint32(0)
-	counter := uint32(0)
+	closed := &atomic.Uint32{}
```
prashantv: i'd like both of these to just be
```go
var (
  counterr atomic.Uint32
  closed atomic.Uint32
)
```

or if we want to be explicit about the 0 value, keep it similar to what we have now
```
closed := atomic.NewUint32(0)
counter := atomic.NewUint32(0)
```

since we're not passing them anywhere, we should't need a pointer

connection.go
```diff
@@ -149,19 +150,19 @@ type Connection struct {
...
-	closeNetworkCalled int32
+	closeNetworkCalled atomic.Int32
 	// stoppedExchanges is atomically set when exchanges are stopped due to error.
-	stoppedExchanges uint32
+	stoppedExchanges atomic.Uint32
 }
 
 // nextConnID gives an ID for each connection for debugging purposes.
-var nextConnID uint32
+var nextConnID = &atomic.Uint32{}
```
prashantv: same as earlier, shouldn't need `&` here

peer_heap.go
```diff
@@ -44,7 +45,7 @@ func (ph *peerHeap) Less(i, j int) bool {
 	if ph.peerScores[i].score == ph.peerScores[j].score {
 		return ph.peerScores[i].order < ph.peerScores[j].order
```
prashantv: order is now an `atomic.Uint64` so it should fail to compile right?

also need to figure out what we want to do here -- this code path needs to be very fast since we have a lot of peers. there's no atomics here currently, I think that's because the caller of the peer heap locks and so it might be safe not to. Let's look into this a little more.

peer_strategies.go
```diff
@@ -20,25 +20,29 @@
...
+import (
+	"math"
+
+	"github.com/uber/tchannel-go/atomic"
+)
 
 // ScoreCalculator defines the interface to calculate the score.
 type ScoreCalculator interface {
-	GetScore(p *Peer) uint64
+	GetScore(p *Peer) *atomic.Uint64
```
prashantv: Why is the `ScoreCalculator` returning an atomic?

testutils/channel_opts.go
```diff
@@ -171,7 +172,11 @@ func DefaultOpts(opts *ChannelOpts) *ChannelOpts {
 
 // WrapLogger wraps the given logger with extra verification.
 func (v *LogVerification) WrapLogger(t testing.TB, l tchannel.Logger) tchannel.Logger {
+	counts := make([]*atomic.Uint32, len(v.Filters))
```
prashantv: shouldn't need pointers here. anything that's addressable can call pointer methods

peer_heap.go
akshayjshah: That's odd - you're right, I'd expect this to be prevent compilation. Weirdly, it doesn't: https://travis-ci.org/uber/tchannel-go/jobs/117770780.

I'll look into whether we need the atomics here.

examples/bench/client/client.go
```diff
@@ -42,7 +42,7 @@ var (
 	getToSetRatio = flag.Int("getToSetRatio", 1, "The number of Gets to do per Set call")
 
 	// counter tracks the total number of requests completed in the past second.
-	counter int64
+	counter = &atomic.Int64{}
```
prashantv: nit: don't need pointer here either

frame_pool_b_test.go
```diff
@@ -23,17 +23,17 @@ package tchannel_test
...
 	. "github.com/uber/tchannel-go"
+	"github.com/uber/tchannel-go/atomic"
 )
 
 func benchmarkUsing(b *testing.B, pool FramePool) {
 	const numGoroutines = 1000
 	const maxHoldFrames = 1000
 
-	var gotFrames uint64
+	gotFrames := &atomic.Uint64{}
```
prashantv: don't need pointer

testutils/channel.go
```diff
@@ -59,7 +60,7 @@ func NewServerChannel(opts *ChannelOpts) (*tchannel.Channel, error) {
 	return ch, nil
 }
 
-var totalClients uint32
+var totalClients = &atomic.Uint32{}
```
prashantv: don't need pointer

testutils/relay.go
```diff
@@ -37,7 +38,7 @@ type frameRelay struct {
 }
 
 func (r frameRelay) listen() (listenHostPort string, cancel func()) {
-	closed := uint32(0)
+	closed := &atomic.Uint32{}
```
prashantv: avoid pointer

prashantv:
lgtm once pointers that aren't needed are removed

testutils/channel.go
```diff
@@ -59,7 +60,7 @@ func NewServerChannel(opts *ChannelOpts) (*tchannel.Channel, error) {
 	return ch, nil
 }
 
-var totalClients uint32
+var totalClients = atomic.Uint32{}
```
prashantv: sorry for the nits, but we should just do
```
var totalClients atomic.Uint32
```
i want to avoid seeing `atomic.Uint32{}` (which is one of the main reasons to avoid the pointer syntax)

can you replace all instances of `atomic.Uint32{}` in the diff.

PR #272, opened 2016-03-22
Only add inbound connections that are active

prashantv:
We accept connectionStartClose in the outbound case to deal with
creating connections during shutdown to drain inbound calls. This
logic was copied from when inbound/outbound shared the same state.
Since they are different now, we do not need to accept inactive
connections on the inbound path.


akshayjshah:
lgtm

PR #275, opened 2016-03-28
Introduce testutils.TestServer 

akshayjshah:
Unify a variety of testing conveniences (`WithServer`, `WithVerifiedServer`,
and introspection) into a single `TestServer` type and a `WithTestServer`
context manager. Migrating to this combined struct will make it easier
to have each test case run with and without an interposed relay.

This commit introduces some temporary code duplication, which will be
resolved once all users of `WithVerifiedServer` are migrated.

### DIFF
--- a/testutils/test_server.go
+++ b/testutils/test_server.go
@@ -1,0 +1,234 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package testutils
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/uber/tchannel-go"
+	"github.com/uber/tchannel-go/atomic"
+	"github.com/uber/tchannel-go/testutils/goroutines"
+)
+
+// Has a previous test already leaked a goroutine?
+var _leakedGoroutine = atomic.NewInt32(0)
+
+// A TestServer encapsulates a TChannel server, a client factory, and functions
+// to ensure that we're not leaking resources.
+//
+// TODO: Include an optional relay.
+type TestServer struct {
+	testing.TB
+
+	// Don't embed the server, since we'll soon want to introduce a relay.
+	server     *tchannel.Channel
+	verifyOpts *goroutines.VerifyOpts
+}
+
+// NewTestServer constructs a TestServer.
+func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
+	return &TestServer{
+		TB:     t,
+		server: NewServer(t, opts),
+	}
+}
+
+// WithTestServer creates a new TestServer, runs the passed function, and then
+// verifies that no resources were leaked.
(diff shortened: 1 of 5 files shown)


testutils/test_server.go
```diff
@@ -0,0 +1,206 @@
...
+		TB:     t,
+		server: NewServer(t, opts),
+	}
+}
+
+// WithTestServer creates a new TestServer, runs the passed function, and then
+// verifies that no resources were leaked.
+//
+// TODO: run function twice; once with a relay, once without.
+func WithTestServer(t testing.TB, chanOpts *ChannelOpts, verifyOpts *goroutines.VerifyOpts, f func(*TestServer)) {
```
prashantv: Since the goroutine verification doesn't happen till after the test runs, we can avoid adding a new parameter by instead having functions on `TestServer` to set the goroutine verify opts.

testutils/test_server.go
```diff
@@ -0,0 +1,206 @@
...
+	// Channel is not closing, fail the test.
+	sinceStart := time.Since(started)
+	ts.Errorf("Channel did not close after %v, last state: %v", sinceStart, state)
+}
+
+func (ts *TestServer) verifyExchangesCleared() {
+	if ts.Failed() {
+		return
+	}
+	// Ensure that all the message exhanges are empty.
```
prashantv: typo in `exchanges`

testutils/test_server.go
```diff
@@ -0,0 +1,206 @@
...
+		time.Sleep(Timeout(sleepFor))
+	}
+
+	// Channel is not closing, fail the test.
+	sinceStart := time.Since(started)
+	ts.Errorf("Channel did not close after %v, last state: %v", sinceStart, state)
+}
+
+func (ts *TestServer) verifyExchangesCleared() {
+	if ts.Failed() {
```
prashantv: :+1:

testutils/test_server.go
```diff
@@ -0,0 +1,206 @@
...
+	serverState := ts.server.IntrospectState(&tchannel.IntrospectionOptions{
+		IncludeExchanges: true,
+	})
+	if exchangesLeft := describeLeakedExchanges(serverState); exchangesLeft != "" {
+		ts.Errorf("Found uncleared message exchanges on server:\n%v", exchangesLeft)
+	}
+}
+
+func (ts *TestServer) verifyNoGoroutinesLeaked(opts *goroutines.VerifyOpts) {
+	if ts.Failed() {
```
prashantv: I wonder if we should disable goroutine leak checking globally once a test fails, since every future test (even if they pass) will fail because of goroutine leaks?

testutils/test_server.go
akshayjshah: Good point, that'll definitely be nicer.

testutils/test_server.go
akshayjshah: Good point - I'll do that too.

testutils/test_server.go
```diff
@@ -0,0 +1,213 @@
...
+	serverState := ts.server.IntrospectState(&tchannel.IntrospectionOptions{
+		IncludeExchanges: true,
+	})
+	if exchangesLeft := describeLeakedExchanges(serverState); exchangesLeft != "" {
+		ts.Errorf("Found uncleared message exchanges on server:\n%v", exchangesLeft)
+	}
+}
+
+func (ts *TestServer) verifyNoGoroutinesLeaked() {
+	if ts.Failed() {
```
prashantv: might want to remove this, and cause it to fail once. otherwise, the fail counter will be 0, and it'll fail on the next passing test, which is kind of strange. other option is to make verify goroutines not care about previously failing, and do that logic in here. it can return a bool that is used to inc an atomic here instead, and you only check it here.

i actually like that more since the goroutines packages shouldn't care about skipping, it's the caller's responsibility

akshayjshah: Good point. I took a slightly different approach to fix the problem.

testutils/test_server.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (ts *TestServer) verifyNoGoroutinesLeaked() {
+	leaks, err := goroutines.IdentifyLeaks(ts.verifyOpts)
+	if err != nil {
+		ts.Error(err.Error())
+		return
+	}
+	// Count the leaked goroutines, even if this test has already failed.
+	if prevLeaks := _previouslyLeakedGoroutines.Add(uint32(len(leaks))); prevLeaks > 0 {
```
prashantv: `prevLeaks` will be the value after `len(leaks)` is added, so even on the first test that leaks, this will be > 0. i think you intended to check `prevLeaks - len(leaks) > 0` perhaps?

testutils/test_server.go
```diff
@@ -0,0 +1,228 @@
...
+		ts.Log("Skipping check for leaked goroutines, since we've already leaked at least %v.", prevLeaks)
```
prashantv: log says skipping check, but there's no return here. also message seems a little abrupt, "since we've already leaked at least 5". maybe "Skipping check for leaked goroutines due to %v previous leaks"?

testutils/test_server.go
prashantv: also previously lekaed goroutines will very likely show up in future tests, so I don't think this add approach will work. you probably want to compare `len(leaks) > prevLeaks`

akshayjshah: Instead of a doomed attempt to atomically count unique leaked goroutines, I'll flip a toggle on the first detected leak.

prashantv:
lgtm

akshayjshah:
Okay. I'm going to squash these commits, rebase, and then merge once tests are passing.

PR #276, opened 2016-03-29
Allow overriding thrift.Context creation

prashantv:
@yurishkuro: we can hook this up internally in xtchannel, what do you think?

r @akshayjshah


yurishkuro:
Yep, looks like it would work.

akshayjshah:
:ok_hand: lgtm

PR #277, opened 2016-03-29
Coveralls is flaky, don't fail the build if Coveralls fails

prashantv:
@akshayjshah


Qix-:
This has always been a problem.

If you'd like, you could even do

```shell
... || echo -e "\x1b[31mCoveralls failed\x1b[m"
```

to make it stand out a bit more.

prashantv:
Sure, colours are great. Currently everything is in the Makefile since that's where the output file is specified.

PR #278, opened 2016-03-29
Uncancelled contexts are leaking goroutines

prashantv:
Fix for contexts created that are not cancelled (from #268).

The leak happens because a goroutine is started to propagate the cancel.

cc @akshayjshah


akshayjshah:
:shipit:

PR #279, opened 2016-03-29
Avoid goroutine creation when creating contexts

prashantv:
Unwrap our custom `headerCtx` when it's used as a parent context, so the underlying context will be recognized by `context.Context` and it won't create a goroutine to track cancellation.

@akshayjshah

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -184,6 +184,22 @@ func TestContextBuilderParentContextReplaceHeaders(t *testing.T) {
 	goroutines.VerifyNoLeaks(t, nil)
 }
 
+func TestContextWrapWithHeaders(t *testing.T) {
+	headers1 := map[string]string{
+		"k1": "v1",
+	}
+	ctx, _ := NewContextBuilder(time.Second).
+		SetHeaders(headers1).
+		Build()
+	assert.Equal(t, headers1, ctx.Headers(), "Headers mismatch after Build")
+
+	headers2 := map[string]string{
+		"k1": "v1",
+	}
+	ctx2 := WrapWithHeaders(ctx, headers2)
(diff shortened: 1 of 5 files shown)


context_test.go
```diff
@@ -180,6 +180,24 @@ func TestContextBuilderParentContextReplaceHeaders(t *testing.T) {
...
+	ctx, _ := NewContextBuilder(time.Second).
+		SetHeaders(headers1).
+		Build()
+	assert.Equal(t, headers1, ctx.Headers(), "Headers mismatch after Build")
+
+	headers2 := map[string]string{
+		"k1": "v1",
+	}
+	ctx2 := WrapWithHeaders(ctx, headers2)
+	assert.Equal(t, headers2, ctx2.Headers(), "Headers mismatch after WrapWithHeaders")
```
akshayjshah: Should we also be checking for leaked goroutines here?

prashantv: This one doesn't cause `context.WithCancel` or `context.WithTimeout` to be called, so it can't leak a goroutine. The unwrapping in this method is just to avoid having a structure like:
```
context.Context -> headerCtx -> headerCtx -> headerCtx
```
Instead we want to flatten to
```
context.Context -> headerCtx
```

(`headerCtx` should always be the last context and we should only need one)

akshayjshah: Doesn't this test call `context.WithTimeout` in `Build()`, just like the other tests? We happen to know that, because of some implementation details, this *shouldn't* orphan a goroutine because (a) we're not setting a parent context and (b) we're now unwrapping the parent context.

Since we've introduced this ContextBuilder abstraction, I guess I'd expect the tests to assert that it's not leaking goroutines instead of assuming that some methods are implemented safely.

PR #282, opened 2016-03-29
Flaky test debugging

prashantv:
This is not intended to be submitted.


prashantv:
We can't enable logs for all tests since it hits the Travis 4MB limit. Instead #291 will only print the debug logs if a test fails.

PR #284, opened 2016-03-29
Make leaked goroutine detector simpler

akshayjshah:
Now that our tests don't leak tons of goroutines, we don't need to be so picky about which blocked goroutines are problematic - just fail the tests if any goroutines aren't making progress. This makes the retry logic much simpler, and should resolve some test flakiness.

Addresses some of #207.

### DIFF
--- a/testutils/goroutines/verify.go
+++ b/testutils/goroutines/verify.go
@@ -21,50 +21,15 @@
 package goroutines
 
 import (
-	"bufio"
-	"bytes"
 	"fmt"
-	"io"
 	"runtime"
-	"strings"
 	"testing"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


testutils/goroutines/verify.go
```diff
@@ -78,92 +43,37 @@ func filterStacks(stacks []Stack, opts *VerifyOpts) []Stack {
...
-					time.Sleep(time.Millisecond)
-				}
-				continue retry
-			}
-		}
-
-		// There are no running/runnable goroutines, so check for bad leaks.
-		leakStacks := getLeakStacks(stacks)
 
 		// If there are leaks found, retry 3 times since the goroutine's state
```
prashantv: stale comment

prashantv:
lgtm

PR #285, opened 2016-03-29
Use go test -i in Makefile

akshayjshah:
Should make tests slightly faster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,10 @@ endif
 install_ci: get_thrift install
 	go get -u github.com/mattn/goveralls
 
+install_test:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,6 +65,10 @@ endif
 install_ci: get_thrift install
 	go get -u github.com/mattn/goveralls
 
+install_test:
+	go test -i $(addprefix github.com/uber/tchannel-go/,$(NO_TESTUTILS_PKGS)) $(TEST_ARG)
```
prashantv: nit: let's use a single `go test -i` command, put all the packages in one line

prashantv:
lgtm

PR #288, opened 2016-03-29
Make TestNoLeakedState run faster, and use CloseAndVerify

prashantv:
Attempt to make the test less flaky (#207)

@akshayjshah


prashantv:
@akshayjshah suggested making all tests verify that the runtime state has no leaks rather than having a special test for it. I think that's a better approach and gives us better verification for all our tests.

PR #274, opened 2016-03-22
Add test to trigger data races during test and fix races

prashantv:
There's a few data races in the connection shutdown path. It's possible for these races to cause a panic, but they would only affect servers that are closing anyway.

Fixes #262, #273

PR #289, opened 2016-03-29
Fix flaky TestCloseSendError test

akshayjshah:
WIP! Temporarily amending Makefile to make sure we can reliably reproduce this test failure on Travis.

Calls to the echo procedure are timing out on Travis; there's not much we can do about it except to increase the timeout.

Partially addresses #207.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ godep:
 test_ci: test
 ifdef SHOULD_LINT
 	@echo Stress testing TestCloseSendError
-	go test $(TEST_ARG) -run "TestCloseSendError" -timeoutMultiplier=10 -parallel=4 -count 500 .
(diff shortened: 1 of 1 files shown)


akshayjshah:
@prashantv This test isn't triggering the failure we're looking for (from https://travis-ci.org/uber/tchannel-go/jobs/119323282):

```
--- FAIL: TestCloseSendError (0.31s)
    close_test.go:446: Call failed: tchannel error ErrCodeTimeout: timeout
    close_test.go:446: Call failed: tchannel error ErrCodeTimeout: timeout
    close_test.go:446: Call failed: tchannel error ErrCodeTimeout: timeout
    close_test.go:446: Call failed: tchannel error ErrCodeTimeout: timeout
```

I'm running it continuously for minutes (until Travis times out the build step), but all I'm surfacing is the data race you're already fixing. I'd rather not mess with the timeouts for no reason...

akshayjshah:
Closing in favor of one of #274's commits.

PR #290, opened 2016-03-30
Verify no state leaks (e.g. connections) in TestServer

prashantv:
@akshayjshah

### DIFF
--- a/testutils/test_server.go
+++ b/testutils/test_server.go
@@ -30,6 +30,8 @@ import (
 	"github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/atomic"
 	"github.com/uber/tchannel-go/testutils/goroutines"
+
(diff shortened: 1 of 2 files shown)


akshayjshah:
lgtm. Can we also delete the `ServerState` method, since I don't think it has any callers now?

testutils/test_server.go
```diff
@@ -27,6 +27,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
akshayjshah: should be in a separate block.

prashantv:
`ServerState` has already been removed.

testutils/test_server.go
prashantv: :+1:

akshayjshah:
lgtm.

PR #291, opened 2016-03-30
Print debug logs on test failures

prashantv:
Write all logs to testing.T's logger, which only prints the output if
the test fails.

If --connectionLog is specified, logs will be streamed as usual.
Otherwise, output will be printed only if the test fails.

@akshayjshah


prashantv:
This prints logs even if `-v` is specified unfortunately.. and `-v` is useful to make sure we're making progress..

prashantv:
Updated to store the buffer and report the failure from `testutils.TestServer`. Should be ready to review.

akshayjshah:
lgtm...logging is a global PITA.

PR #294, opened 2016-03-31
Ignore subchannels and peer during state leak checking

prashantv:
A service may have new subchannels and peers since they get added/modified during tests.

We only care about connections which is in root peers, so it's safe to ignore the Peers.

@akshayjshah


akshayjshah:
:shipit:

PR #296, opened 2016-03-31
Test timeouts should not require goroutine + increase timeouts

prashantv:
@akshayjshah

### DIFF
--- a/testutils/timeout.go
+++ b/testutils/timeout.go
@@ -57,20 +57,14 @@ func SetTimeout(t *testing.T, timeout time.Duration) func() {
 	timeout = Timeout(timeout)
 
 	caller := getCallerName()
-	c := make(chan struct{})
 
-	go func() {
-		select {
-		case <-c:
-			// Test is complete, don't need to do anything.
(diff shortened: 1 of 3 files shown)


testutils/timeout.go
```diff
@@ -57,20 +57,14 @@ func SetTimeout(t *testing.T, timeout time.Duration) func() {
...
-	}()
+	timer := time.AfterFunc(timeout, func() {
+		t.Logf("Test %s timed out after %v", caller, timeout)
+		// Unfortunately, tests cannot be failed from new goroutines, so use a panic.
+		panic(fmt.Errorf("Test %s timed out after %v", caller, timeout))
+	})
 
 	return func() {
-		c <- struct{}{}
+		timer.Stop()
```
akshayjshah: Can we just `return timer.Stop`?

prashantv: `timer.Stop` returns a bool so we'd need to change the signature, and I think it's cleaner for this function to just return `func()`

akshayjshah:
:shipit:

PR #297, opened 2016-03-31
Don't look for leaks if leak detector has previously gone off

prashantv:
If the leak detector has previously failed, it will most likely identify the same leaks
again. Running it again will spend a second per test just waiting
to see if the leak goes away (which it probably won't).

@akshayjshah


akshayjshah:
:shipit:

PR #299, opened 2016-03-31
Add logs for connection timeouts and non-retriable failures

prashantv:
@akshayjshah

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -205,6 +205,7 @@ func (ch *Channel) newOutboundConnection(timeout time.Duration, hostPort string,
 	conn, err := net.DialTimeout("tcp", hostPort, timeout)
 	if err != nil {
 		if ne, ok := err.(net.Error); ok && ne.Timeout() {
+			ch.log.WithFields(LogField{"hostPort", hostPort}, LogField{"timeout", timeout}).Infof("Outbound net.Dial timed out")
 			err = ErrTimeout
 		}
 		return nil, err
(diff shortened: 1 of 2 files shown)


akshayjshah:
lgtm. Does it make sense to drop the level to `Debug`? Production logging via Logrus/Kafka REST/etc. is relatively slow and alloc-heavy for an expected error that we're supposed to be handling gracefully.

prashantv:
Yeah I wasn't sure about the level, the way we've configured TChannel in xtchannel is that only TChannel Warn+ logs go to the application's logger, so these logs won't show up in application logs.

The difference between Info and Debug is that I want to enable Info logs during development, where I want to expose things that shouldn't really happen during development (like connections timing out), while Debug logs are things that are only useful when you're debugging TChannel itself. What do you think?

akshayjshah:
Sure, that works for me.

On Thu, Mar 31, 2016 at 10:56 AM Prashant Varanasi <notifications@github.com>
wrote:

> Yeah I wasn't sure about the level, the way we've configured TChannel in
> xtchannel is that only TChannel Warn+ logs go to the application's logger,
> so these logs won't show up in application logs.
>
> The difference between Info and Debug is that I want to enable Info logs
> during development, where I want to expose things that shouldn't really
> happen during development (like connections timing out), while Debug logs
> are things that are only useful when you're debugging TChannel itself. What
> do you think?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/uber/tchannel-go/pull/299#issuecomment-204052444>
>
-- 
Akshay Shah
RPC
shah@uber.com
(818) 312-4647

PR #305, opened 2016-04-02
TestServer should allow creating validated clients/servers

prashantv:
TestServer.New{Client,Server} will return a channel that is validated. Channels are closed and validated in most recently created order.

This is required to avoid doing goroutine verification for the inner channels. If the inner channel does goroutine verification, the outer server's accept loop will cause the verification to fail. We skip goroutine verification till the parent TestServer is closed.

@akshayjshah


prashantv:
@akshayjshah merging to unblock more changes + 1.0.5 release, but would like you to review this change when you're back.

PR #312, opened 2016-04-04
Improve debugging log for debugging unexpected connection change

prashantv:


akshayjshah:
lgtm

prashantv:
No longer need this change, since I've found the cause and opened a separate PR: #313

PR #316, opened 2016-04-06
Use UnmarshalText for FailStrategy

prashantv:
This makes it compatible with JSON and YAML (and a bunch of other encoding formats).


akshayjshah:
lgtm

PR #317, opened 2016-04-15
Use a separate map for inbound/outbound relay items

prashantv:
Currently, we use a single map that remaps from ID -> some connection, but IDs are not global across a connection. E.g. if we have `s1` and `s2`, `s1` may make a callReq with ID 1 to `s2`, and `s2` may also start a callReq with ID 1 to `s1`.

Change the relays to use separate maps for the outbounds (callReq starting on this connection) vs inbounds (callReq started on some other connection being forwarded to this connection).

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -2,6 +2,7 @@ package tchannel
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


relay.go
```diff
@@ -50,37 +69,55 @@ func (r *Relayer) Hosts() RelayHosts {
...
+	{
+		// NOTE, this block is safety checking. We should not enable this in
+		// production builds.
+
+		// If we receive a response frame, we expect to find that ID in our outbound.
+		// If we receive a request frame, we expect to find that ID in our inbound.
+		items := r.outbound
+		if fType == requestFrame {
+			items = r.inbound
+		}
```
akshayjshah: This block reappears a few times. Should we pull it out into a separate method - pass in a frameType, get a map?

relay.go
```diff
@@ -50,37 +69,55 @@ func (r *Relayer) Hosts() RelayHosts {
...
-	_, ok := r.items[f.Header.ID]
-	r.RUnlock()
-	if !ok {
-		r.logger.WithFields(
-			LogField{"ID", f.Header.ID},
-		).Warn("Received a frame without a RelayItem.")
+// Receive receives frames intended for this connection.
+func (r *Relayer) Receive(f *Frame, fType frameType) {
+	{
+		// NOTE, this block is safety checking. We should not enable this in
```
akshayjshah: Should this be a `TODO` so we can find it with the others?

akshayjshah:
nits and data races aside, lgtm

relay.go
prashantv: Done

relay.go
prashantv: Yeah thought about this a few times. I didn't want a generic method since it actually depends on whether you're the receiver or it originated on your connection. Put it into a method, but made it specific `receiverItems`

PR #318, opened 2016-04-15
Add periodic health checks to TChannel

prashantv:
If a TChannel connection stalls (e.g., the remote side has disappeared,
and traffic starts black-holing), TChannel waits for TCP to detect this
issue, but this is often not configurable, and it can be a long period
of time (we've observed many minutes in production).
    
In many P2P situations, it's desirable to detect these failures as soon
as possible. Support active health checking using TChannel pings on a
periodic basis, with configurable timeouts, number of failures etc.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -136,6 +136,10 @@ type ConnectionOptions struct {
 
 	// ToS class name marked on outbound packets.
 	TosPriority tos.ToS
+
+	// HealthChecks configures active connection health checking for this channel.
+	// By default, health checks are not enabled.
+	HealthChecks HealthCheckOptions
 }
 
 // connectionEvents are the events that can be triggered by a connection.
@@ -186,6 +190,11 @@ type Connection struct {
 	// remotePeerAddress is used as a cache for remote peer address parsed into individual
 	// components that can be used to set peer tags on OpenTracing Span.
 	remotePeerAddress peerAddressComponents
+
+	// healthCheckCtx/Quit are used to stop health checks.
+	healthCheckCtx     context.Context
+	healthCheckQuit    context.CancelFunc
+	healthCheckHistory *healthHistory
 }
 
 type peerAddressComponents struct {
@@ -237,6 +246,7 @@ func (co ConnectionOptions) withDefaults() ConnectionOptions {
 	if co.SendBufferSize <= 0 {
 		co.SendBufferSize = defaultConnectionBufferSize
 	}
+	co.HealthChecks = co.HealthChecks.withDefaults()
 	return co
 }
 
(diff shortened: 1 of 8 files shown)


connection.go
```diff
@@ -891,11 +901,62 @@ func (c *Connection) closeNetwork() {
...
+		c.log.Debug("Health checks disabled by negative healthCheckPeriod")
+		return
+	}
+
+	if interval == 0 {
+		interval = time.Second
+	}
+	if timeout == 0 {
+		timeout = time.Second
+	}
```
akshayjshah: Do we want to validate/enforce that timeout <= duration?

Motiejus:
The "health" branch of tchannel-go is good for ringpop: if destination blackholes, we see that within 1 second, without filling any outgoing buffers, and ringpop has a better chance to detect the transport error.

I am all-in for this feature.

prashantv:
Is there any concerns about having constant pings on any TCP connection? tchannel-go does not currently close working TCP connections, so over time, if ringpop ends up connecting to every other host, each of those connections will have constant health checks.

The main concern I have about this change is the default health timeout period, since some services may have much slower responses. We could even ship this change but have a pretty high value (like 10 seconds or something) -- although I imagine ringpop would want a much lower value. Since it doesn't create the channel, the service owner would need to set a lower value, which seems like extra configuration that we could avoid by having a lower default.

connection.go
```diff
@@ -891,11 +901,62 @@ func (c *Connection) closeNetwork() {
...
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		if err := c.ping(ctx); err != nil {
+			// If the health check failed because the connection is closed then
+			// we don't need to do any extra logging or close the connection.
+			if err != ErrInvalidConnectionState {
+				c.connectionError("healthCheck", fmt.Errorf("healthCheck failed: %v", err))
+			}
+			return
+		}
+		cancel()
```
yurishkuro: is ok that `cancel()` is not called when this returns in L948?

connection.go
```diff
@@ -891,11 +901,62 @@ func (c *Connection) closeNetwork() {
...
+	}
+}
+
+func (c *Connection) stophealthCheck() {
```
yurishkuro: upper-case H in health

yurishkuro:
@prashantv is there an eta for this change?

prashantv:
I'll probably not get to this in the next week, there's a bunch of relay work that's higher priority right now.

madhuravi:
@prashantv Are there any plans to get this in? Looks like it's been over a year since it was originally created. This would be useful for cadence.

prashantv:
I've updated the diff with full test coverage. Health checking is opt-in, and by default no active health checks are performed.

The caller can choose the interval between checks, what the timeouts are for each ping message, and how many consecutive failures should cause a connection to be closed.

health.go
```diff
@@ -0,0 +1,169 @@
...
+	consecutiveFailures := 0
+	for {
+		select {
+		case <-ticker.C:
+		case <-c.healthCheckCtx.Done():
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(c.healthCheckCtx, opts.Timeout)
+		defer cancel()
```
akshayjshah: Do we want to defer this until the function returns, or should we make sure to execute it at the end of this loop iteration?

akshayjshah:
I am *very* excited about this.

health.go
prashantv: Good catch! Could have been a huge memory leak, changed ti to `cancel` immediately after the `ping` method, since we don't use the `ctx` for anything else.

PR #320, opened 2016-04-19
Package names should always be lowercase

prashantv:
Fix bug where thrift-gen did not lowercase the package name
for imported packages.

@abhinav 
cc: @akshayjshah


abhinav:
👍

PR #321, opened 2016-04-20
Fix ErrInvalidConnectionState not being retried

prashantv:
It's possible that ErrInvalidConnectionState is retried in a race
between selecting a connection, and the connection being closed.

This should be treated as a network error which can be retried.

Caught this while debugging why TestRequestStateRetry was flaky, this
fixes the flakiness.

To test the flakiness, I ran this test 1000 times till it failed,
https://travis-ci.org/uber/tchannel-go/jobs/124585965

After the fix, it passes even after 1000 iterations,
https://travis-ci.org/uber/tchannel-go/jobs/124586838

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -33,7 +33,9 @@ import (
 
 var (
 	// ErrInvalidConnectionState indicates that the connection is not in a valid state.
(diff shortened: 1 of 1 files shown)


blampe:
🚢

PR #323, opened 2016-04-21
Relay should handle error frames, and finish relays after errors

prashantv:


connection.go
```diff
@@ -742,7 +742,7 @@ func (c *Connection) readFrames(_ uint32) {
 func (c *Connection) handleFrameRelay(frame *Frame) bool {
 	// TODO: handle error frames.
```
akshayjshah: Remove this todo?

akshayjshah:
nit aside, lgtm.

connection.go
prashantv: 👍

PR #324, opened 2016-04-21
Create new lazy message types that wrap a *Frame

prashantv:
The wrapped classes are the same size as the pointer, and allow us to
add messageType-specific methods on top of a frame rather than adding
them to the frame directly.

In future, we may want to cache items in the wrapped class.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -38,10 +38,6 @@ const (
 
 	// MaxFramePayloadSize is the maximum size of the payload for a single frame
 	MaxFramePayloadSize = MaxFrameSize - FrameHeaderSize
-
-	_flagsIndex       = 0
-	_serviceLenIndex  = 1 /* flags */ + 4 /* ttl */ + 25 /* tracing */
-	_serviceNameIndex = _serviceLenIndex + 1
 )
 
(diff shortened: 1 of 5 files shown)


akshayjshah:
Go 1.4 seems to think that we're not cleaning up relay items:

```
=== RUN TestRelayConnectionCloseDrainsRelayItems-2
--- FAIL: TestRelayConnectionCloseDrainsRelayItems-2 (0.03s)
	Error Trace:	relay_test.go:33
			relay_test.go:95
			relay_test.go:165
	Error:		Not equal: 0 (expected)
			        != 1 (actual)
	Messages:	Found 1 left-over items in relayer for 127.0.0.1:43465.
		
	Error Trace:	relay_test.go:33
			relay_test.go:95
			relay_test.go:165
	Error:		Not equal: 0 (expected)
			        != 1 (actual)
	Messages:	Found 1 left-over items in relayer for 127.0.0.1:37902.
```

prashantv:
I don't think this change is responsible for that error, seems like we have a flaky test.

akshayjshah:
Agreed. Added it to the flaky tests issue. o/w lgtm.

PR #325, opened 2016-04-21
Add an option to relay TestServer

akshayjshah:
1. Move the SimpleRelayHosts type and constructor to the testutils package, where it belongs. Previously, it lived in a _test.go file in tchannel.
2. Add an option to use a relay in TestServer. This fails virtually all tests.


testutils/test_server.go
```diff
@@ -67,12 +71,17 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
...
 
-	return &TestServer{
+	ts := &TestServer{
 		TB:            t,
 		channels:      []*tchannel.Channel{ch},
 		channelStates: states,
 		postFns:       opts.postFns,
 	}
+
+	if opts.IncludeRelay {
```
prashantv: For tests that can be run with a relay, I was thinking we should run them twice. Once without, and once with the relay.

This would stay the same, but if someone calls `WithTestServer` with `IncludeRelay`, we should need to copy the options and run with that set to false once.

What do you think?

prashantv:
Can we enable the relay in a couple of our simple connection tests (like just testing a successful request or a large request)?

testutils/test_server.go
akshayjshah: Yep, that's what I had in mind. Once relaying is working well, I'd like to make that behavior opt-out instead of opt-in.

prashantv:
lgtm

PR #326, opened 2016-04-21
Improve introspection for exchanges/relay items

prashantv:
Relay items introspection is similar to message exchanges -- we should add as much information as possible when `IncludeExchanges` is enabled.

Use a`map[string]T` for the exchanges/items map to match the internal state.


prashantv:
Need to open this against the muttley branch.

PR #319, opened 2016-04-19
Support a separate connection timeout from the context timeout

prashantv:
Fixes #300 

@akshayjshah @nomis52

PR #328, opened 2016-04-21
Run some tests with & without a relay

akshayjshah:
Opt a few tests into running twice, both with and without a relay.

### DIFF
--- a/testutils/test_server.go
+++ b/testutils/test_server.go
@@ -89,14 +89,12 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
 //
 // TODO: run function twice; once with a relay, once without.
 func WithTestServer(t testing.TB, chanOpts *ChannelOpts, f func(*TestServer)) {
-	ts := NewTestServer(t, chanOpts)
-	// Note: We use defer, as we want the postFns to run even if the test
-	// goroutine exits (e.g. user calls t.Fatalf).
-	defer ts.post()
-
-	f(ts)
-	ts.Server().Logger().Debugf("TEST: Test function complete")
-	ts.CloseAndVerify()
+	withServer(t, chanOpts, f)
+	if chanOpts != nil && chanOpts.IncludeRelay {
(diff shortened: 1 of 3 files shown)


testutils/test_server.go
```diff
@@ -89,14 +89,12 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
...
-	// Note: We use defer, as we want the postFns to run even if the test
-	// goroutine exits (e.g. user calls t.Fatalf).
-	defer ts.post()
-
-	f(ts)
-	ts.Server().Logger().Debugf("TEST: Test function complete")
-	ts.CloseAndVerify()
+	withServer(t, chanOpts, f)
+	if chanOpts != nil && chanOpts.IncludeRelay {
+		repeatOpts := chanOpts.Copy()
```
prashantv: Few things:
 - we're copying it too late, we should be copying before running it at all.
 - this is going to clash with a change I have on dev: https://github.com/uber/tchannel-go/pull/319/commits/6d1b837b43cb163db0a5be69fedc49e8958ac5cf (part of #319).
 - we should run without the relay first (in case there's test bugs when developing, I'd rather see the direct connection errors first).

We should:
 - make a separate PR to add the `Copy` method and use a copy when running WithServer at all (on to `dev`)
 - then rebase muttley off dev, and use the same thing here, the logic should really be:

```go
func WithTestServer(t testing.TB, origOpts *ChannelOpts, f func(*TestServer)) {
  opts :=origOpts.Copy() // this should return empty opts if it's nil automatically

  // Run the test without a relay initially
  opts.IncludeRelay = false
  withServer(t, opts, f)

  if origOpts != nil && origOpts.IncludeRelay {
    repeatOpts := origOpts.Copy()
    withServer(t, repeatOpts, f)
  }
```

akshayjshah: Argh, didn't realize that we were mutating the options.

prashantv:
lgtm

PR #329, opened 2016-04-21
Ensure that all created channels are closed

prashantv:
I'm moving towards outputting the introspected state on test failures (e.g. #326), which includes `OtherChannels`. To keep the errors readable and of a reasonable length, we should make sure we always Close channels in tests.

We verify that all channels are closed once all the tests are run successfully in `TestMain`.


akshayjshah:
:+1: sounds good.

PR #334, opened 2016-04-22
Remove withRelayTest and use TestServer instead

prashantv:
Add a new option to `TestServer` to only run with a relay, which is used by the relay tests.


akshayjshah:
lgtm

PR #335, opened 2016-04-22
Expose RunCount to run a test multiple times.

prashantv:
This will make debugging flaky tests much easier, and can be useful
for tests that use random and may need to run a few times

### DIFF
--- a/testutils/channel_opts.go
+++ b/testutils/channel_opts.go
@@ -54,6 +54,10 @@ type ChannelOpts struct {
 	// OnlyRelay instructs TestServer the test must only be run with a relay.
 	OnlyRelay bool
 
+	// RunCount is the number of times the test should be run. Zero or
+	// negative values are treated as a single run.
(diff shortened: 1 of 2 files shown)


akshayjshah:
:shipit: lgtm

relay_test.go
```diff
@@ -71,7 +71,6 @@ func DisabledTestRelayHandlesCrashedPeers(t *testing.T) {
 }
 
 func TestRelayConnectionCloseDrainsRelayItems(t *testing.T) {
-	opts := serviceNameOpts("s1").SetRelayOnly()
```
akshayjshah: Pretty sure the compiler needs this ;)

prashantv: whoops, was doing testing with runcount set to 1000, and "reverted" by just deleting the whole line. travis is unhappy.

akshayjshah:
looks good!

PR #338, opened 2016-04-23
TestServer fixes and clean up

prashantv:
(This PR backports some ocde from the muttley branch, and specifically, has extra functions like `withTestServer` to make it easy to merge with the `muttley` branch).

There's been a bunch of testutils.TestServer issues:
 - Logs not always being verified
 - Debug logs not always printed for all channels
 - Unclear mutations of options

This change cleans up:
 - Mutation of options. Now we only do it in `update*` methods, and exported functions never modify the passed in `opts` (they always make a copy). I've tried to reduce the amount of options mutations as well (e.g. `GetTChannelOptions`).
 - Internal functions that require mutation (e.g. for the post-verification steps), will use unexported methods
 - TestServer no longer has a special first channel, but adds the initial channel using the same `NewTestServer` path that other channels go through
 - State verification is now a post-step since we need to wait for runnable goroutines to end to avoid flaky tests

Some new failures started after the fixes as some tests weren't seeing their log filter options correctly.

### DIFF
--- a/close_test.go
+++ b/close_test.go
@@ -21,7 +21,6 @@
 package tchannel_test
 
 import (
-	"log"
 	"math/rand"
 	"sync"
 	"testing"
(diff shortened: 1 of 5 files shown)


akshayjshah:
lgtm - this is much nicer.

PR #339, opened 2016-04-25
Simplify testutils again

prashantv:
Remove testutils.ChannelOpts.optFn completely. All modifications
to opts only happen inside of updateOpts*.

Previously, optFn was updating the options, and was run inside of
GetTChannelOptions which was working on a copy. This meant that any
changes optFn made were lost. Simplifying to reduce confusion.

@akshayjshah


akshayjshah:
:shipit:

PR #340, opened 2016-04-25
Migrate the remaining connection tests to testutils.TestServer

akshayjshah:
Migrate the rest of the tests. One of the tests (`TestRemotePeer`) can work with a relay, but the remainder won't work with relays until we handle timeouts properly.

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -369,7 +369,7 @@ func (h onErrorTestHandler) OnError(ctx context.Context, err error) {
 }
 
 func TestTimeout(t *testing.T) {
-	// TODO: Add relaying.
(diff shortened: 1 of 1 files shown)


connection_test.go
```diff
@@ -351,19 +369,21 @@ func (h onErrorTestHandler) OnError(ctx context.Context, err error) {
 }
 
 func TestTimeout(t *testing.T) {
-	WithVerifiedServer(t, nil, func(ch *Channel, hostPort string) {
+	// TODO: Add relaying.
```
prashantv: change todo to mention that we need to implement timeouts in the relay before we can enable this test

prashantv:
lgtm

PR #341, opened 2016-04-26
Fix unexpected goroutine for long runnign tests

prashantv:
The background "RunTests" goroutine changes state from:
"chan wait" to "chan wait, waiting for N minutes".

We should ignore suffixes for "chan wait" (similar to the syscall) to
avoid failures on long running tests.

While tests shouldn't run for > a minute, it can happen when debugging
and running the test in a loop.


akshayjshah:
lgtm

PR #333, opened 2016-04-22
Convert more connection tests to check relaying behavior

akshayjshah:
These tests work without problems - they just needed to be converted to use `WithTestServer` instead of `WithVerifiedServer`.

I'm still working on a handful of other tests that currently don't work.

PR #342, opened 2016-04-26
Fix TestConnectTimeout flakes

prashantv:
Couple of bugs caused flakiness:
 - the testutils.Relay had a data race
 - The test was using too short of a connect timeout and was ending before the TCP connections had been made

### DIFF
--- a/testutils/relay.go
+++ b/testutils/relay.go
@@ -61,6 +61,7 @@ func (r *frameRelay) listen() (listenHostPort string, cancel func()) {
 			r.Lock()
 			r.conns = append(r.conns, c)
 			r.Unlock()
+
 			r.relayConn(c)
 		}
 	}()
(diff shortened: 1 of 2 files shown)


akshayjshah:
lgtm

testutils/relay.go
```diff
@@ -84,8 +85,14 @@ func (r *frameRelay) relayConn(c net.Conn) {
 		return
 	}
 	r.Lock()
+	defer r.Unlock()
+
+	if r.closed.Load() > 0 {
+		outC.Close()
+		return
+	}
+
```
akshayjshah: We're losing track of the rebases here...this diff doesn't include the changes from https://github.com/uber/tchannel-go/pull/333/commits/6017cf77af9468f6d199ae173ca5130cd1907c8c.

PR #344, opened 2016-04-28
Set test timeout scale factor using environment variable instead of flag

prashantv:
Using an environment variable means it can be set in the .travis.yml file and the Makefile does not need to distinguish between packages that use testutils and packages that don't.

This should also fix some "timeout" flakes we see, it looks like we were running tests for coverage without the timeout multiplier, and so tests were being run with a lower timeout than expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ go:
   - 1.6
   - tip
 install: make install_ci
+env:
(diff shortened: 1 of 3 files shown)


akshayjshah:
mmhmmm, yup - much better. :+1:

PR #346, opened 2016-04-29
Create new benchmark package with benchmark server/client

prashantv:
Make it easy to run these benchmarks as a separate process by passing an
option. Use new benchmark package for Thrift benchmarks.

Goal is to also add benchmarks to `_test.go` in the same folder, with the same code being used for raw and Thrift. That will come as a separate change.


akshayjshah:
lgtm

akshayjshah:
This is pretty awesome!

PR #347, opened 2016-04-29
Handle timeouts (with tombstones) in relay

akshayjshah:
Add timeout handling and tombstones to the relay, and enabled relaying in the remaining connection tests.

### DIFF
--- a/close_test.go
+++ b/close_test.go
@@ -27,7 +27,6 @@ import (
 	"time"
 
 	. "github.com/uber/tchannel-go"
-
 	"github.com/uber/tchannel-go/atomic"
 	"github.com/uber/tchannel-go/raw"
 	"github.com/uber/tchannel-go/testutils"
@@ -79,7 +78,8 @@ func TestCloseNewClient(t *testing.T) {
 }
 
 func TestCloseAfterTimeout(t *testing.T) {
-	testutils.WithTestServer(t, nil, func(ts *testutils.TestServer) {
+	opts := testutils.NewOpts().SetRelay()
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
 		testHandler := onErrorTestHandler{newTestHandler(t), func(_ context.Context, err error) {}}
 		ts.Register(raw.Wrap(testHandler), "block")
 
--- a/introspection.go
+++ b/introspection.go
@@ -36,6 +36,9 @@ type IntrospectionOptions struct {
 
 	// IncludeEmptyPeers will include peers, even if they have no connections.
 	IncludeEmptyPeers bool `json:"includeEmptyPeers"`
+
+	// IncludeTombstones will include tombstones when introspecting relays.
+	IncludeTombstones bool `json:"includeTombstones"`
 }
 
 // RuntimeState is a snapshot of the runtime state for a channel.
@@ -157,6 +160,7 @@ type RelayItemState struct {
 	ID                      uint32 `json:"id"`
 	RemapID                 uint32 `json:"remapID"`
 	DestinationConnectionID uint32 `json:"destinationConnectionID"`
+	Tomb                    bool   `json:"tomb"`
 }
 
 // PeerRuntimeState is the runtime state for a single peer.
@@ -318,28 +322,34 @@ func (c *Connection) IntrospectState(opts *IntrospectionOptions) ConnectionRunti
 
 // IntrospectState returns the runtime state for this relayer.
 func (r *Relayer) IntrospectState(opts *IntrospectionOptions) RelayerRuntimeState {
-	r.RLock()
-	defer r.RUnlock()
+	count := r.inbound.Count() + r.outbound.Count()
 	return RelayerRuntimeState{
-		Count:         len(r.inbound) + len(r.outbound),
-		InboundItems:  r.IntrospectItems(opts, "inbound", r.inbound),
-		OutboundItems: r.IntrospectItems(opts, "outbound", r.outbound),
+		Count:         count,
+		InboundItems:  r.inbound.IntrospectState(opts, "inbound"),
+		OutboundItems: r.outbound.IntrospectState(opts, "outbound"),
 	}
 }
 
-// IntrospectItems returns the runtime state for the given relay items.
-func (r *Relayer) IntrospectItems(opts *IntrospectionOptions, name string, items map[uint32]relayItem) RelayItemSetState {
+// IntrospectState returns the runtime state for this relayItems.
+func (ri *relayItems) IntrospectState(opts *IntrospectionOptions, name string) RelayItemSetState {
+	ri.RLock()
+	defer ri.RUnlock()
+
 	setState := RelayItemSetState{
 		Name:  name,
-		Count: len(items),
+		Count: ri.Count(),
 	}
 	if opts.IncludeExchanges {
-		setState.Items = make(map[string]RelayItemState, len(items))
-		for k, v := range items {
+		setState.Items = make(map[string]RelayItemState, len(ri.items))
+		for k, v := range ri.items {
+			if !opts.IncludeTombstones && v.tomb {
+				continue
+			}
 			state := RelayItemState{
 				ID:                      k,
 				RemapID:                 v.remapID,
 				DestinationConnectionID: v.destination.conn.connID,
+				Tomb: v.tomb,
 			}
 			setState.Items[strconv.Itoa(int(k))] = state
 		}
--- a/relay.go
+++ b/relay.go
@@ -8,6 +8,13 @@ import (
 	"time"
 )
 
+// _maxRelayTombs is the maximum number of tombs we'll accumulate in a single
+// relayItems.
+const _maxRelayTombs = 1e4
+
+// _relayTombTTL is the length of time we'll keep a tomb before GC'ing it.
+const _relayTombTTL = time.Second
+
 // RelayHosts allows external wrappers to inject peer selection logic for
 // relaying.
 type RelayHosts interface {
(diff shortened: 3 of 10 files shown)


testutils/logger.go
```diff
@@ -177,12 +177,13 @@ func (l errorLogger) checkFilters(msg string) bool {
...
 	return uint(matchCount) <= l.v.Filters[match].Count
 }
+
 func (l errorLogger) checkErr(prefix, msg string) {
 	if l.checkFilters(msg) {
 		return
 	}
 
-	l.t.Errorf("%v: %s %v", prefix, msg, l.Logger.Fields())
+	l.t.Errorf("Unexpected failure during test: %v: %s %v", prefix, msg, l.Logger.Fields())
```
prashantv: nit: since this prefixes every log, I'd prefer something concise like,
```go
l.t.Errorf("Unexpected log: %v: %s %v", prefix, msg, l.Logger.Fields())
```

relay.go
```diff
@@ -17,8 +24,85 @@ type RelayHosts interface {
...
+// Count returns the number of items in the relay.
+func (r *relayItems) Count() int {
+	r.RLock()
+	n := len(r.items)
+	r.RUnlock()
+	return n
+}
+
+// Get checks for a relay item by ID. It returns the item, if it's still active,
+// whether the item is a tombstone,
```
prashantv: comment seems incorrect (and truncated) -- `Get` just returns whether something was found.

relay.go
```diff
@@ -17,8 +24,85 @@ type RelayHosts interface {
...
+	delete(r.items, id)
+	r.Unlock()
+}
+
+// Entomb sets the tomb bit on a relayItem and schedules a garbage collection.
+func (r *relayItems) Entomb(id uint32, deleteAfter time.Duration) {
+	r.Lock()
+	if r.tombs > _maxRelayTombs {
+		r.logger.WithFields(LogField{"id", id}).Warn("Too many tombstones, deleting relay item immediately.")
+		r.Unlock()
```
prashantv: here and below, `Unlock` before the log

introspection.go
```diff
@@ -157,6 +157,7 @@ type RelayItemState struct {
 	ID                      uint32 `json:"id"`
 	RemapID                 uint32 `json:"remapID"`
 	DestinationConnectionID uint32 `json:"destinationConnectionID"`
+	Tomb                    bool   `json:"tomb"`
```
prashantv: Shouldn't the `TestServer` state tests be failing since the pre-state doesn't have any relay items, but the post-state has tombstoned relay items?

relay.go
```diff
@@ -8,6 +8,13 @@ import (
 	"time"
 )
 
+// _maxRelayTombs is the maximum number of tombs we'll accumulate in a single
+// relayItems.
+const _maxRelayTombs = 1 << 20
+
+// _relayTombTTL is the length of time we'll keep a tomb before GC'ing it.
+const _relayTombTTL = 10 * time.Minute
```
prashantv: this is really high. I think:
 - we should keep them for a short period of time (order of seconds, like 1-5 seconds)
 - reduce the max size

relay.go
```diff
@@ -17,8 +24,85 @@ type RelayHosts interface {
...
+		r.logger.WithFields(LogField{"id", id}).Warn("Can't find relay item to entomb.")
+		r.Unlock()
+		return
+	}
+	r.tombs++
+	item.tomb = true
+	r.items[id] = item
+	r.Unlock()
+
+	time.AfterFunc(deleteAfter, func() { r.Delete(id) })
```
prashantv: (no action needed), long term, do you think we should clear these out in batches?

relay.go
```diff
@@ -56,8 +138,8 @@ func NewRelayer(ch *Channel, conn *Connection) *Relayer {
 	return &Relayer{
 		metrics:  conn.statsReporter,
 		hosts:    ch.relayHosts,
-		outbound: make(map[uint32]relayItem),
-		inbound:  make(map[uint32]relayItem),
+		outbound: newRelayItems(ch.Logger()),
```
prashantv: nit: maybe add "outbound" and "inbound" to the logger for the relays?

relay.go
akshayjshah: Sure, works for me. Since I was picking random numbers, I thought I'd start by erring on the high side.

relay.go
```diff
@@ -134,15 +209,15 @@ func (r *Relayer) handleCallReq(f lazyCallReq) error {
 			ErrField(err),
 			LogField{"hostPort", hostPort},
 		).Warn("Failed to connect to relay host.")
-		// TODO : return an error frame.
-		return nil
+		return ErrNoPeers
```
prashantv: discard this change -- this should be a separate error that needs to be retried. i'll do a separate diff for that

relay.go
akshayjshah: Yes, definitely.

introspection.go
akshayjshah: Hmmm, that's a good point. I'll look into this.

akshayjshah: Added an option to include tombstones in state verification, in case we need it later.

close_test.go
```diff
@@ -78,7 +78,9 @@ func TestCloseNewClient(t *testing.T) {
 }
 
 func TestCloseAfterTimeout(t *testing.T) {
-	testutils.WithTestServer(t, nil, func(ts *testutils.TestServer) {
+	// TODO: Test with relay (once relays handle timeouts).
```
prashantv: fix TODO since relay now handles timeouts?

close_test.go
```diff
@@ -174,7 +176,8 @@ func TestCloseStress(t *testing.T) {
 	for i := 0; i < numHandlers; i++ {
 		wg.Add(1)
 		go func() {
-			testutils.WithTestServer(t, nil, func(ts *testutils.TestServer) {
+			opts := testutils.NewOpts()
```
prashantv: why do you need to create opts here?

close_test.go
```diff
@@ -78,7 +78,9 @@ func TestCloseNewClient(t *testing.T) {
 }
 
 func TestCloseAfterTimeout(t *testing.T) {
-	testutils.WithTestServer(t, nil, func(ts *testutils.TestServer) {
+	// TODO: Test with relay (once relays handle timeouts).
+	opts := testutils.NewOpts()
```
prashantv: just use `nil` as opts below if you don't need to add `SetRelay`

introspection.go
```diff
@@ -318,28 +322,39 @@ func (c *Connection) IntrospectState(opts *IntrospectionOptions) ConnectionRunti
 
 // IntrospectState returns the runtime state for this relayer.
 func (r *Relayer) IntrospectState(opts *IntrospectionOptions) RelayerRuntimeState {
-	r.RLock()
-	defer r.RUnlock()
+	var count int
+	if opts.IncludeTombstones {
+		count = r.inbound.CountAll() + r.outbound.CountAll()
```
prashantv: prefer if we just maintain two separate counters, Count (only active) and Tombstoned. The "IncludeTombstones" should only affect whether we display the tombstoned relays as part of the items list

relay.go
```diff
@@ -17,8 +24,96 @@ type RelayHosts interface {
...
+	r.Lock()
+	r.items[id] = item
+	r.Unlock()
+}
+
+// Delete removes a relayItem completely (without leaving a tombstone).
+func (r *relayItems) Delete(id uint32) {
+	r.Lock()
+	if item, ok := r.items[id]; ok {
+		item.Stop()
```
prashantv: what do you think about doing all extra operations on the tombstone outside of the lock?

relay.go
```diff
@@ -160,24 +246,26 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
...
 
-	r.RLock()
-	item, ok := items[f.Header.ID]
-	r.RUnlock()
+	item, ok := items.Get(f.Header.ID)
 	if !ok {
 		return errors.New("non-callReq for inactive ID")
 	}
+	if item.tomb {
+		// Call timed out, ignore this frame.
```
prashantv: (TODO: we should track the number of response frames we receive after the item has timed out)

close_test.go
akshayjshah: I don't - I just ran the same macro on the whole file without thinking about it. I'll back out the changes we don't need yet.

relay.go
```diff
@@ -188,17 +276,23 @@ func (r *Relayer) addRelayItem(isOriginator bool, id, remapID uint32, destinatio
...
-	r.Lock()
-	delete(items, id)
-	r.Unlock()
+func (r *Relayer) timeoutRelayItem(items *relayItems, id uint32, isOriginator bool) {
+	items.Entomb(id, _relayTombTTL)
+	if isOriginator {
+		// TODO: As above. What's the span in the error frame for?
+		r.conn.SendSystemError(id, nil, ErrTimeout)
+	}
+	r.decPending()
```
prashantv: (feel free to add TODO for this)
we should be more defensive when decrementing this count -- we should only do it if `Entomb` set tomb=true for the first time. Same thing below, we should only decrement if delete actually removed the item and it wasn't already pending. Otherwise, we might want to log a warning since it would mean we have a bug.

relay_messages.go
```diff
@@ -20,12 +20,18 @@
...
+	"encoding/binary"
+	"fmt"
+	"time"
+)
 
 const (
 	_serviceLenIndex  = 1 /* flags */ + 4 /* ttl */ + 25 /* tracing */
 	_serviceNameIndex = _serviceLenIndex + 1
 	_flagsIndex       = 0
+	_ttlLen           = 4
```
prashantv: can you rearrange these and use `_ttlLen` in `_serviceLenIndex`,
```go
const (
  _ttlIndex = 1 /* flags */
  _ttlLen = 4
  _serviceLenIndex = 1 /* flags */ + _ttlLen + 25 /* tracing */
  _serviceNameIndex = _serviceLenIndex + 1
)
```

testutils/test_server.go
```diff
@@ -60,15 +60,17 @@ type TestServer struct {
 	// as part of the TestServer (including the server).
 	channelStates map[*tchannel.Channel]*tchannel.RuntimeState
 
-	verifyOpts *goroutines.VerifyOpts
-	postFns    []func()
+	introspectOpts *tchannel.IntrospectionOptions
```
prashantv: do we need this? we should always be validating with include tombstones.

close_test.go
akshayjshah: Yep, will do.

testutils/test_server.go
akshayjshah: Don't think we need the method to modify the options, but it's nice to specify the options we're using in one place. Introducing this field removed 3 or 4 places we were hard-coding the same options.

relay.go
akshayjshah: Fixed.

relay.go
akshayjshah: Added a TODO.

testutils/test_server.go
```diff
@@ -60,15 +60,17 @@ type TestServer struct {
...
 }
 
 // NewTestServer constructs a TestServer.
 func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
 	ts := &TestServer{
-		TB:            t,
-		channelStates: make(map[*tchannel.Channel]*tchannel.RuntimeState),
+		TB:             t,
+		channelStates:  make(map[*tchannel.Channel]*tchannel.RuntimeState),
+		introspectOpts: &tchannel.IntrospectionOptions{IncludeExchanges: true},
```
prashantv: IncludeTombstones by default, if the test fails, we want to see all the relay items

prashantv:
lgtm

akshayjshah:
@prashantv Rebased to pull in the benchmarking code.

prashantv:
lgtm

PR #348, opened 2016-04-29
Fix flaky TestWriteArg3AfterTimeout

prashantv:
Test was failing because the timeout was too low -- it was timing out
before the handler got the call (at the connection stage).

The test needs the handler to get the call, so increase the timeout.


akshayjshah:
lgtm

blampe:
sure

akshayjshah:
I'll apply this same change to my timeouts diff; I'd rather avoid another rebase.

Issue #261, opened 2016-03-18
Mechanism to hook in to generated server code with objects generated at runtime for Request Logging


IvanMalison:
> Hi @IvanMalison, can you give examples of how you expect this to be used?

See https://code.uberinternal.com/diffusion/EVGOL/browse/master/requestlogger/request_logger.go;44392961db1b3250fb8d2fdb7caceba9add04db4$85

Basically, it should be possible to intercept the arguments that are received by an arbitrary call into the Server object in a uniform way. The existing `Handle` interface does not suffice for this because the only interface it provides is a `thrift.TProtocol` object which can only be read once.

> I'd like to brainstorm to come up with solutions that:

> don't involve use of empty interfaces, at least not in the generated code

I'm pretty skeptical of any dogma against empty interfaces. We are already using `athrift.TStruct` which might as well be empty interfaces since they just have serialization methods. 

> similar to above, not have to do type cases.

I don't think this approach used any type cases. Are you referring to the type assertions that are used once the types are passed back to handleArgs so they can be passed to their respective handlers?

> no perf impact

> Ideally none, but minimal changes to the generated code

This is a strange constraint.

> If there needs more changes to the generated code, then thrift-gen supports templates, so extra code can be generated if the user wants, and the normal generated code should have just the minimal hook.

I'm not sure that templates will suffice in this case, since the generated code needs to have access to objects that are constructed at runtime, elsewhere in the application.
 

--------------------------------------------------
I'm definitely willing to compromise here, but I'm also very skeptical of aggressive performance optimization without benchmarking to back up the idea that performance is a real issue. Do we really think that a signficant portion of a services cpu cycles will be consumed by marshalling between interfaces{} and concrete struct types -- I'd be very surprised to find that the changes that I've made here have even a 10% overall performance impact on even the leanest of services. Premature optimization is the root of all evil and all that stuff. I realize that tchannel-go needs to be quite lean because it is something that will be consumed by absolutely everyone, but for the same reason, it is critically important that the interfaces it provides are convenient and clean. 

An alternative approach that could work to solve the problem of request logging is to add a Method

```AddRequestHook(hook func (ctx Context, args interface{}, resp interface{}, unhandled error))```

to the TChanServer Interface that would populate an instance variable on the server that the generated code would check for and call where appropriate.

The biggest problem with this approach from my perspective is that it makes it much more difficult (possibly impossible) to do something like uncaught exception handling with recover.

prashantv:
[...]

PR #349, opened 2016-05-04
Add Interceptors to TChanServer

IvanMalison:
This is an implementation of the concept proposed in #261. There are some subtle differences between the originally proposed behavior and the behavior in this PR. specifically, the Post methods of ALL interceptors are always called, and all of the Pre methods of the interceptors are called if ANY are called.

### DIFF
--- a/examples/keyvalue/gen-go/keyvalue/tchan-keyvalue.go
+++ b/examples/keyvalue/gen-go/keyvalue/tchan-keyvalue.go
@@ -71,14 +71,16 @@ type tchanAdminServer struct {
 	thrift.TChanServer
 
 	handler TChanAdmin
+
+	interceptorRunner thrift.InterceptorRunner
 }
 
 // NewTChanAdminServer wraps a handler for TChanAdmin so it can be
 // registered with a thrift.Server.
 func NewTChanAdminServer(handler TChanAdmin) thrift.TChanServer {
 	return &tchanAdminServer{
-		NewTChanBaseServiceServer(handler),
-		handler,
+		TChanServer: NewTChanBaseServiceServer(handler),
+		handler:     handler,
 	}
 }
 
@@ -94,6 +96,11 @@ func (s *tchanAdminServer) Methods() []string {
 	}
 }
 
+// RegisterInterceptors registers the provided interceptors with the server.
+func (s *tchanAdminServer) RegisterInterceptorRunner(runner thrift.InterceptorRunner) {
+	s.interceptorRunner = runner
+}
+
 func (s *tchanAdminServer) Handle(ctx thrift.Context, methodName string, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
 	switch methodName {
 	case "clearAll":
@@ -106,31 +113,44 @@ func (s *tchanAdminServer) Handle(ctx thrift.Context, methodName string, protoco
 	}
 }
 
-func (s *tchanAdminServer) handleClearAll(ctx thrift.Context, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+func (s *tchanAdminServer) handleClearAll(ctx thrift.Context, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, retErr error) {
 	var req AdminClearAllArgs
 	var res AdminClearAllResult
+	const serviceMethod = "Admin::clearAll"
 
-	if err := req.Read(protocol); err != nil {
-		return false, nil, err
+	if readErr := req.Read(protocol); readErr != nil {
+		return false, nil, readErr
 	}
 
-	err :=
-		s.handler.ClearAll(ctx)
+	postRun, err := s.interceptorRunner.RunPre(ctx, serviceMethod, &req)
+
+	defer func() {
+		resp = &res
+		retErr = postRun(resp, err)
+		handled = retErr == nil
+		if retErr != nil {
+			resp = nil
+			switch v := retErr.(type) {
+			case *NotAuthorized:
+				if v == nil {
+					retErr = fmt.Errorf("Handler for notAuthorized returned non-nil error type *NotAuthorized but nil value")
+				} else {
+					res.NotAuthorized = v
+					retErr = nil
+					resp = &res
+				}
+			}
+		}
+	}()
 
 	if err != nil {
-		switch v := err.(type) {
-		case *NotAuthorized:
-			if v == nil {
-				return false, nil, fmt.Errorf("Handler for notAuthorized returned non-nil error type *NotAuthorized but nil value")
-			}
-			res.NotAuthorized = v
-		default:
-			return false, nil, err
-		}
-	} else {
+		return
 	}
 
-	return err == nil, &res, nil
+	err =
+		s.handler.ClearAll(ctx)
+
+	return
 }
 
 type tchanKeyValueClient struct {
@@ -191,14 +211,16 @@ type tchanKeyValueServer struct {
 	thrift.TChanServer
 
 	handler TChanKeyValue
+
+	interceptorRunner thrift.InterceptorRunner
 }
 
 // NewTChanKeyValueServer wraps a handler for TChanKeyValue so it can be
 // registered with a thrift.Server.
 func NewTChanKeyValueServer(handler TChanKeyValue) thrift.TChanServer {
 	return &tchanKeyValueServer{
-		NewTChanBaseServiceServer(handler),
-		handler,
+		TChanServer: NewTChanBaseServiceServer(handler),
+		handler:     handler,
 	}
 }
 
@@ -215,6 +237,11 @@ func (s *tchanKeyValueServer) Methods() []string {
 	}
 }
 
+// RegisterInterceptors registers the provided interceptors with the server.
+func (s *tchanKeyValueServer) RegisterInterceptorRunner(runner thrift.InterceptorRunner) {
+	s.interceptorRunner = runner
+}
+
 func (s *tchanKeyValueServer) Handle(ctx thrift.Context, methodName string, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
 	switch methodName {
 	case "Get":
@@ -229,64 +256,96 @@ func (s *tchanKeyValueServer) Handle(ctx thrift.Context, methodName string, prot
 	}
 }
 
-func (s *tchanKeyValueServer) handleGet(ctx thrift.Context, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+func (s *tchanKeyValueServer) handleGet(ctx thrift.Context, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, retErr error) {
 	var req KeyValueGetArgs
 	var res KeyValueGetResult
+	const serviceMethod = "KeyValue::Get"
 
-	if err := req.Read(protocol); err != nil {
-		return false, nil, err
+	if readErr := req.Read(protocol); readErr != nil {
+		return false, nil, readErr
+	}
+
+	postRun, err := s.interceptorRunner.RunPre(ctx, serviceMethod, &req)
+
+	defer func() {
+		resp = &res
+		retErr = postRun(resp, err)
+		handled = retErr == nil
+		if retErr != nil {
+			resp = nil
+			switch v := retErr.(type) {
+			case *KeyNotFound:
+				if v == nil {
+					retErr = fmt.Errorf("Handler for notFound returned non-nil error type *KeyNotFound but nil value")
+				} else {
+					res.NotFound = v
+					retErr = nil
+					resp = &res
+				}
+			case *InvalidKey:
+				if v == nil {
+					retErr = fmt.Errorf("Handler for invalidKey returned non-nil error type *InvalidKey but nil value")
+				} else {
+					res.InvalidKey = v
+					retErr = nil
+					resp = &res
+				}
+			}
+		}
+	}()
+
+	if err != nil {
+		return
 	}
 
 	r, err :=
 		s.handler.Get(ctx, req.Key)
 
-	if err != nil {
-		switch v := err.(type) {
-		case *KeyNotFound:
-			if v == nil {
-				return false, nil, fmt.Errorf("Handler for notFound returned non-nil error type *KeyNotFound but nil value")
-			}
-			res.NotFound = v
-		case *InvalidKey:
-			if v == nil {
-				return false, nil, fmt.Errorf("Handler for invalidKey returned non-nil error type *InvalidKey but nil value")
-			}
-			res.InvalidKey = v
-		default:
-			return false, nil, err
-		}
-	} else {
+	if err == nil {
 		res.Success = &r
 	}
 
-	return err == nil, &res, nil
+	return
 }
 
-func (s *tchanKeyValueServer) handleSet(ctx thrift.Context, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+func (s *tchanKeyValueServer) handleSet(ctx thrift.Context, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, retErr error) {
 	var req KeyValueSetArgs
 	var res KeyValueSetResult
+	const serviceMethod = "KeyValue::Set"
 
-	if err := req.Read(protocol); err != nil {
-		return false, nil, err
+	if readErr := req.Read(protocol); readErr != nil {
+		return false, nil, readErr
 	}
 
-	err :=
-		s.handler.Set(ctx, req.Key, req.Value)
+	postRun, err := s.interceptorRunner.RunPre(ctx, serviceMethod, &req)
+
+	defer func() {
+		resp = &res
+		retErr = postRun(resp, err)
+		handled = retErr == nil
+		if retErr != nil {
+			resp = nil
+			switch v := retErr.(type) {
+			case *InvalidKey:
+				if v == nil {
+					retErr = fmt.Errorf("Handler for invalidKey returned non-nil error type *InvalidKey but nil value")
+				} else {
+					res.InvalidKey = v
+					retErr = nil
+					resp = &res
+				}
+			}
+		}
+	}()
 
 	if err != nil {
-		switch v := err.(type) {
-		case *InvalidKey:
-			if v == nil {
-				return false, nil, fmt.Errorf("Handler for invalidKey returned non-nil error type *InvalidKey but nil value")
-			}
-			res.InvalidKey = v
-		default:
-			return false, nil, err
-		}
-	} else {
+		return
 	}
 
-	return err == nil, &res, nil
+	err =
+		s.handler.Set(ctx, req.Key, req.Value)
+
+	return
 }
 
 type tchanBaseServiceClient struct {
@@ -318,13 +377,15 @@ func (c *tchanBaseServiceClient) HealthCheck(ctx thrift.Context) (string, error)
 
 type tchanBaseServiceServer struct {
 	handler TChanBaseService
+
+	interceptorRunner thrift.InterceptorRunner
 }
 
 // NewTChanBaseServiceServer wraps a handler for TChanBaseService so it can be
 // registered with a thrift.Server.
 func NewTChanBaseServiceServer(handler TChanBaseService) thrift.TChanServer {
 	return &tchanBaseServiceServer{
-		handler,
+		handler: handler,
 	}
 }
 
@@ -338,6 +399,11 @@ func (s *tchanBaseServiceServer) Methods() []string {
 	}
 }
 
+// RegisterInterceptors registers the provided interceptors with the server.
+func (s *tchanBaseServiceServer) RegisterInterceptorRunner(runner thrift.InterceptorRunner) {
+	s.interceptorRunner = runner
+}
+
 func (s *tchanBaseServiceServer) Handle(ctx thrift.Context, methodName string, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
 	switch methodName {
 	case "HealthCheck":
@@ -348,22 +414,36 @@ func (s *tchanBaseServiceServer) Handle(ctx thrift.Context, methodName string, p
 	}
 }
 
-func (s *tchanBaseServiceServer) handleHealthCheck(ctx thrift.Context, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+func (s *tchanBaseServiceServer) handleHealthCheck(ctx thrift.Context, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, retErr error) {
 	var req BaseServiceHealthCheckArgs
 	var res BaseServiceHealthCheckResult
+	const serviceMethod = "baseService::HealthCheck"
 
-	if err := req.Read(protocol); err != nil {
-		return false, nil, err
+	if readErr := req.Read(protocol); readErr != nil {
+		return false, nil, readErr
+	}
+
+	postRun, err := s.interceptorRunner.RunPre(ctx, serviceMethod, &req)
+
+	defer func() {
+		resp = &res
+		retErr = postRun(resp, err)
+		handled = retErr == nil
+		if retErr != nil {
+			resp = nil
+		}
+	}()
+
+	if err != nil {
+		return
 	}
 
 	r, err :=
 		s.handler.HealthCheck(ctx)
 
-	if err != nil {
-		return false, nil, err
-	} else {
+	if err == nil {
 		res.Success = &r
 	}
 
-	return err == nil, &res, nil
+	return
 }
--- a/thrift/interfaces.go
+++ b/thrift/interfaces.go
@@ -45,9 +45,9 @@ type TChanServer interface {
 	Methods() []string
 }
 
-// InterceptorRunnerRegistrar is a target for the registration of an
+// interceptorRunnerRegistrar is a target for the registration of an
 // interceptor runner.
-type InterceptorRunnerRegistrar interface {
+type interceptorRunnerRegistrar interface {
 	// RegisterInterceptorRunner registers the interceptor runner.
 	RegisterInterceptorRunner(InterceptorRunner)
 }
--- a/thrift/interfaces.go
+++ b/thrift/interfaces.go
@@ -44,3 +44,38 @@ type TChanServer interface {
 	// Methods returns the method names handled by this server.
 	Methods() []string
 }
+
+// InterceptorRunnerRegistrar is a target for the registration of an
+// interceptor runner.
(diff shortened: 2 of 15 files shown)


blampe:
What's the use case here? This looks like it overlaps with what @breerly is doing in yarpc.

IvanMalison:
> What's the use case here? This looks like it overlaps with what @breerly is doing in yarpc.

The main use case here is request logging/ handling uncaughts and modifying return values from tchannel requests. I'm not sure that its possible to do this without changes to the generated code so i think that it is probably not exactly the same as what @breerly is doing.

Can you elaborate?

breerly:
@IvanMalison we're adding Middleware to YARPC (java at the moment, go within a week) which has a [different shape](https://github.com/yarpc/yarpc-java/blob/crossdock-errors/yarpc-core/src/main/java/io/yarpc/interceptor/Interceptor.java
):

```java
package io.yarpc.interceptor;

import io.yarpc.Request;
import io.yarpc.Response;

public interface Interceptor {
    Response<?> handle(Request<?> request, InterceptorNode next) throws Exception;
}
```

We'll be implementing the same design for all of YARPC (java, go, python, & node).

We ended up with this style of middleware because:

- Easier to write complex middleware 
- Short cutting, aka not calling "next", is very clear
- Exception handling is standard - aka when calling next check the error or catch the exception
- Testing is easier, instead of calling the methods of an interceptor in the right order, there is only 1 way to call it.
- Composing middleware is simpler because of the single `handle` method.
- The approach you have with `before/after` methods can be implemented on top of the approach we use

breerly:
Created a YARPC-central issue to track middleware impls https://github.com/yarpc/yarpc/issues/57

blampe:
@prashantv / @abhinav do you have thoughts on this?

examples/keyvalue/gen-go/keyvalue/tchan-keyvalue.go
```diff
@@ -94,6 +96,41 @@ func (s *tchanAdminServer) Methods() []string {
...
 }
 
+// RegisterInterceptors registers the provided interceptors with the server.
+func (s *tchanAdminServer) RegisterInterceptors(interceptors ...thrift.Interceptor) {
+	if s.interceptors == nil {
+		interceptorsLength := len(interceptors)
+		s.interceptors = make([]thrift.Interceptor, interceptorsLength, interceptorsLength)
+	}
+
+	s.interceptors = append(s.interceptors, interceptors...)
```
prashantv: I don't think this works correctly.

If `s.interceptors` is `nil`, and the user tries to add 3 interceptors, you create a slice with length = 3 and capacity = 3. Then you append to that list, you're so the length now becomes 6. You have created 3 `nil` entries.

thrift/interfaces.go
```diff
@@ -20,27 +20,73 @@
...
+// TChanClient abstracts calling a Thrift endpoint, and is used by the
+// generated client code.
 type TChanClient interface {
-	// Call should be passed the method to call and the request/response Thrift structs.
-	Call(ctx Context, serviceName, methodName string, req, resp athrift.TStruct) (success bool, err error)
+	// Call should be passed the method to call and the
+	// request/response Thrift structs.
+	Call(
+		ctx Context, serviceName, methodName string, req, resp athrift.TStruct,
+	) (success bool, err error)
```
prashantv: please revert all whitespace changes (I don't want to lose useful information in `git blame` about when and why a specific line was changed).

We don't enforce any 80 character limits in the tchannel-go codebase -- we tend to use ~100 chars for comments, but there's no hard limit.

thrift/thrift-gen/tchannel-template.go
```diff
@@ -118,14 +120,49 @@ func (s *{{ .ServerStruct }}) Service() string {
...
+func (s *{{ .ServerStruct }}) RegisterInterceptors(interceptors ...thrift.Interceptor) {
+	if s.interceptors == nil {
+		interceptorsLength := len(interceptors)
+		s.interceptors = make([]thrift.Interceptor, 0, interceptorsLength)
+	}
+
+	s.interceptors = append(s.interceptors, interceptors...)
+}
+
+func (s *{{ .ServerStruct }}) callInterceptorsPre(ctx {{ contextType }}, method string, args athrift.TStruct) error {
```
prashantv: the templates should be as minimal as possible, all of the logic should live inside of the `thrift` package.

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +179,61 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
...
 
 {{ range .Methods }}
-	func (s *{{ $svc.ServerStruct }}) {{ .HandleFunc }}(ctx {{ contextType }}, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+	func (s *{{ $svc.ServerStruct }}) {{ .HandleFunc }}(ctx {{ contextType }}, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, err error) {
 		var req {{ .ArgsType }}
 		var res {{ .ResultType }}
+		serviceMethod := "{{ .ThriftName }}.{{ .Name }}"
 
-		if err := req.Read(protocol); err != nil {
+		defer func () {
```
prashantv: why are we handling panics here?

prashantv:
@IvanMalison we're currently working on YARPC which will have proper middleware support.

I'm not against Thrift middleware in TChannel, but I want to make sure it's discouraged, and YARPC is the preferred option -- especially since the middlewares would not be compatible.

In terms of this specific change, there's a few issues that are blocking:
 - The generated code tries to do the calls to interceptors, which should really happen in the Thrift library
 - We should only be calling "post" for the interceptors that had their "pre" called.
 - No unit tests

examples/keyvalue/gen-go/keyvalue/tchan-keyvalue.go
IvanMalison: Yeah this was fixed in the actual template (this is generated code here), but I don't think i re ran thrift-gen, which is why this has the old version.

thrift/thrift-gen/tchannel-template.go
IvanMalison: We need code that accesses the private member interceptors here though, which is why I think this code would need to live inside of the template.

thrift/thrift-gen/tchannel-template.go
IvanMalison: I thought we talked about this before. This was one of the requirements of this change. We can't handle panics in the interceptors since they don't actually wrap the call into the handler.

breerly:
Are these server-only middleware?

thrift/thrift-gen/tchannel-template.go
prashantv: Could the interceptors live on the `thrift.Server` as well?

thrift/thrift-gen/tchannel-template.go
prashantv: I think the handling of the panics should happen at the TChannel level, where it kicks off a goroutine. I have ideas for how to handle that, let's do that separately.

thrift/interfaces.go
```diff
@@ -20,27 +20,73 @@
...
+	RegisterInterceptors(...Interceptor)
+}
+
+// PanicErr is returned from Handle when a panic occurs in a handler method
+type PanicErr struct {
+	Value interface{}
+}
+
+func (e PanicErr) Error() string {
+	return fmt.Sprintf("Panic with value %v", e.Value)
```
abhinav: Error messages usually start with lower case.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

thrift/thrift-gen/tchannel-template.go
```diff
@@ -118,14 +120,49 @@ func (s *{{ .ServerStruct }}) Service() string {
...
 }
 
+// RegisterInterceptors registers the provided interceptors with the server.
+func (s *{{ .ServerStruct }}) RegisterInterceptors(interceptors ...thrift.Interceptor) {
+	if s.interceptors == nil {
+		interceptorsLength := len(interceptors)
+		s.interceptors = make([]thrift.Interceptor, 0, interceptorsLength)
+	}
+
+	s.interceptors = append(s.interceptors, interceptors...)
```
abhinav: You can drop the if statement. `nil` is a valid slice to `append`.

prashantv:
We need some sort of interceptors to unblock @IvanMalison's team, since they can't use YARPC right now.

Since we don't want to end up building up large amounts of interceptors that will not be compatible with YARPC, let's discourage the use through naming when adding the Pre/Post interceptors. Maybe something like `DoNotUseAddPre` or something? Open to ideas

thrift/thrift-gen/tchannel-template.go
```diff
@@ -118,14 +120,49 @@ func (s *{{ .ServerStruct }}) Service() string {
...
+
+func (s *{{ .ServerStruct }}) callInterceptorsPre(ctx {{ contextType }}, method string, args athrift.TStruct) error {
+	if s.interceptors == nil {
+		return nil
+	}
+	var firstErr error
+	for _, interceptor := range s.interceptors {
+		err := interceptor.Pre(ctx, method, args)
+		if err != nil && firstErr == nil {
+			firstErr = err
```
abhinav: So if you run into an error while executing a `Pre`, you still run through them all? That isn't a good idea.

Generally, interceptors semantics are:

- Stop at the first error in the 'pre' stage.
- Execute the 'post' of all interceptors that successfully executed and the one that failed **in reverse order**. That is, as soon as an interceptor fails, its post should be executed, then the post of the one before that, and so on. This allows for interceptors to perform clean up in the correct order.
- If there were no failures, simply run the 'post' of all interceptors in the reverse order. 

```
a.pre
    b.pre
        c.pre FAILED
            If c.pre had not failed, we would call the rest of the interceptors here
            followed by the actual handler
        c.post
    b.post
a.post
```

thrift/interfaces.go
IvanMalison: Thats strings, not type names

abhinav: I'm not talking about the type name. I'm talking about "Panic with value". You really want that to be,

    fmt.Sprintf("panic with value: %v", e.Value)

prashantv:
Pushed a branch with some ideas of how to structure this code such that most of the logic lives in `thrift.Server`, minimal changes to `TChanServer` so that existing generated code won't break,
https://github.com/uber/tchannel-go/commit/05feb85c293b60b7184b06368d5b8c202a938a63

In short:
 - The `thrift.Server` will register itself as an `InterceptorRunner` that the generated code will call.
 - The interceptor runner will have a `Pre` that return a `Post` function that will only run interceptors that the Pre ran
 - Panic handling will not be in `thrift.Server` -- the handler can choose to handle the panic in the `Post`, since we will `defer` the `Post`
 - Unit tests will add a new method to `test.thrift` that returns value + exception, and use that for different interceptor test scenarios.

Thanks for making the changes @IvanMalison

IvanMalison:
> Panic handling will not be in thrift.Server -- the handler can choose to handle the panic in the Post, since we will defer the Post

This actually doesn't work since you can't recover on behalf of a deferred function that has called you.

Here's an example that illustrates this point P20323: https://code.uberinternal.com/P20323

prashantv:
@IvanMalison ah right, `recover` only works from the deferred function. So in the generated code, we could have:
```
defer runner.CheckPanic()
defer func() {
  // check err result?
}
```

That should work:
https://play.golang.org/p/Bnu_M-P7QD

And the `thrift.Server` determines whether to handle panics, and can have options for enabling/disabling panic handlers. If that approach works, that should be completely orthogonal to that change and should be done as a separate PR since it involves more changes.

IvanMalison:
bump @prashantv can you look at this?

prashantv:
@IvanMalison I wasn't aware that you had updated the diff since you didn't mention that in your previous comment. I'll look at this today.

Makefile
```diff
@@ -150,6 +150,11 @@ thrift_gen:
 	$(BUILD)/thrift-gen --generateThrift --inputFile hyperbahn/hyperbahn.thrift --outputDir hyperbahn/gen-go
 	rm -rf trace/thrift/gen-go/tcollector && $(BUILD)/thrift-gen --generateThrift --inputFile trace/tcollector.thrift --outputDir trace/thrift/gen-go/
 
+test_mocks:
+	GOPATH=$(OLDGOPATH) mockery -name TChanSimpleService -recursive -dir ./thrift/gen-go/test/ -output ./thrift/mocks
```
prashantv: Can we make this `go generate` rules in `thrift/thrift_test.go` instead?

examples/keyvalue/gen-go/keyvalue/admin.go
```diff
@@ -1,4 +1,4 @@
-// Autogenerated by Thrift Compiler (1.0.0-dev)
+// Autogenerated by Thrift Compiler (0.9.3)
```
prashantv: Can you symlink `go-build`'s thrift binary in to your `PATH` so this is still the `1.0.0-dev` version which is what we've been using internally and will not have this change.

thrift/interfaces.go
```diff
@@ -44,3 +44,44 @@ type TChanServer interface {
...
+type InterceptorPostRunner func(response athrift.TStruct, err error) error
+
+// InterceptorRunner handles the excecution of a collection of
+// Interceptor Pre functions and returns a Interceptor post runner
+// that handles the execution of the Post methods of those same
+// Interceptors.
+type InterceptorRunner interface {
+	// RunPre will run the pre interceptors in order, and return
+	// a function for the post interceptors to run.
+	RunPre(ctx Context, method string, args athrift.TStruct) (
```
prashantv: nit: here and below, we don't follow a 80 char width rule, so I'd prefer to put this all on one line to match the style in the rest of the file

thrift/server.go
```diff
@@ -79,13 +80,46 @@ func (s *Server) Register(svr TChanServer, opts ...RegisterOption) {
...
+// executed before and after each request
+func (s *Server) RegisterInterceptor(interceptor Interceptor) {
+	s.interceptors = append(s.interceptors, interceptor)
+}
+
+// RunPre executes the pre function intercepters registered on the
+// server until one of them fails, and produces a
+// InterceptorPostRunner that will excute the Post methods of all
+// inteceptors whose Pre methods were executed.
+func (s *Server) RunPre(ctx Context, method string, args thrift.TStruct) (InterceptorPostRunner, error) {
```
prashantv: can we have a "fast" path for when there's no interceptors,
```go
if len(s.interceptors) == 0 {
  return noopPostRunner, nil
}
```

and have the noop
```go
func noopPostRunner(...) {
  return nil
}
```

this avoids an extra allocation for the closure when it's not needed.

thrift/interfaces.go
```diff
@@ -44,3 +44,44 @@ type TChanServer interface {
...
+// Interceptor represents operations that should be performed before
+// and after a tchannel request.
+type Interceptor interface {
+	// Pre is called before the handler code for a request is
+	// entered. If an error is returned, the request is short circuited
+	// and the Post method of all interceptors is called with the
+	// error.
+	Pre(ctx Context, method string, args athrift.TStruct) error
+
+	// Post is called when we have a response (or error). If a panic is
```
prashantv: the `panic` stuff is not relevant here right? can we remove that

thrift/tchan-meta.go
```diff
@@ -102,6 +102,10 @@ func (s *tchanMetaServer) Methods() []string {
 	}
 }
 
+// RegisterInterceptors has no effect on the tchan-meta service
+func (s *tchanMetaServer) RegisterInterceptors(interceptors ...Interceptor) {
```
prashantv: nit: can we remove this? should not be needed right

IvanMalison: yep. this was there from before

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,57 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
 }
 
 {{ range .Methods }}
-	func (s *{{ $svc.ServerStruct }}) {{ .HandleFunc }}(ctx {{ contextType }}, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+	func (s *{{ $svc.ServerStruct }}) {{ .HandleFunc }}(ctx {{ contextType }}, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, err error) {
```
prashantv: nit: can we call this `retErr` and keep the other errors as `err`

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,57 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
...
+		postRun, err := s.interceptorRunner.RunPre(ctx, serviceMethod, &req)
 
-		if err != nil {
-			{{ if .HasExceptions }}
-			switch v := err.(type) {
-				{{ range .Exceptions }}
+		defer func () {
+			err = postRun(resp, err)
+			if err != nil {
+				resp = nil
```
prashantv: can you put this in a `default` case inside the `switch`?

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,57 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
...
+				switch v := err.(type) {
+					{{ range .Exceptions }}
 					case {{ .ArgType }}:
 						if v == nil {
-							return false, nil, fmt.Errorf("Handler for {{ .Name }} returned non-nil error type {{ .ArgType }} but nil value")
+							err = fmt.Errorf("Handler for {{ .Name }} returned non-nil error type {{ .ArgType }} but nil value")
+						} else {
+							res.{{ .ArgStructName }} = v
+							err = nil
+							resp = &res
```
prashantv: once you move the `resp = nil` into a default, you shouldn't need this line

thrift/server.go
IvanMalison: noopPostRunner would need to return any incoming error, right? but yeah that sounds good

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,57 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
...
-					default:
-						return false, nil, err
 			}
-			{{ else }}
-				return false, nil, err
-			{{ end }}
-		} else {
-    {{ if .HasReturn }}
-		  res.Success = {{ .WrapResult "r" }}
+		}()
```
prashantv: i don't think `handled` is being set correctly in the case where a post interceptor returns a non-thrift error?

thrift/thrift_test.go
```diff
@@ -226,9 +227,11 @@ func TestClientHostPort(t *testing.T) {
 	assert.Equal(t, "call1", res)
 
 	// When we call using a client that specifies host:port, it should call that server.
-	c2 := gen.NewTChanSecondServiceClient(NewClient(s1ch, s1ch.PeerInfo().ServiceName, &ClientOptions{
-		HostPort: s1ch.PeerInfo().HostPort,
-	}))
+	c2 := gen.NewTChanSecondServiceClient(
```
prashantv: nit: revert whitespace change

prashantv: same throughout this file + diff

thrift/thrift_test.go
```diff
@@ -310,6 +315,137 @@ func TestThriftContextFn(t *testing.T) {
...
 
+type interceptor struct {
+	post func(
+		ctx Context, method string, args, response athrift.TStruct, err error,
+	) error
+	pre func(ctx Context, method string, args athrift.TStruct) error
+}
+
+func (i interceptor) Pre(ctx Context, method string, args athrift.TStruct) error {
+	if i.pre != nil {
```
prashantv: nit: usually these are written to reduce nested code,
```go
if i.pre == nil {
  return nil
}
return i.pre(ctx, method, args)
```

thrift/thrift_test.go
```diff
@@ -310,6 +315,137 @@ func TestThriftContextFn(t *testing.T) {
...
+		return i.pre(ctx, method, args)
+	}
+	return nil
+}
+
+func (i interceptor) Post(
+	ctx Context, method string, args, response athrift.TStruct, err error,
+) error {
+	if i.post != nil {
```
prashantv: same as above, can reduce nesting for this

Makefile
IvanMalison: Do you mean just as a comment? This needs to be executed before that code is compiled right?

thrift/thrift_test.go
```diff
@@ -310,6 +315,137 @@ func TestThriftContextFn(t *testing.T) {
...
+
+		ctx, cancel := NewContext(time.Second)
+		defer cancel()
+
+		result, err := args.c1.Throws(ctx, "test")
+
+		assert.Equal(t, result, "")
+		assert.Equal(t, secondTransformedError, err)
+
+		firstInterceptor.AssertExpectations(t)
```
prashantv: nit: you could do
```go
mock.AssertExpectationsForObjects(t, firstInterceptor, errorInterceptor, uncalledInterceptor)
```
https://godoc.org/github.com/stretchr/testify/mock#AssertExpectationsForObjects

Makefile
prashantv: It's a comment that works when you run `go generate`,
https://blog.golang.org/generate

thrift/thrift-gen/tchannel-template.go
prashantv: we should try to add a unit test to catch the fact that handled is not updated in the defer

thrift/thrift-gen/tchannel-template.go
IvanMalison: you mean the resp = nil?

The reason I moved it outside of that is we have the 

{{ if .HasExceptions }} wrapping the switch, so that line wont appear in methods that have no exceptions.

prashantv:
This is looking pretty good @IvanMalison -- just a few small comments, no major changes needed. Thanks for the change.

thrift/server.go
prashantv: yes, good point, it would have to return the incoming error

thrift/thrift-gen/tchannel-template.go
prashantv: Ah, I see. fair enough, leave this as is.

thrift/thrift-gen/tchannel-template.go
IvanMalison: yeah thats how it was originally. its just less LOC to do it this way.

dont feel strongly so ill change it if you do

prashantv: Can leave this as is.

thrift/interfaces.go
IvanMalison: Is there any char width rule at all? How do you decide when to wrap a line?

prashantv: I try to keep comments < 100 chars. If a line is like > 120, then i'll break it up at a place that seems logical

no hard rule though

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,59 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
...
-
-		if err != nil {
-			{{ if .HasExceptions }}
-			switch v := err.(type) {
-				{{ range .Exceptions }}
+		postRun, err := s.interceptorRunner.RunPre(ctx, serviceMethod, &req)
+
+		defer func () {
+			resp = &res
+			retErr = postRun(resp, err)
```
IvanMalison: I'm not sure why this is showing up at a different indentation level.

IvanMalison:
Updated with comments

thrift/interfaces.go
```diff
@@ -44,3 +44,38 @@ type TChanServer interface {
 	// Methods returns the method names handled by this server.
 	Methods() []string
 }
+
+// InterceptorRunnerRegistrar is a target for the registration of an
+// interceptor runner.
+type InterceptorRunnerRegistrar interface {
```
prashantv: nit: i think we can keep this type private for now.

thrift/thrift-gen/tchannel-template.go
prashantv: it looks like it's 2 tabs indented vs 3 above, seems like it needs an extra tab

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,59 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
 }
 
 {{ range .Methods }}
-	func (s *{{ $svc.ServerStruct }}) {{ .HandleFunc }}(ctx {{ contextType }}, protocol athrift.TProtocol) (bool, athrift.TStruct, error) {
+	func (s *{{ $svc.ServerStruct }}) {{ .HandleFunc }}(ctx {{ contextType }}, protocol athrift.TProtocol) (handled bool, resp athrift.TStruct, retErr error) {
```
prashantv: I think `handled` should actually be called `handlerSuccess` or just `success`.

thrift/thrift-gen/tchannel-template.go
```diff
@@ -142,46 +149,59 @@ func (s *{{ .ServerStruct }}) Handle(ctx {{ contextType }}, methodName string, p
...
+			handled = retErr == nil
+			if retErr != nil {
+				resp = nil
+				{{ if .HasExceptions }}
+				switch v := retErr.(type) {
+					{{ range .Exceptions }}
 					case {{ .ArgType }}:
 						if v == nil {
-							return false, nil, fmt.Errorf("Handler for {{ .Name }} returned non-nil error type {{ .ArgType }} but nil value")
+							retErr = fmt.Errorf("Handler for {{ .Name }} returned non-nil error type {{ .ArgType }} but nil value")
```
prashantv: nit: not changed by you, but errors are supposed to be lowercase, can you make `Handler` `handler`

thrift/thrift_test.go
prashantv: nit: this isn't done yet

prashantv:
functionally lgtm, just some formatting things to fix up

IvanMalison:
oops need to rerun thrift-gen

prashantv:
@IvanMalison Is there anything you need to wrap up this change? I left some minor formatting changes in the last round which weren't addressed, but functionally things looked good.

prashantv:
Closing stale review.

PR #351, opened 2016-05-04
Use glide instead of godep to manage dependencies

prashantv:
Glide is cleaner to use, and separates the version we want vs the commit
hash that we're locked to.

In our Makefile, we don't use the vendor directory, but install
dependencies to the GOPATH. This keeps it compatible with go1.4, but the
goal was not for compatibility. This was required because of thrift-gen
tests. thrift-gen tests build Thrift generated code to a temporary
directory, which references the tchannel/thrift package. If we use the
vendor directory, the tchanne/thrift package references the vendored
version of Apache's thrift library, and the generated code references
the non-vendored version fo the dependencies. This leads to errors like:

```
  *S1M1Args does not implement "github.com/uber/tchannel-go/vendor/github.com/apache/thrift/lib/go/thrift".TStruct (wrong type for Read method)
          have Read("github.com/apache/thrift/lib/go/thrift".TProtocol) error
          want Read("github.com/uber/tchannel-go/vendor/github.com/apache/thrift/lib/go/thrift".TProtocol) error
```
The fix it to not use the vendor directory when running tests.

@akshayjshah


breerly:
Hard to argue - thanks @prashantv 

:shipit:

akshayjshah:
:100: lgtm

PR #355, opened 2016-05-11
Avoid creating a context on every request

prashantv:
The context is only required if we need to create a new connection. Move
the context creation to creating a new connection.

This reduces per-request allocations from 20 to 5.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -24,6 +24,7 @@ import (
 	"container/heap"
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/uber/tchannel-go/atomic"
 	"github.com/uber/tchannel-go/trand"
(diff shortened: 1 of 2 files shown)


peer.go
```diff
@@ -345,11 +346,20 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 	}
 
 	// No active connections, make a new outgoing connection.
-	c, err := p.Connect(ctx)
-	if err != nil {
-		return nil, err
+	return p.Connect(ctx)
+}
+
+// getConnectionRelay gets a connection, and uses the given timeout if a new
```
akshayjshah: nit: method name changed after you wrote comment.

relay.go
```diff
@@ -217,16 +217,15 @@ func (r *Relayer) handleCallReq(f lazyCallReq) error {
 	}
 	peer := r.peers.GetOrAdd(hostPort)
 
-	ctx, cancel := NewContext(time.Second)
-	defer cancel()
-
-	remoteConn, err := peer.GetConnection(ctx)
+	// TODO: Should connections use the call timeout? Or a separate timeout?
```
akshayjshah: mmm, I think we should use a separate timeout. Reasonable application-level timeouts are much, much longer than reasonable connection timeouts.

peer.go

relay.go
prashantv: Yeah, these really should be completely orthogonal. I think we want to get to:
 - separate connection management that limits the number of connection attempts we are making and has separate timeouts
 - in the call request path, if we need a connection, we pass how long we're willing to wait to get a connection

And in general, there's a lot of connection management (e.g., when should connections be closed)

prashantv:
Updated with fix for wrong method name

PR #356, opened 2016-05-12
relay: Support lazily reading method to determine destination

prashantv:
Instead of passing a string with service, instead abstract the frame via
an interface that allows accessing the service and method.

Update tests to ensure this works regardless of the number of headers,
whether the frame contains a checksum, etc.

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -51,6 +51,31 @@ func (f lazyCallReq) Service() string {
 	return string(f.Payload[_serviceNameIndex : _serviceNameIndex+l])
 }
 
+// Method returns the name of the method being called. It panics if called for
+// a non-callReq frame.
+func (f lazyCallReq) Method() string {
+	serviceLen := f.Payload[_serviceLenIndex]
+
+	// nh:1 (hk~1 hv~1){nh}
+	headerStart := _serviceLenIndex + 1 /* length byte */ + serviceLen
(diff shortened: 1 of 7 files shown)


relay_messages.go
```diff
@@ -51,6 +51,31 @@ func (f lazyCallReq) Service() string {
...
+	// csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2
+	checkSumType := ChecksumType(f.Payload[cur])
+	cur += 1 /* checksum */ + checkSumType.ChecksumSize()
+
+	// arg1~2
+	arg1Len := int(binary.BigEndian.Uint16(f.Payload[cur : cur+2]))
+	cur += 2
+	arg1 := f.Payload[cur : cur+arg1Len]
+	return string(arg1)
+}
```
akshayjshah: If we're going to be using these more heavily, should we cache the results on the frame?

akshayjshah:
o/w lgtm

relay_messages.go
prashantv: Good question, so I forgot to mention in my description but my plan is:
 - eagerly parse all fields we know we'll need (e.g. for stats or routing, so service, caller method, timeout) and store
 - use a pointer for the `lazyCallFrame` since it's no longer tiny + pool these objects

akshayjshah:
Cool, sounds good.

PR #357, opened 2016-05-12
Expose the peer information for the outbound call.

nomis52:


outbound.go
```diff
@@ -234,6 +234,11 @@ func (call *OutboundCall) Arg3Writer() (ArgWriter, error) {
 	return call.arg3Writer()
 }
 
+// RemotePeer returns the peer info.
```
prashantv: nit: "returns the peer info." to "returns the peer info of the selected connection for this call" maybe?

prashantv:
can you rebase to get rid of the merge commits in the commit history?

outbound.go
nomis52: Done. See https://github.com/uber/tchannel-go/pull/358

PR #359, opened 2016-05-12
Create a package for relaying interfaces

akshayjshah:
Also, add `vendor` to `.gitignore`. Not sure how we missed that...


prashantv:
lgtm

PR #360, opened 2016-05-12
Make `make benchmark` only run benchmarks

blampe:
I noticed this was running some other tests that were flaking locally (TestWriteArg3AfterTimeout). To keep things clean I'd rather retrict `make benchmark` to only benchmark tests.

@akshayjshah @prashantv

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ test: clean setup install_test
 
 benchmark: clean setup
 	echo Running benchmarks:
-	go test $(PKGS) -bench=. -parallel=4
+	go test $(PKGS) -bench=. -parallel=4 -run ^$
 
 cover_profile: clean setup
 	@echo Testing packages:


Makefile
```diff
@@ -87,7 +87,7 @@ test: clean setup install_test
 
 benchmark: clean setup
 	echo Running benchmarks:
-	go test $(PKGS) -bench=. -parallel=4
+	go test $(PKGS) -bench=. -parallel=4 -run NONE
```
prashantv: nit: `run` takes a regex, so you could do `-run ^$`

prashantv:
thanks for fixing this! lgtm

Makefile
akshayjshah: Don't forget that Make needs `$` escaped, so you'll want `-run="^$$"`.

blampe: womp, I'll change this back to NONE because it's more obvious what's going on.

prashantv: goddamn it make! yep NONE is more readable :)

PR #362, opened 2016-05-17
Allow Copy to be used outside of tests

prashantv:
There's no other supported way to iterate over all the peers in a
PeerList, so make Copy the supported way to access peers

@akshayjshah @yurishkuro


breerly:
:shipit:

PR #363, opened 2016-05-18
Remove duplicate callOptions nil check

prashantv:
Since we already check at the top, the second check is unnecessary.

Thanks for catching this @yurishkuro 
@akshayjshah


akshayjshah:
lgtm

PR #365, opened 2016-05-18
Fix peer score not being calculated when adding outbound connections

### LINKED ISSUE #364: Peer score is not calculated when a new outbound connection is created
If a connection is created using `GetConnection` and not used immediately (e.g. it's not created in response to a call or ping, but because the user explicitly created a connection), we weren't recalculating the peer's score.

This only affects outbound connections, inbound connections have their score calculated correctly as soon as the connection is created.

cc @yurishkuro

prashantv:
Use the same connectionEvents.onActive path for adding outbound
connections to a peer, and update the score. This makes the
inbound and outbound paths similar.

Fixes #364

cc @yurishkuro

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -404,7 +404,7 @@ func (ch *Channel) serve() {
 
 		// Register the connection in the peer once the channel is set up.
 		events := connectionEvents{
-			OnActive:           ch.incomingConnectionActive,
+			OnActive:           ch.inboundConnectionActive,
 			OnCloseStateChange: ch.connectionCloseStateChange,
 			OnExchangeUpdated:  ch.exchangeUpdated,
 		}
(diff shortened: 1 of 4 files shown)


yurishkuro:
lgtm

channel.go
```diff
@@ -540,20 +541,35 @@ func (ch *Channel) updatePeer(p *Peer) {
...
+func (ch *Channel) outboundConnectionActive(c *Connection) {
+	c.log.Debugf("Add outbound connection as an active peer for %v", c.remotePeerInfo.HostPort)
+	p := ch.rootPeers().GetOrAdd(c.remotePeerInfo.HostPort)
+	if err := p.AddOutboundConnection(c); err != nil {
+		c.log.WithFields(LogField{"remoteHostPort", c.remotePeerInfo.HostPort}).
+			Warn("Failed to add outbound connection to peer")
+	}
+
+	ch.updatePeer(p)
+}
```
akshayjshah: Where are we adding the connection to the channel's mutable conns map?

prashantv: Refactored things a little.

prashantv:
Updated

akshayjshah:
:shipit: Love it.

PR #366, opened 2016-05-18
Fix flaky TestWriteAfterConnectionError

prashantv:
Test is flaky because the addRelay uses a fixed set of options which
enables log verification, but the caller test explicitly is causing
errors that can cause warnings.

Update TestServer to use the top-level log verification options for the
relay that's created.

### DIFF
--- a/testutils/test_server.go
+++ b/testutils/test_server.go
@@ -77,8 +77,8 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
 	}
 
 	ts.NewServer(opts)
-	if opts.IncludeRelay {
-		ts.addRelay()
+	if opts != nil && opts.IncludeRelay {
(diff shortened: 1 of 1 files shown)


testutils/test_server.go
```diff
@@ -78,7 +78,11 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
 
 	ts.NewServer(opts)
 	if opts.IncludeRelay {
-		ts.addRelay()
+		var relayLogOpts LogVerification
+		if opts != nil {
```
akshayjshah: Can `opts` be nil here? Seems like we'd already have paniced on L80. Maybe `NewTestServer` should handle nil options at the top.

prashantv: 👍 moved `!= nil` check up

akshayjshah:
lgtm

PR #368, opened 2016-05-19
Fix flaky TestPeerScoreOnNewConnection

prashantv:
Test failed due to the score not being updated by the time the check
runs for inbound connections.


akshayjshah:
lgtm

Issue #125, opened 2015-12-22
init-handling error testing scenarios


prashantv:
Migrated from https://github.com/uber/tchannel/issues/279

__client side__
- [ ] server returns unsupported protocol version
- [ ] server does not return supported headers
- [ ] server hangs up after we've sent the init-req
- [ ] server sends us a message before we've sent the init-req
- [ ] server sends something other than an init-res in response to our init-req
- [ ] server sends a protocol error in response to our init-req

__server side__
- [x] client sends an unsupported protocol version
- [x] client does not include supported headers
- [ ] client hangs up before we've sent the init-res
- [ ] client sends something other than an init-req as the first message
- [ ] client sends bad data on init-req

PR #369, opened 2016-05-19
Validate initRes headers and stop exchanges on errors

prashantv:
This fixes the following issues:
 - we weren't validating the `initRes` headers returned from the remote side
 - protocol errors didn't cause all exchanges to be stopped

Addresses some of the scenarios in #125


akshayjshah:
lgtm

PR #373, opened 2016-05-23
Remove unused method which isn't correct

prashantv:
cc @akshayjshah


breerly:
*removes unused unexported method* 👏

PR #374, opened 2016-05-23
relay: Use connection's logger which includes connID and addresses

prashantv:
Right now, we're using the channel logger, so we don't have any unique information for each `Relayer` even though we have multiple relayers (one per conneciton).

cc @akshayjshah


breerly:
:shipit:

PR #377, opened 2016-05-24
relay: Fix race between connection close and new call

### LINKED ISSUE #370: Relay: race between items being added and the connection closing
To avoid races between relay items and connection state changes, we need to atomically increment pending and check the connection state. Similarly, the connection close checks needs to check the pending count atomically and change the state to closed.

prashantv:
Currently, relay items are added without checking the connection state.
This can lead to new items being added while a connection is closing.

This change makes checking the connection state atomic with incrementing
the pending count. This ensures that only active connections can accept
new incoming calls.

Fixes #370

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -214,11 +214,23 @@ func (r *Relayer) Receive(f *Frame, fType frameType) {
 	}
 }
 
-func (r *Relayer) handleCallReq(f lazyCallReq) error {
+func (r *Relayer) canHandleNewCall() bool {
+	var canHandle bool
+	r.conn.withStateRLock(func() error {
+		canHandle = r.conn.state == connectionActive
+		if canHandle {
+			r.pending.Inc()
+		}
+		return nil
+	})
+	return canHandle
+}
+
+func (r *Relayer) getDestination(f lazyCallReq) (*Connection, bool, error) {
 	if _, ok := r.outbound.Get(f.Header.ID); ok {
 		r.logger.WithFields(LogField{"id", f.Header.ID}).Warn("received duplicate callReq")
 		// TODO: this is a protocol error, kill the connection.
-		return errors.New("callReq with already active ID")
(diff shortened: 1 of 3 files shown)


relay.go
```diff
@@ -214,7 +214,30 @@ func (r *Relayer) Receive(f *Frame, fType frameType) {
 	}
 }
 
+func (r *Relayer) canHandleNewCall(outbound bool) bool {
+
```
abhinav: nit: delete empty line

relay_test.go
```diff
@@ -6,11 +6,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	. "github.com/uber/tchannel-go"
+
+	"github.com/uber/tchannel-go/atomic"
```
breerly: How generic is this package? Perhaps move to uber-go?

relay_test.go
```diff
@@ -160,3 +163,46 @@ func TestErrorFrameEndsRelay(t *testing.T) {
 		assert.Equal(t, ErrCodeBadRequest, se.Code(), "Expected BadRequest error")
 	})
 }
+
+// Trigger a race between receiving a new call and a connection closing
+// by closing the relay while a lot of background calls are being made.
+func TestRaceCloseWithNewCall(t *testing.T) {
```
breerly: How does this test behave with the implementation backed out? Hangs forever?

relay_test.go
prashantv: already in progress!
https://github.com/uber-go/atomic/pull/1

relay_test.go
prashantv: Sometimes causes races, sometimes panics.

prashantv: Added more clients making concurrent calls so now it always panics without this change.

akshayjshah:
lgtm.

PR #378, opened 2016-05-25
benchmem by default and only test 1 core

blampe:
I think this is more in-line with how we were measuring these numbers.

@prashantv @akshayjshah

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ test: clean setup install_test
 
 benchmark: clean setup
 	echo Running benchmarks:
-	go test $(PKGS) -bench=. -parallel=1 -benchmem -run NONE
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -87,7 +87,7 @@ test: clean setup install_test
 
 benchmark: clean setup
 	echo Running benchmarks:
-	go test $(PKGS) -bench=. -parallel=4 -run NONE
+	go test $(PKGS) -bench=. -parallel=1 -benchmem -run NONE
```
prashantv: we should replace `-parallel=1` with `-cpu=`

`-parallel` just makes tests run in parallel (which we don't really enable anyway).

prashantv:
lgtm with small change

akshayjshah:
lgtm - test failure is a known race. I'm just gonna merge this.

PR #379, opened 2016-05-26
Add lazy wrappers for call res and error frames

akshayjshah:
To track relaying metrics properly, we need to know whether a relayed
frame indicates an error on the caller's side, the callee's side, or the
forwarding plane. This PR extends the `lazyCallReq` pattern to add
convenient accessors to the call res and error frame types.

### DIFF
--- a/relay_messages_test.go
+++ b/relay_messages_test.go
@@ -20,6 +20,7 @@
 package tchannel
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


relay_messages_test.go
```diff
@@ -89,12 +192,6 @@ func TestLazyCallReqRejectsOtherFrames(t *testing.T) {
 	}, "Should panic when creating lazyCallReq from non-callReq frame.")
```
prashantv: nit: can we add a general `Panics` function for each of the lazy types, I want to make sure we don't miss the validation there.

This test can probably be simplified now to use the other types to generate a frame and verify that it panics on that frame, rather than using init req.

akshayjshah: Done. Mind taking another look?

prashantv:
lgtm

PR #380, opened 2016-05-26
Fix conn.onCloseStateChange not called on conn.Close

prashantv:
If conn.Close closes the connection immediately (because of the
connection state), then we closed the sendCh but we did not call
the onCloseStateChange event.

Avoid duplication of the close logic, and instead set the connection
state to connectionInboundClosed. The existing call to checkExchanges
will check the state, transition it to connectionClosed if possible
and go through the standard shutdown path (which will call the
onCloseStateChange event).


akshayjshah:
lgtm

PR #381, opened 2016-05-26
Always call checkExchanges when we decrement pending

prashantv:
checkExchanges wasn't called if we decremented pending after
selecting a remote connection that is not active.

If this pending call is the last call, then we never end up
closing the relay connection which causes tests to fail sometimes
because the relay did not close.


akshayjshah:
lgtm

PR #382, opened 2016-05-26
Improve introspection and logging when tests fail

prashantv:


testutils/test_server.go
```diff
@@ -252,7 +253,7 @@ func (ts *TestServer) waitForChannelClose(ch *tchannel.Channel) {
 	started := time.Now()
 
 	var state tchannel.ChannelState
-	for i := 0; i < 50; i++ {
+	for i := 0; i < 60; i++ {
```
akshayjshah: Is 60 somehow meaningful, or is it just a random number that's larger than 50?

testutils/test_server.go
```diff
@@ -28,6 +28,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
```
akshayjshah: nit: this import belongs in the block below.

testutils/test_server.go
prashantv: Just a number larger than 50

prashantv:
Updated with import fix

PR #383, opened 2016-05-26
introspection: Add test coverage

prashantv:
Excercise the introspection path to ensure it works and doesn't cause
any unexpected failures. This also gives us coverage on the
introspection code.

Turns out we're not using the `PeerList.IntrospectState` method, so removing it.

### DIFF
--- a/introspection_test.go
+++ b/introspection_test.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


introspection_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	. "github.com/uber/tchannel-go"
+
+	"github.com/uber/tchannel-go/json"
+	"github.com/uber/tchannel-go/testutils"
```
akshayjshah: nit: import grouping seems off.

akshayjshah:
lgtm

PR #384, opened 2016-05-26
Don't call onActive for connection that aren't active

prashantv:
When waiting for an initRes, if the connection is closed, then the
state change fails. Previously, we ignored the state change failure
and called onActive. This caused inactive connections to be added to
the channel's connection list and peer list which would never be closed
since the channel may already have closed.

The fix ensures that any unexpected state while changing the state
will cause the connection to return an error and onActive will not
be called.


akshayjshah:
lgtm

PR #385, opened 2016-05-26
Only add new active connections if channel is active

prashantv:
We previously checked if the channel was active using a read lock, but
this could go stale, as the channel could be closed between the time the
check was done and the connection being added -- and the connection
would be left in an active state, causing the channel to never close.

To avoid this, we now check the state and add the connection atomically.

If the connection is not added, it means the channel is closed, and so
we close the connection to avoid leaving a rogue TCP connection.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -534,8 +534,38 @@ func (ch *Channel) updatePeer(p *Peer) {
 	p.callOnUpdateComplete()
 }
 
+// addConnection adds the connection to the channel's list of connection
+// if the channel is in a valid state to accept this connection. It returns
+// whether the connection was added.
+func (ch *Channel) addConnection(c *Connection, direction connectionDirection) bool {
+	ch.mutable.Lock()
+	defer ch.mutable.Unlock()
+
+	switch state := ch.mutable.state; state {
+	case ChannelStartClose:
+		// Outbound connections are allowed in ChannelStartClose, but the
+		// connection state should be set to connectionStartClose.
+		if direction != outbound {
+			return false
+		}
+	case ChannelClient, ChannelListening:
+	default:
+		return false
+	}
+
+	ch.mutable.conns[c.connID] = c
+	return true
+}
+
 func (ch *Channel) connectionActive(c *Connection, direction connectionDirection) {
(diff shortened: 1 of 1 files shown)


channel.go
```diff
@@ -534,8 +523,49 @@ func (ch *Channel) updatePeer(p *Peer) {
 	p.callOnUpdateComplete()
 }
 
+// addConnection adds the connection to the channel's list of connection
+// if the channel is in a valid state to accept this connection. It returns
+// whether the connection was added.
```
akshayjshah: nit: update doc comment to mention both return values.

channel.go
```diff
@@ -534,8 +523,49 @@ func (ch *Channel) updatePeer(p *Peer) {
...
+	}
+	if setStartClose {
+		// TODO(prashant): If Connect is called, but no outgoing calls are made, then this connection
+		// will block Close, as it will never get cleaned up.
+		c.withStateLock(func() error {
+			c.state = connectionStartClose
+			return nil
+		})
+		c.log.Debugf("Channel is in start close, set connection to start close")
+	}
```
akshayjshah: Rather than having `addConnection` return two values, one of which triggers a state change in the connection, why not handle the connection state change inside `addConnection` itself?

channel.go
prashantv: 👍

channel.go
prashantv: `addConnection` holds the channel lock and so:
 - I want to minimize code run within the lock
 - Avoid functionality which needs another lock while inside of the channel lock

In an upcoming diff, we will actually need the connection lock even with the channel lock held
At that point, we should look into whether it should move into `addConnection`.

As part of this change though, I want to minimize the changes to the locking chain.

prashantv:
Updated. good question re: changing the state in `addConnection`, would like to defer that (since it changes locking semantics) to a future change which affects locking semantics.

channel.go
akshayjshah: Fair enough.

prashantv:
Test failures caught a real issue with the change, turns out we can't move the connection state change for `startClose`. The `onActive` callback happens before `checkExchanges`, and since the state is `startClose`, and there's no pending calls, the connection is closed.

Reverted the start-close related logic, but the remaining logic is the same.

PR #387, opened 2016-05-28
Don't allow outgoing connections while a channel is closing

prashantv:
Previously, we allowed creating a new connection to be made during
shutdown, but this had a few drawbacks:
- Channel may never close in specific scenarios (connect without call
  being made)
- Code had hacks to set the connection state specifically for this
  scenario

There doesn't seem to be a strong benefit in supporting this, as it
slows down the shutdown path.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -472,9 +472,6 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string) (*Connection, e
 	switch state := ch.State(); state {
 	case ChannelClient, ChannelListening:
 		break
-	case ChannelStartClose:
-		// We still allow outgoing connections during Close, but the connection has to immediately
-		// be Closed after opening
 	default:
 		ch.log.Debugf("Connect rejecting new connection as state is %v", state)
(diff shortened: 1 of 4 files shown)


channel.go
```diff
@@ -546,13 +532,8 @@ func (ch *Channel) addConnection(c *Connection, direction connectionDirection) b
 	}
 
 	switch state := ch.mutable.state; state {
```
akshayjshah: Is breaking out of a short switch like this preferred to an if?

```
if state := ch.mutable.state; state != ChannelClient && state != ChannelListening {
    return false
}
```

prashantv: I tend to use `switch` if there's more than one conditional, I find it easier to read. The `if` doesn't clearly show "these 2 states are fine, everything else is bad" without a lot more parsing

akshayjshah: Fair enough. I'll add that to my bag of Go idioms.

akshayjshah:
lgtm

PR #388, opened 2016-05-31
Reproduce race between Receive and timeout, fix race by not closing sendCh

### LINKED ISSUE #372: Relay: race between Receive and timeouts
During a close, it's possible to cause a panic because of the following race between Receive and timeout:
 - G1: Receive gets a frame, checks the relay item's `tomb` bit which is false
 - G2: mex times out, pending count is decremented
 - G2: `checkExchanges` closes the connection, closes `sendCh` since ref count is 0
 - G1: Receive tries to send the frame to `sendCh`. which is now closed

There's a couple of options to avoid this race::
- keep a refcount of how many "active" goroutines are using the relay item (this needs to be an atomic increment + check `tomb`), and only decrement pending when this is 0
- keep a single refcount of running goroutines on the `Relayer` and only close `sendCh` once this refcount is 0.

prashantv:
See commit messages for more information.

Fixes #372

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -146,6 +146,7 @@ type Connection struct {
 	localPeerInfo   LocalPeerInfo
 	remotePeerInfo  PeerInfo
 	sendCh          chan *Frame
+	stopCh          chan struct{}
 	state           connectionState
 	stateMut        sync.RWMutex
 	inbound         *messageExchangeSet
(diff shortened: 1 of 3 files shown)


prashantv:
Updated, need to drain `sendCh` before we return from `writeFrames`.

Tests are now passing, ready to review.

akshayjshah:
lgtm

PR #389, opened 2016-05-31
Add Caller method to lazyCallReq

akshayjshah:
Add a method to `lazyCallReq` to extract the caller name from the variable-length transport headers. Along the way, fix some bugs in the flag logic for the relay messages - we (mostly I) were routinely checking flags incorrectly.


prashantv:
lgtm, little scary that the bit-flag logic in tests was completely off.

PR #393, opened 2016-06-02
Fix channel close not checking state under write lock

prashantv:
It's possible for multiple connections to close and trigger
connectionCloseStateChange from multiple goroutines.

This could lead to:

G1: C1 closes, and runs connectionCloseStateChange
G1: Finds minState connectionInboundClosed
G1: updateTo = ChannelInboundClosed
<g1 scheduled off before it completes>
G2: C2 closes, and runs connectionCloseStateChange
G2: Finds minState connectionClosed
G2: updateTo = ChannelClosed
G2: Updates channel state to ChannelClosed
<g1 is rescheduled>
G1: Grabs state lock, and sets state to ChannelInboundClosed

Since all connections are closed, the channel will be stuck in
ChannelInboundClosed.

We should be checking that the decision we made is not stale by
comparing the state matches what we had before the lock when making
the change.

Failure noticed here:
https://travis-ci.org/uber/tchannel-go/jobs/134879684

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -583,6 +583,16 @@ func (ch *Channel) removeClosedConn(c *Connection) {
 	ch.mutable.Unlock()
 }
 
+func (ch *Channel) getMinConnectionState() connectionState {
+	minState := connectionClosed
+	for _, c := range ch.mutable.conns {
+		if s := c.readState(); s < minState {
+			minState = s
+		}
+	}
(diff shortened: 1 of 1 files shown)


akshayjshah:
lgtm - sorry, looked at this yesterday and apparently didn't leave a comment.

PR #396, opened 2016-06-03
Remove atomic package and use uber-go/atomic

### LINKED ISSUE #395: Migrate to using uber-go/atomic

prashantv:
Mostly import changes. Updated peer_test.go to use
uber-go/atomic instead of sync/atomic.

cc @akshayjshah 

Fixes #395


abhinav:
:shipit:

PR #397, opened 2016-06-03
Add relaying stats

akshayjshah:
I'm going to re-open this as a stacked diff (or a series of smaller PRs), but here's everything.

This PR adds stats to the relaying path, and I've started to add stats-related assertions to the relaying and connection tests.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -199,18 +199,23 @@ func (r *Relayer) Receive(f *Frame, fType frameType) {
 	// If we receive a response frame, we expect to find that ID in our outbound.
 	// If we receive a request frame, we expect to find that ID in our inbound.
 	items := r.receiverItems(fType)
-	isOriginator := fType == responseFrame
 
 	item, ok := items.Get(f.Header.ID)
 	if !ok {
 		r.logger.WithFields(
 			LogField{"ID", f.Header.ID},
 		).Warn("Received a frame without a RelayItem.")
-	} else if isOriginator {
-		if msg := relayErrMsg(f); msg != "" {
-			item.stats.Failed(msg)
-		} else if indicatesSuccess(f) {
+		return
+	}
+
+	// call res frames don't include the OK bit, so we can't wait until the last
+	// frame of a relayed RPC to determine if the call succeeded.
+	isOriginator := fType == responseFrame
+	if isOriginator {
+		if succeeded, failed, failMsg := determinesCallSuccess(f); succeeded {
 			item.stats.Succeeded()
+		} else if failed {
+			item.stats.Failed(failMsg)
 		}
 	}
 
@@ -416,29 +421,25 @@ func errUnknownGroup(group string) error {
 	return NewSystemError(ErrCodeDeclined, "no peers for %q", group)
 }
 
-// relayErrMsg checks whether a relayer should mark this RPC as failed, and if
-// so, what metrics key to use when reporting the failure.
-func relayErrMsg(f *Frame) string {
-	switch f.messageType() {
-	case messageTypeError:
-		return newLazyError(f).Code().MetricsKey()
-	case messageTypeCallRes:
-		if ok := newLazyCallRes(f).OK(); !ok {
-			return "application-error"
-		}
-		return ""
-	default:
-		return ""
-	}
-}
+func determinesCallSuccess(f *Frame) (succeeded bool, failed bool, failMsg string) {
+	ftype := f.messageType()
 
-// indicatesSuccess checks whether receipt of a frame, on the originating
-// connection, indicates that the RPC succeeded.
-func indicatesSuccess(f *Frame) bool {
-	switch f.messageType() {
-	case messageTypeCallRes:
-		return newLazyCallRes(f).OK()
-	default:
-		return false
+	if ftype == messageTypeError {
+		// The call failed unexpectedly.
+		msg := newLazyError(f).Code().MetricsKey()
+		return false, true, msg
 	}
+
+	if ftype != messageTypeCallRes {
+		// The call hasn't succeeded or failed yet.
+		return false, false, ""
+	}
+
+	// The frame is a callRes, which could be either a success or an
+	// application error.
+	if ok := newLazyCallRes(f).OK(); !ok {
+		return false, true, "application-error"
+	}
+
+	return true, false, ""
 }
--- a/relay.go
+++ b/relay.go
@@ -210,11 +210,12 @@ func (r *Relayer) Receive(f *Frame, fType frameType) {
 
 	// call res frames don't include the OK bit, so we can't wait until the last
 	// frame of a relayed RPC to determine if the call succeeded.
-	isOriginator := fType == responseFrame
-	if isOriginator {
-		if succeeded, failed, failMsg := determinesCallSuccess(f); succeeded {
+	if fType == responseFrame {
+		// If we've gotten a response frame, we're the originating relayer and
+		// should handle stats.
+		if succeeded, failMsg := determinesCallSuccess(f); succeeded {
 			item.stats.Succeeded()
-		} else if failed {
+		} else if len(failMsg) > 0 {
 			item.stats.Failed(failMsg)
 		}
 	}
@@ -421,25 +422,17 @@ func errUnknownGroup(group string) error {
 	return NewSystemError(ErrCodeDeclined, "no peers for %q", group)
 }
 
-func determinesCallSuccess(f *Frame) (succeeded bool, failed bool, failMsg string) {
-	ftype := f.messageType()
-
-	if ftype == messageTypeError {
-		// The call failed unexpectedly.
+func determinesCallSuccess(f *Frame) (succeeded bool, failMsg string) {
+	switch f.messageType() {
+	case messageTypeError:
 		msg := newLazyError(f).Code().MetricsKey()
-		return false, true, msg
+		return false, msg
+	case messageTypeCallRes:
+		if newLazyCallRes(f).OK() {
+			return true, ""
+		}
+		return false, "application-error"
+	default:
+		return false, ""
 	}
-
-	if ftype != messageTypeCallRes {
-		// The call hasn't succeeded or failed yet.
-		return false, false, ""
-	}
-
-	// The frame is a callRes, which could be either a success or an
-	// application error.
-	if ok := newLazyCallRes(f).OK(); !ok {
-		return false, true, "application-error"
-	}
-
-	return true, false, ""
 }
--- a/relay.go
+++ b/relay.go
@@ -41,6 +41,7 @@ const _relayTombTTL = time.Second
 type relayItem struct {
 	*time.Timer
 
+	stats       relay.CallStats
 	remapID     uint32
 	destination *Relayer
 	tomb        bool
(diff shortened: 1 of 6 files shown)


relay/relay.go
```diff
@@ -35,3 +37,18 @@ type Hosts interface {
...
 	Get(CallFrame) string
 }
+
+// CallStats is a reporter for per-request stats.
+type CallStats interface {
+	// The call succeeded (possibly after retrying).
+	Succeeded()
+	// The RPC failed.
+	Failed(reason string)
+	// Stop all call-related timers. Must be called exactly once per relayed call.
```
prashantv: Call this `Done` instead?
Or `Begin` / `End` as discussed offline

relay.go
```diff
@@ -229,9 +235,10 @@ func (r *Relayer) canHandleNewCall() bool {
 	return canHandle
 }
 
-func (r *Relayer) getDestination(f lazyCallReq) (*Connection, bool, error) {
+func (r *Relayer) getDestination(f lazyCallReq, cs relay.CallStats) (*Connection, bool, error) {
 	if _, ok := r.outbound.Get(f.Header.ID); ok {
 		r.logger.WithFields(LogField{"id", f.Header.ID}).Warn("received duplicate callReq")
+		cs.Failed(ErrCodeProtocol.MetricsKey())
```
prashantv: here and below, any errors that come out of the relay code should be prefixed with `relay-`

relay.go
```diff
@@ -307,8 +321,7 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return errors.New("non-callReq for inactive ID")
 	}
 	if item.tomb {
-		// Call timed out, ignore this frame.
-		// TODO: Add metrics for this case.
+		// Call timed out, ignore this frame. (We've already handled stats.)
```
prashantv: TODO: track races between received frames and timeout

akshayjshah: :+1: done.

akshayjshah:
Pulled out the unrelated bits into separate PRs and updated this with a coherent series of commits. Happy to update with any feedback, but this should be close to ready.

I'm going to investigate more, but from a quick look it seems like we need to lock the `MockCallStats` structs - otherwise the `End` and `AssertEnded` calls race in the tests.

relay.go
```diff
@@ -193,18 +196,21 @@ func (r *Relayer) Relay(f *Frame) error {
...
 
-		if _, ok := items.Get(f.Header.ID); !ok {
-			r.logger.WithFields(
-				LogField{"ID", f.Header.ID},
-			).Warn("Received a frame without a RelayItem.")
+	item, ok := items.Get(f.Header.ID)
+	if !ok {
+		r.logger.WithFields(
+			LogField{"ID", f.Header.ID},
+		).Warn("Received a frame without a RelayItem.")
```
prashantv: should we return if we don't know the ID?

relay.go
```diff
@@ -193,18 +196,21 @@ func (r *Relayer) Relay(f *Frame) error {
...
+	} else if isOriginator {
+		if msg := relayErrMsg(f); msg != "" {
```
prashantv: technically we only need `item` in the `isOriginator` path I think?

relay.go
```diff
@@ -394,3 +415,30 @@ func frameTypeFor(f *Frame) frameType {
...
+// so, what metrics key to use when reporting the failure.
+func relayErrMsg(f *Frame) string {
+	switch f.messageType() {
+	case messageTypeError:
+		return newLazyError(f).Code().MetricsKey()
+	case messageTypeCallRes:
+		if ok := newLazyCallRes(f).OK(); !ok {
+			return "application-error"
+		}
+		return ""
```
prashantv: optionally, can skip default and move return to the end of the function:
```go
switch f.messageType() {
case messageTypeError:
  return newLazyError(f).Code().MetricsKey()
case messageTypeCallRes:
  if ok := newLazyCallRes(f).OK(); !ok {
    return "application-error"
  }
}
return ""
```

relay.go
```diff
@@ -394,3 +415,30 @@ func frameTypeFor(f *Frame) frameType {
...
+	default:
+		return ""
+	}
+}
+
+// indicatesSuccess checks whether receipt of a frame, on the originating
+// connection, indicates that the RPC succeeded.
+func indicatesSuccess(f *Frame) bool {
```
prashantv: can reduce nesting:
```go
if f.messageType() != messageTypeCallRes {
  return false
}
return newLazyCallRes(f).OK()
```

relay.go
```diff
@@ -193,18 +196,21 @@ func (r *Relayer) Relay(f *Frame) error {
...
+	item, ok := items.Get(f.Header.ID)
+	if !ok {
+		r.logger.WithFields(
+			LogField{"ID", f.Header.ID},
+		).Warn("Received a frame without a RelayItem.")
+	} else if isOriginator {
+		if msg := relayErrMsg(f); msg != "" {
+			item.stats.Failed(msg)
+		} else if indicatesSuccess(f) {
+			item.stats.Succeeded()
```
prashantv: I wonder if `relayErrMsg` and `indicatesSuccess` should actually be the same function since they both create a `lazyCallResFrame` and then check `OK`.

relay/relay.go
```diff
@@ -37,3 +37,18 @@ type Hosts interface {
...
 	Get(CallFrame) string
 }
+
+// CallStats is a reporter for per-request stats.
+type CallStats interface {
+	// The call succeeded (possibly after retrying).
+	Succeeded()
+	// The RPC failed.
+	Failed(reason string)
+	// End stats collection for this RPC. Must be called exactly once.
```
prashantv: Since this is the interface that other packages provide, documentation should be written for the implementor, e.g., instead of `Must be called exactly once.`, it should be `Will be called exactly once.`

I'd also put more information here about why `End` exists (maybe specify `Success` and `Failed` are called on the call res, but `End` is called when the whole call completes, so `End` should be used for latency tracking

relay_test.go
```diff
@@ -25,18 +25,18 @@ func serviceNameOpts(s string) *testutils.ChannelOpts {
 	return testutils.NewOpts().SetServiceName(s)
 }
 
-func withRelayedEcho(t testing.TB, f func(relay, server, client *Channel)) {
+func withRelayedEcho(t testing.TB, f func(relay, server, client *Channel, ts *testutils.TestServer)) {
```
prashantv: nit: can you break out a separate PR where this is changed to instead be `f func(ts *testutils.TestServer, client *Channel)` and the below tests just create locals for the `relay` and `server`?

relay_test.go
```diff
@@ -54,11 +54,18 @@ func TestRelay(t *testing.T) {
 			assert.Equal(t, tt.header, string(arg2), "Header was mangled during relay.")
 			assert.Equal(t, tt.body, string(arg3), "Body was mangled during relay.")
 		}
+		stats := ts.RelayStats()
+		stats.AssertEdges(1)
+		callStats := stats.Get("client", "test", "echo")
+		require.Equal(t, 2, len(callStats), "Expected call stats for each RPC.")
+		for _, cs := range callStats {
+			cs.AssertCalled(1)
+		}
```
prashantv: Can we make this a little lighter-weight yet as thorough, something like:
```go
expected := mockstats.Expected()
expected.Get("client", "test", echo").Success()
assert.Equal(t, expected, ts.RelayStats(), "Unexpected stats")
```

If there are multiple calls, with the same call succeeding and failing, something like:
```go
expected := mockstats.Expected()
echoCall := expected.Call("client", "test", echo")
echoCall.Success().Fail("bad-request").Success()
expected.Call("client", "test", "fail").Fail("error")
assert.Equal(t, expected, ts.RelayStats(), "Unexpected stats")
```

prashantv:
We shouldn't need a lock in `MockCallStats` -- if there's a race between `End` and `AssertEnded`, then a lock will stop the race detector, but it means we have a real race in the test, and so the test may become flaky.

relay.go
akshayjshah: Yes, probably. The existing code just logged a warning and forwarded the frame, and I didn't want to change any more than necessary.

Happy to make that change, though.

relay.go
akshayjshah: True. Will change.

relay.go
akshayjshah: Yep, forgot that we talked about that in person.

relay_test.go
akshayjshah: Yes - I'll make that change in a follow-on PR.

relay_test.go
akshayjshah: :+1:

relay.go
akshayjshah: Changeed.

akshayjshah:
Re: locking `MockCallStats` - yeah, you're totally right, I just didn't have time to dig into it Friday afternoon. Fixed now.

abhinav:
LGTM

relay.go
prashantv: Yeah, we have to change this since we now might access `item`, and if we don't find it, it would cause a panic.

relay.go
```diff
@@ -193,18 +196,26 @@ func (r *Relayer) Relay(f *Frame) error {
...
+	}
 
-		if _, ok := items.Get(f.Header.ID); !ok {
-			r.logger.WithFields(
-				LogField{"ID", f.Header.ID},
-			).Warn("Received a frame without a RelayItem.")
+	// call res frames don't include the OK bit, so we can't wait until the last
+	// frame of a relayed RPC to determine if the call succeeded.
+	isOriginator := fType == responseFrame
+	if isOriginator {
```
prashantv: nit: can inline `isOriginator`

relay.go
```diff
@@ -394,3 +420,26 @@ func frameTypeFor(f *Frame) frameType {
 func errUnknownGroup(group string) error {
 	return NewSystemError(ErrCodeDeclined, "no peers for %q", group)
 }
+
+func determinesCallSuccess(f *Frame) (succeeded bool, failed bool, failMsg string) {
+	ftype := f.messageType()
```
prashantv: this whole method might read nicer with a switch
```go
switch ftype := f.messageType(); ftype {
case messageTypeError:
  msg := newLazyError(f).Code().MetricsKey()
  return false, msg
case messageTypeCallRes:
  if ok := newLazyCallRes(f),OK(); ok {
    return true, ""
  }
  return false, "application-error"
default:
  return false, ""
}
```

relay.go
```diff
@@ -394,3 +420,26 @@ func frameTypeFor(f *Frame) frameType {
 func errUnknownGroup(group string) error {
 	return NewSystemError(ErrCodeDeclined, "no peers for %q", group)
 }
+
+func determinesCallSuccess(f *Frame) (succeeded bool, failed bool, failMsg string) {
```
prashantv: optional: what do you think about using `len(failMsg) > 0` as `failed`? explicit is fine too, just means we have to be careful about making sure that when we receive a `failMsg`, we're also returning `failed = true`

relay/fakes.go
```diff
@@ -0,0 +1,152 @@
...
+
+type noopCallStats struct{}
+
+func (n noopCallStats) Succeeded() CallStats      { return noopCallStats{} }
+func (n noopCallStats) Failed(_ string) CallStats { return noopCallStats{} }
+func (n noopCallStats) End()                      {}
+
+// MockCallStats is a testing spy for the CallStats interface.
+type MockCallStats struct {
+	sync.Mutex
```
prashantv: can delete the mutex I think?

prashantv:
some minor comments, thanks for making the test changes, the tests look much nicer now

lgtm

relay.go
akshayjshah: Sure. I went back and forth about a switch with more nesting v. successive ifs.

relay.go
akshayjshah: Fine by me.

relay/fakes.go
akshayjshah: Yep. :facepalm:

akshayjshah:
Sorry, assumed you were on vacation and hadn't gotten to this. Nits addressed, will merge when tests pass.

PR #401, opened 2016-06-10
Remove an argument from AssertRelayStats

akshayjshah:
The test server itself implements `testing.TB`, so there's no need for
the caller to pass one into `AssertRelayStats`.


breerly:
lgtm

prashantv:
lgtm

PR #405, opened 2016-06-13
lazyCallReq: Only parse transport headers once

prashantv:
I'm parsing the frame bytes explicitly instead of using `typed` to avoid extra allocations from creating strings for each header.

Benchmark comparison:
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkCallReqFrame-8     1046          214           -79.54%

benchmark                   old allocs     new allocs     delta
BenchmarkCallReqFrame-8     24             5              -79.17%

benchmark                   old bytes     new bytes     delta
BenchmarkCallReqFrame-8     120           48            -60.00%
```

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -27,7 +27,7 @@ import (
 	"time"
 )
 
-var callerNameKeyBytes = []byte(CallerName)
+var _callerNameKeyBytes = []byte(CallerName)
 
 const (
 	// Common to many frame types.
@@ -77,12 +77,12 @@ func (cr lazyCallRes) OK() bool {
 	return cr.Payload[_resCodeIndex] == _resCodeOK
 }
 
-// TODO: Use []byte instead of string to avoid allocations.
+// TODO: Use []byte instead of string for caller/method to avoid allocations.
 type lazyCallReq struct {
 	*Frame
 
-	caller []byte
-	method []byte
+	caller string
+	method string
 }
 
 // TODO: Consider pooling lazyCallReq and using pointers to the struct.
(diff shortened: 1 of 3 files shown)


relay_messages.go
```diff
@@ -21,13 +21,14 @@
...
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
 	"time"
-
-	"github.com/uber/tchannel-go/typed"
 )
 
+var callerNameKeyBytes = []byte(CallerName)
```
akshayjshah: nit: underscore-prefix globals.

relay_messages.go
```diff
@@ -113,26 +140,7 @@ func (f lazyCallReq) Service() string {
...
-	// csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2
-	checkSumType := ChecksumType(f.Payload[cur])
-	cur += 1 /* checksum */ + checkSumType.ChecksumSize()
-
-	// arg1~2
-	arg1Len := int(binary.BigEndian.Uint16(f.Payload[cur : cur+2]))
-	cur += 2
-	arg1 := f.Payload[cur : cur+arg1Len]
-	return string(arg1)
+	return string(f.method)
```
akshayjshah: Why not eat this allocation in the constructor instead of on each method call? We know that we're going to call both `Caller` and `Method` at least once for nearly every call req.

relay_messages.go
prashantv: 👍

relay_messages.go
prashantv: true, made `Caller` and `Method` use a string backed private field. I was initially against it since the performance is different for `Caller` vs `Service`, but better to have it be faster now, and make it all consistent when we move everything to `[]byte`.

While we could also add a private `string` field for `Service`, I'd like to avoid adding private fields for anything that is at a fixed offset in the payload. If we keep `lazyCallReq` very small, we can avoid pooling and just pass the value around and avoid some more complication in the code.

prashantv:
Updated branch + benchmarks in original PR description, PTAL

akshayjshah:
lgtm

PR #406, opened 2016-06-15
Non error is an invalid error

kriskowal:
Calling GetSystemErrorCode(nil) should return ErrCodeInvalid for testing the happy path.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -177,6 +177,10 @@ func GetContextError(err error) error {
 // GetSystemErrorCode returns the code to report for the given error.  If the error is a
 // SystemError, we can get the code directly.  Otherwise treat it as an unexpected error
 func GetSystemErrorCode(err error) SystemErrCode {
+	if err == nil {
+		return ErrCodeInvalid
+	}
+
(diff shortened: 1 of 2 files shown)


errors.go
```diff
@@ -153,7 +153,7 @@ func GetSystemErrorCode(err error) SystemErrCode {
 		return se.Code()
 	}
 
-	return ErrCodeUnexpected
+	return ErrCodeInvalid
```
prashantv: I think this line should remain the same, and instead you should have an if at the very top,
```go
if err == nil {
  return ErrCodeInvalid
}
```

prashantv:
Tests are failing because non-nil errors which are not a `tchannel.SystemError` currently return `ErrCodeUnexpected`

kriskowal:
@prashantv Indeed on both counts. Pushed my amendments.

prashantv:
Can we add a simple unit test for this as well?

prashantv:
Perfect, thanks for adding unit tests for the other errors too.

lgtm

akshayjshah:
:+1:

PR #408, opened 2016-06-16
Fix trace span encoding fields in the wrong order

### LINKED ISSUE #407: Ordering of trace span fields is incorrect
The spec says:
```
spanid:8 parentid:8 traceid:8 traceflags:1
```
http://tchannel.readthedocs.io/en/latest/protocol/#tracing

However, we write/read:
```go
w.WriteUint64(s.traceID)
w.WriteUint64(s.parentID)
w.WriteUint64(s.spanID)
```

Great catch @yurishkuro and @thepaulm

prashantv:
Fixes #407

We might want to release a v1.0.6 with this change and have TChannel clients update. (cc @blampe)

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -51,17 +51,17 @@ func (s Span) String() string {
 }
 
 func (s *Span) read(r *typed.ReadBuffer) error {
-	s.traceID = r.ReadUint64()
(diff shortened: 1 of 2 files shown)


abhinav:
lol :shipit:

PR #410, opened 2016-06-17
Fix import ordering

prashantv:
Imports are ordered:
- standard imports
- optional dot import for package being tested
- internal tchannel-go imports
- external imports


abhinav:
👍

PR #411, opened 2016-06-17
Add call validation, and fix flaky call validation tests

prashantv:
Previously, we only validated that a timeout was set, not that a
servicename was specified when making an outbound call.

Add validation, and do it before we get a connection. This avoids
creating a connection that will not be used.

Previously, we would create a connection, do validation, and close the
connection immediately. On the inbound side, this led to a new
connection that was closed almost immediately, and so sometimes the test
would cause a:
"Failed to add connection to peer" "connection is in an invalid state"

By doing validation before we create a connection we make TestNoTimeout
less flaky, while also returning errors earlier and faster.

See test failure:
https://gist.github.com/prashantv/269a2c2f1242e8e2aff9ae2b201cac96

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -314,8 +314,19 @@ func TestNoTimeout(t *testing.T) {
 
 		ctx := context.Background()
 		_, _, _, err := raw.Call(ctx, ts.Server(), ts.HostPort(), "svc", "Echo", []byte("Headers"), []byte("Body"))
-		require.NotNil(t, err)
 		assert.Equal(t, ErrTimeoutRequired, err)
+
+		ts.AssertRelayStats(relay.NewMockStats())
+	})
+}
+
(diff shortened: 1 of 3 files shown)


abhinav:
Isn't this technically a breaking change?

prashantv:
We already did the validation in different places:
 - Method name too large was validated later in `call.writeMethod` called in `conn.beginCall`
 - we were previously validating the timeout in `conn.beginCall`
 - empty service name was validated when you try to create / get a subchannel

This just moves all the validation to happen in a single place before we even get the connection. For the empty service name, the error message may change (previously it was an unexported error about not being able to find the subchannel), but I don't consider error messages as part of the API -- only publicly exported errors are part of the API.

abhinav:
👍  :shipit:

PR #414, opened 2016-06-18
Add Peer struct to represent a peer, pass it to stats

prashantv:
We may select a peer as part of some group (e.g. canary), and we want to
track the fact that this is a canary peer through stats. Allow users to
return an assignment as part of the relay.Hosts.Get method and set this
peer with assignment on the stats object.

### DIFF
--- a/relay/fakes.go
+++ b/relay/fakes.go
@@ -42,8 +42,9 @@ func (n noopStats) Begin(_ CallFrame) CallStats {
 
 type noopCallStats struct{}
 
-func (n noopCallStats) Succeeded() CallStats      { return noopCallStats{} }
-func (n noopCallStats) Failed(_ string) CallStats { return noopCallStats{} }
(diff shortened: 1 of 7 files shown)


relay/fakes.go
```diff
@@ -42,8 +42,9 @@ func (n noopStats) Begin(_ CallFrame) CallStats {
 
 type noopCallStats struct{}
 
-func (n noopCallStats) Succeeded() CallStats      { return noopCallStats{} }
-func (n noopCallStats) Failed(_ string) CallStats { return noopCallStats{} }
+func (n noopCallStats) Succeeded() CallStats      { return n }
+func (n noopCallStats) Failed(_ string) CallStats { return n }
+func (n noopCallStats) SetPeer(_ Peer) CallStats  { return n }
```
akshayjshah: :+1:

akshayjshah:
lgtm

PR #415, opened 2016-06-18
relay: Use []byte instead of string for Caller/Service/Method

prashantv:
We should avoid converting from `[]byte` to `string` since Go optimizes map lookups which do `map[string(bs)]` to avoid conversions.

This makes `BenchmarkCallReqFrame` much faster, from:
```
BenchmarkCallReqFrame-8 10000000               216 ns/op              48 B/op          5 allocs/op
```

to:
```
BenchmarkCallReqFrame-8 20000000                75.4 ns/op             0 B/op          0 allocs/op
```

### DIFF
--- a/relay_messages_test.go
+++ b/relay_messages_test.go
@@ -206,7 +206,7 @@ func TestLazyCallReqRejectsOtherFrames(t *testing.T) {
 func TestLazyCallReqService(t *testing.T) {
 	withLazyCallReqCombinations(func(crt testCallReq) {
 		cr := crt.req()
-		assert.Equal(t, "bankmoji", cr.Service(), "Service name mismatch")
+		assert.Equal(t, "bankmoji", string(cr.Service()), "Service name mismatch")
 	})
 }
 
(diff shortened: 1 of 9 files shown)


relay_messages_test.go
```diff
@@ -206,25 +206,25 @@ func TestLazyCallReqRejectsOtherFrames(t *testing.T) {
 func TestLazyCallReqService(t *testing.T) {
 	withLazyCallReqCombinations(func(crt testCallReq) {
 		cr := crt.req()
-		assert.Equal(t, "bankmoji", cr.Service(), "Service name mismatch")
+		assert.Equal(t, []byte("bankmoji"), cr.Service(), "Service name mismatch")
```
akshayjshah: Perhaps we can cast `cr.Service()` to a string instead? I don't find the testify output from failed assertions on byte slices very helpful - strings are much easier.

Same below.

akshayjshah:
Nice! Apart from tests, lgtm.

relay_messages_test.go
prashantv: Good catch, yep we should be comparing strings for nicer assertions. Fixed.

PR #417, opened 2016-06-20
relay: Allow relay.Hosts to return an error selecting peers

prashantv:
This gives the caller more control over the error, and also allows for
more error types. E.g., use Declined for peer not found, but Busy for
rate limiting errors.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -1,6 +1,7 @@
 package tchannel_test
 
 import (
+	"errors"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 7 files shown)


relay_test.go
```diff
@@ -137,29 +138,57 @@ func TestRelayIDClash(t *testing.T) {
...
+			}
+
+			assert.Equal(t, tt.wantErr, err, "%v: unexpected error", tt.desc)
+
+			calls := relay.NewMockStats()
+			calls.Add(client.PeerInfo().ServiceName, "svc", "echo").
+				ExpectNoPeer().Failed(tt.statsKey).End()
+			ts.AssertRelayStats(calls)
+		})
+	}
```
akshayjshah: impressed_okay

akshayjshah:
:+1: makes sense.

PR #419, opened 2016-06-22
Fix assertion that warmup succeeded

akshayjshah:
In the preamble to this benchmark, we need to successfully warm up each
TCP client. However, this assertion is testing an error created earlier;
fix it so that the benchmark doesn't proceed if we fail warmup.


prashantv:
lgtm

breerly:
cool :shipit:

PR #420, opened 2016-06-22
relay: Precalculate frame properties before writing to sendCh

prashantv:
Previously, we wrote a frame to sendCh, then checked properties on the
frame. After writing to sendCh, the receiver may write out the frame and
release the frame back to the pool, before we've had a chance to run.

This meant we could be deleting the wrong relay item (or not deleting
anything), which manifested itself as a timeout error frame being sent
after a call response frame.


akshayjshah:
:+1: not controversial.

PR #423, opened 2016-06-22
Add CallOptions() to IncomingCall for proxying

sectioneight:
This lets you proxy an incoming request to another node and set the
callerName to that of the original request.

### DIFF
--- a/context.go
+++ b/context.go
@@ -59,6 +59,11 @@ type IncomingCall interface {
 	// If the caller is an ephemeral peer, then the HostPort cannot be used to make new
 	// connections to the caller.
 	RemotePeer() PeerInfo
+
+	// CallOptions returns the call options set for the incoming call. It can be useful
+	// if you are forwarding a request and wish to retain the CallerName(), which is not
+	// possible to set manually.
+	CallOptions() *CallOptions
 }
 
 func getTChannelParams(ctx context.Context) *tchannelCtxParams {
--- a/inbound_test.go
+++ b/inbound_test.go
@@ -101,3 +101,44 @@ func TestInboundConnection(t *testing.T) {
 		require.NoError(t, err, "Call failed")
 	})
 }
+
+func TestInboundConnection_CallOptions(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	WithVerifiedServer(t, nil, func(serverCh *Channel, serverHostPort string) {
+		testutils.RegisterFunc(serverCh, "test", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
+			assert.Equal(t, "client", CurrentCall(ctx).CallerName(), "Expected caller name to be passed through")
+			return &raw.Res{}, nil
+		})
+
+		backendName := serverCh.PeerInfo().ServiceName
+
+		proxyCh, err := testutils.NewServerChannel(&testutils.ChannelOpts{ServiceName: "proxy"})
+		require.NoError(t, err, "Proxy channel creation failed")
+		defer proxyCh.Close()
+
+		subCh := proxyCh.GetSubChannel(backendName)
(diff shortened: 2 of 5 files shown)


context.go
```diff
@@ -59,6 +59,11 @@ type IncomingCall interface {
 	// If the caller is an ephemeral peer, then the HostPort cannot be used to make new
 	// connections to the caller.
 	RemotePeer() PeerInfo
+
+	// CallOptions returns the call options set for the incoming call. It can be useful
+	// if you are forwarding a request and wish to retain the CallerName(), which is not
+	// possible to set manually.
+	CallOptions() *CallOptions
```
sectioneight: This is a pretty generic name that may lead people to use it when they shouldn't. Should we make the signature mention something about "Preserve", "Forward", or "Proxy"?

sectioneight: Also should these be alphabetized? It's slightly out-of-order already

prashantv: They're basically in date added order, not very useful, but I'd like to preserve history for when they were added, so can leave as is.

For the name, I think it's a little generic, but it's on the `IncomingCall` interface, so I think it makes sense (reads to me like incoming call options, which is what it is).

prashantv: `ForwardCallOptions` is also fine though

testutils/call.go
```diff
@@ -64,6 +64,14 @@ func (f *FakeIncomingCall) RemotePeer() tchannel.PeerInfo {
 	return f.RemotePeerF
 }
 
+// CallOptions returns the incoming call options suitable for proxying a request.
+func (f *FakeIncomingCall) CallOptions() *tchannel.CallOptions {
+	return &tchannel.CallOptions{
+		ShardKey:        f.ShardKey(),
```
prashantv: should we also have caller name and format here?

inbound_test.go
```diff
@@ -101,3 +101,44 @@ func TestInboundConnection(t *testing.T) {
 		require.NoError(t, err, "Call failed")
 	})
 }
+
+func TestInboundConnection_CallOptions(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	WithVerifiedServer(t, nil, func(serverCh *Channel, serverHostPort string) {
```
prashantv: can you make this test use `testutils.WithTestServer`, and then instead of `testutils.NewServerChannel` we should use `ts.NewServer` and `ts.NewClient` for the client.

there's extra verification for things created using the `testutils.TestServer`

inbound_test.go
```diff
@@ -101,3 +101,44 @@ func TestInboundConnection(t *testing.T) {
...
+			require.NoError(t, NewArgReader(outbound.Response().Arg3Reader()).Read(&arg3), "Read arg3 failed")
+			require.NoError(t, NewArgWriter(inbound.Response().Arg2Writer()).Write(arg2), "Write arg2 failed")
+			require.NoError(t, NewArgWriter(inbound.Response().Arg3Writer()).Write(arg3), "Write arg3 failed")
+		}))
+
+		clientCh, err := NewChannel("client", nil)
+		require.NoError(t, err, "Create client channel failed")
+		defer clientCh.Close()
+
+		clientCh.Peers().Add(proxyCh.PeerInfo().HostPort)
```
prashantv: can delete this line since `raw.Call` specifies the hostPort explicitly

testutils/call.go
sectioneight: I can set a dummy value for format (it's not a field on fake at the moment).

I can't set callerName since it's unexported and in a different package.

inbound_test.go
sectioneight: Done

testutils/call.go
prashantv: Ah makes sense thanks

inbound_test.go
sectioneight: Done

prashantv:
This looks great, 🚢 once Travis is happy.

Thanks @sectioneight!

PR #424, opened 2016-06-23
Add a Zone field to Peer

akshayjshah:
To integrate with TC, we need more information available on the Peer
struct. Along the way, remove the word "Assignment" - it's too similar
to terms used in the rest of our architecture.


prashantv:
lgtm

PR #426, opened 2016-06-24
Integrate OpenTracing

yurishkuro:
Ready for review

Changes:
* internal span reporting completely removed and replaced with OpenTracing
* span context and baggage are propagated via application headers (json only atm, thrift next)
* channel can be configured with `opentracing.Tracer`, otherwise it defaults to `opentracing.GlobalTracer()` (which is no-op by default)
* outbound calls tracing
  * span is started in `(c *Connection) beginCall`
  * if tracer understands "zipkin" format, then the tracing fields are populated in `callReq`
  * the actual span/baggage info is injected in `json/call.go`
  * span is stored in `Response` and finished in a couple of places in `outbound.go`
* inbound requests tracing
  * if tracer understands "zipkin" format, the span is started in `inbound.go` and baggage is added in the encoding handler
  * otherwise span is started in the encoding handler only
  * finished generically in completion callbacks in `inbound.go`

TODO:
 * [x] ~~avoid duplicating trace IDs in the headers and in the frame~~
 * [x] support thrift
 * [x] support raw
 * [x] support for disabling tracing by instructing Tracer to not sample spans
 * [x] generate random IDs for native `Span` if tracer is not zipkin-compatible


yurishkuro:
@prashantv @uber/rpc

prashantv:
@yurishkuro Looking over the changes now, some high level feedback:
- I think `inbound.go` should initially create the span and set the span in the context. I think the subpackages are responsible for possibly adding baggage to the span from the headers.
- Spans seem like they are mutable, does this mean that anyone who accesses the context could modify tags/baggage?
- It seems like the `tracingDisabled` bit uses `Flags` in opentracing, but I don't see anything about Flags in the opentracing godoc, https://godoc.org/github.com/opentracing/opentracing-go#Span

I think the outbound side looks good, just a little unsure about inbound. If the TChannel library is used without any of the `json` or `thrift` packages, will the context still be created with the opentracing span and be propagated correctly? I would expect everything except tags/baggage to work in that case.

yurishkuro:
> I think `inbound.go` should initially create the span and set the span in the context. 

The problem with that is it will make the solution really tied to Zipkin-style tracing systems, or at least those who support injection carrier matching `tchannel.Span` (which Jaeger does). In the current implementation, the outgoing span is injected into app headers using standard OT API, not Zipkin-specific, but if the tracer happens to support Zipkin style, then the trace IDs are also added to the frame tracing fields, as a bonus. So the tracing will still work with non-Zipkin tracers.

On the inbound side, if we start the span before app headers are known, we'd have to start it from the frame tracing fields, i.e. only in Zipkin-specific way. Whereas the span created in `json/handler` with work with any tracer.

One dirty way would be to start a span in `inbound.go` with zipkin-specific API, but then if the app headers are available then start a new one and replace the first one. In Jaeger starting and not finishing a span is not a problem, it just gets garbage collected, may not be the case with other tracers.

> Spans seem like they are mutable, does this mean that anyone who accesses the context could modify tags/baggage?

Yes, which is in fact the intention, the user code may add other tags/events/baggage to the span.

> It seems like the tracingDisabled bit uses Flags in opentracing, but I don't see anything about Flags in the opentracing godoc, https://godoc.org/github.com/opentracing/opentracing-go#Span

Correct, there are no "flags" in OpenTracing, so the way these are exposed to `tchannel.Span` is via above-mentioned Zipkin-specific API. There's also no explicit API to _disable_ tracing, but there is a way to force the sampling on or off, which is what I used.

yurishkuro:
@prashantv I think this current form does what you want. If the tracer supports zipkin-style span context, then the span is created directly in the `inbound.go`, and once the headers are read the baggage is added to that span (currently by creating a new span and discarding it, but with the upcoming [PR #82](https://github.com/opentracing/opentracing-go/pull/82) in opentracing-go I will be able to do that without starting a span). If the tracer does not support zipkin-style, then the span is only created in `json/handler.go` for standard OpenTracing API.

I've extended the propagation test to use 3 tracers, the default Noop tracer, Jaeger tracer, and basictracer-go (which does not support zipkin style).

prashantv:
@yurishkuro Nice! Thanks for making the changes, I definitely like this approach more.

- Should we export `Span` directly on the response? Can we make it a method (there's some users of TChannel that have built their own interface to match the calls, and that's only possible because we use methods for exporting any data)
- I want to make sure that disabling of a trace works across all tracers
- If the Jaeger tracer is used, then the tracing information is propagated via TChannel's fixed fields -- is this information duplicated in headers? Or is it only in the fixed fields?
- Little worried about Span being `nil` in a few places, is it safer to use an empty span in those cases? Want to make sure that we don't have nil dereference panics, and instead return some zero value instead

I'm assuming that all opencontext headers are added with a prefix even in TChannel's headers so they won't clash right?

Should we put the code to extract/inject tracing into some internal package that can be used by Thrift + JSON.

yurishkuro:
> I want to make sure that disabling of a trace works across all tracers

We can do it in two different ways. What I've done currently is disabling the _sampling_ of the current trace (or rather span and its children). OT tracers are not obligated to respect the sampling hints. Another approach would be to completely disable all calls to OpenTracing API if disableTracing is requested.

> If the Jaeger tracer is used, then the tracing information is propagated via TChannel's fixed fields -- is this information duplicated in headers? Or is it only in the fixed fields?

It is duplicated. The headers are read/set via standard OpenTracing API. If we really want to avoid duplication, I can hack it via yet another special Format, like BaggageOnlyTextMapFormat, which will only be known to Jaeger and TChannel, and if the tracer says that format is not supported, then fallback to full span context headers

> Little worried about Span being nil in a few places, is it safer to use an empty span in those cases? Want to make sure that we don't have nil dereference panics, and instead return some zero value instead

I can use opentracing.Noop span, by default, but will need another flag to distinguish default from actual span, because `json/handler.go` branches depending on whether the span was created in `inbound.go`

> I'm assuming that all opencontext headers are added with a prefix even in TChannel's headers so they won't clash right?

Normally the headers returned by a Tracer are prefixed with tracer-specific strings. E.g. Jaeger's headers are either `uber-trace-id` or `uberctx-***`. Given that tchannel has special handing of the headers, it's possible to further prefix those, but I am not sure it's necessary.

yurishkuro:
@prashantv @breerly this is ready for review

breerly:
Exciting - will comb through today!

Makefile
```diff
@@ -116,12 +116,15 @@ else
...
 endif
 	@echo "Checking for unresolved FIXMEs"
-	-git grep -i fixme | $(FILTER) | grep -v -e Makefile | tee -a lint.log
+	-git grep -i -n fixme | $(FILTER) | grep -v -e Makefile | tee -a lint.log
 	@[ ! -s lint.log ]
 else
 	@echo "Skipping linters on" $(GO_VERSION)
 endif
 
+todos:
```
breerly: suggest moving this to diff PR

tracing_test.go
```diff
@@ -21,314 +21,515 @@
...
-		tt.serverOpts = testutils.DefaultOpts(tt.serverOpts).SetTimeNow(serverNow).NoRelay()
-		tt.clientOpts = testutils.DefaultOpts(tt.clientOpts).SetTimeNow(clientNow)
+type TracerType string
 
-		WithVerifiedServer(t, tt.serverOpts, func(ch *Channel, hostPort string) {
-			state.signal = make(chan struct{})
+const (
+	Noop   TracerType = "NOOP"
+	Basic  TracerType = "BASIC"
+	Jaeger TracerType = "JAEGER"
```
breerly: Perhaps add a comment describing what these use. Also, is this an enum? Might use an [iota](https://golang.org/ref/spec#Iota) if so.

yurishkuro: Adding comments. I chose strings because in case of test errors I get a nice log like this:
```
Messages:	Wrong span count; test {forwardCount:2 tracingDisabled:false format:raw expectedBaggage: expectedSpanCount:2} 
with tracer {tracerType:BASIC tracer:0xc820016a00 spansRecorded:0x16f870 resetSpans:0x16f8b0 isFake:false zipkinCompatible:false}
```
(note `tracerType:BASIC`), but if I use int/iota, then it prints ``tracerType:1` - ¯\_(ツ)_/¯

outbound.go
```diff
@@ -24,8 +24,11 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/uber/tchannel-go/typed"
+	"github.com/opentracing/opentracing-go"
 	"golang.org/x/net/context"
+
+	"github.com/opentracing/opentracing-go/ext"
+	"github.com/uber/tchannel-go/typed"
```
abhinav: Import groups are off.

tracing.go
```diff
@@ -22,32 +22,37 @@ package tchannel
...
 import (
 	"fmt"
-	"math/rand"
+	"net"
+	"strconv"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+	"golang.org/x/net/context"
```
abhinav: This import group should be below the tchannel-go one.

tracing.go
```diff
@@ -84,53 +87,135 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
-	if tracingEnabled {
-		flags = tracingFlagEnabled
+// ExtractInboundSpan deserializes tracing span from the incoming `headers`.
+// If the response object already contains a pre-deserialized span (only for
+// Zipkin-compatible tracers), then the baggage is extracted from the headers
+// and added to that span.
+func ExtractInboundSpan(ctx context.Context, call *InboundCall, headers map[string]string, tracer opentracing.Tracer) context.Context {
+	var span = call.Response().span
+	operationName := call.ServiceName() + "::" + call.MethodString()
+	if span != nil {
```
abhinav: NIt: Reduce nesting by inverting this:

```go
if span == nil {
  return opentracing.ContextWithSpan(ctx, span)
}

// copy baggage from headers
if headers != nil {
  ...
}

// ...
```

tracing.go
yurishkuro: is it codified somewhere in this repo?  The order I use is (1) std libs, (2) other external deps, (3) this repo own packages

abhinav: This repo uses (or at least is supposed to use): std libs, local imports, external imports. Don't know if it's documented anywhere.

CC @prashantv

yurishkuro: ok. I don't get the logic of this particular order, but since it's already everywhere I'l update my changes.

context.go
```diff
@@ -76,9 +76,7 @@ func getTChannelParams(ctx context.Context) *tchannelCtxParams {
 
 // NewContext returns a new root context used to make TChannel requests.
 func NewContext(timeout time.Duration) (context.Context, context.CancelFunc) {
-	return NewContextBuilder(timeout).
-		setSpan(NewRootSpan()).
```
prashantv: When you make a new context, is there any way to know what the trace ID will be for the call you're about to make?

glide.yaml
```diff
@@ -44,3 +44,8 @@ import:
   - quantile
 - package: github.com/uber-go/atomic
 - package: github.com/prashantv/protectmem
+- package: github.com/opentracing/opentracing-go
+- package: github.com/uber/jaeger-client-go
+- package: github.com/opentracing/basictracer-go
+- package: github.com/gogo/protobuf
```
prashantv: This shouldn't be in the glide.yaml, I don't think tchannel-go is directly using protobuf anywhere right?

inbound.go
```diff
@@ -185,9 +172,11 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
 		return
 	}
 
-	call.commonStatsTags["endpoint"] = string(call.method)
+	call.commonStatsTags["endpoint"] = call.methodString
```
prashantv: thanks for fixing this (and same below)

outbound.go
```diff
@@ -332,13 +307,22 @@ func cloneTags(tags map[string]string) map[string]string {
...
 func (response *OutboundCallResponse) doneReading(unexpected error) {
-	now := response.GetTime()
-	response.AddAnnotationAt(AnnotationKeyClientReceive, now)
-	response.Report()
+	now := response.timeNow()
 
 	isSuccess := unexpected == nil && !response.ApplicationError()
 	lastAttempt := isSuccess || !response.requestState.HasRetries(unexpected)
 
+	// TODO how should this work with retries?
```
prashantv: This is a good question, I think previously we would report multiple separate spans. We should make sure it still makes sense -- I think separate spans for each attempt would be the best approach?

tracing.go
```diff
@@ -22,32 +22,37 @@ package tchannel
...
+	"github.com/opentracing/opentracing-go/ext"
+	"golang.org/x/net/context"
 )
 
-// traceRng is a thread-safe random number generator for generating trace IDs.
-var traceRng = trand.NewSeeded()
+// ZipkinSpanFormat defines a name for OpenTracing carrier format that tracer may support.
+// It is used to extract zipkin-style trace/span IDs from the OpenTracing Span, which are
+// otherwise not exposed explicitly.
+const ZipkinSpanFormat string = "zipkin-span-format"
```
prashantv: nit: don't need `string` here

context.go
yurishkuro: First, trace ID is not a concept that's exposed in OpenTracing API, so strictly speaking - no.

In practice, if there is already an OpenTracing span in the current context, then a child span will be created when the outbound call is made (but not when the context is created). If the tracer is Zipkin-compatible, then technically it's possible to extract trace ID from the current span, and internally the outbound call will do that and store that trace ID in the frame's `Trace` fields.

tracing.go
```diff
@@ -84,53 +87,133 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+	var parent opentracing.SpanContext // ok to be nil
+	if s := opentracing.SpanFromContext(ctx); s != nil {
+		parent = s.Context()
 	}
-	if s.spanID == 0 {
-		childSpan.spanID = childSpan.traceID
-	} else {
-		childSpan.spanID = uint64(traceRng.Int63())
+	span := c.Tracer().StartSpan(
+		serviceName+"::"+methodName,
```
prashantv: the `serviceName` here is not the Thrift service name, it's the routing service name, so for example in moe, the values would be:
`serviceName = "moe"`
`methodName = "Stooges::nyuck"`

If you append them, the spans will become `moe::Stooges::nyuck` which I don't think is what we want.

tracing.go
```diff
@@ -84,53 +87,133 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+	span.SetTag("as", call.callReq.Headers[ArgScheme])
+	call.callReq.Tracing.initFromOpenTracing(span)
+	return span
 }
 
-func (s *Span) sampleRootSpan(sampleRate float64) {
-	// Sampling only affects root spans
-	if s.ParentID() != 0 {
-		return
+// InjectOutboundSpan serializes the tracing span from the `response` into the `headers`.
```
prashantv: Can you explain this a little more, and explain how it should be used

tracing.go
```diff
@@ -84,53 +87,133 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
-	flags := byte(0)
-	if tracingEnabled {
-		flags = tracingFlagEnabled
+// ExtractInboundSpan deserializes tracing span from the incoming `headers`.
+// If the response object already contains a pre-deserialized span (only for
+// Zipkin-compatible tracers), then the baggage is extracted from the headers
+// and added to that span.
+func ExtractInboundSpan(ctx context.Context, call *InboundCall, headers map[string]string, tracer opentracing.Tracer) context.Context {
+	var span = call.Response().span
+	operationName := call.ServiceName() + "::" + call.MethodString()
```
prashantv: same comment as earlier, this will lead to operation being `moe::Stooges::nyuck`

tracing.go
```diff
@@ -84,53 +87,133 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+		}
+	} else {
+		var parent opentracing.SpanContext
+		if headers != nil {
+			carrier := opentracing.TextMapCarrier(headers)
+			parent, _ = tracer.Extract(opentracing.TextMap, carrier)
+		}
+		span = tracer.StartSpan(operationName, ext.RPCServerOption(parent))
+		ext.PeerService.Set(span, call.CallerName())
+		span.SetTag("as", string(call.Format()))
```
prashantv: this looks very similar to `extractInboundSpan`

context.go
prashantv: One use case we have is `yab` which makes a call from an outbound perspective, but then outputs the traceID (so you can find the trace later). It uses the `CurrentSpan(ctx)` after creating a context, so I wonder if every new context should include a new root span. While `yab` can be updated, I don't want to break any other uses of this pattern.

yurishkuro: The main issue I had with Context doing anything related to spans is that spans can only be created by the Tracer. Passing the tracer to Context methods seemed very wrong.

I think we should revisit the use case of `yab`. What you describe sounds like a debugging feature, which is much better handled by configuring the Tracer with a logging reporter (alone or in addition to real span reporter). For example, in Jaeger this is achieved by setting `logSpans: true` in the config (or programmatically).

prashantv: While it is a debugging feature, the TChannel library should allow these use cases. E.g. `yab` should have control over how the trace ID is outputted. Passing tracer does seem pretty wrong, would setting a parent span (that the user creates) be a better solution?

There also seems internal users using `CurrentSpan(ctx)`, not sure how they're using it or whether they'll be affected.

yurishkuro: Setting parent (root) span on the context would indeed allow access to the trace ID prior to child span being created inside of tchannel.

The way `CurrentSpan(ctx)` works is it tries to find an OpenTracing Span in the context, and then tries to convert it to internal `Span` format (assuming tracer supports that). Otherwise it returns nil.

yurishkuro: NB: if returning nil from CurrentSpan is a concern, we could return a pointer to a default/empty `Span{}` with 0 IDs.

prashantv: Sorry I should have been clearer. Returning `nil` isn't the issue, it's more that there's existing internal callers to `CurrentCall` that might be depending on the trace ID being set as they're about to make an outbound call.

I'll sync up over email about what existing callers are and we can figure out whether they need it to work or not.

outbound.go
yurishkuro: That would be hard, because at this low level we don't have access to the headers. I think a better alternative (I believe Finagle does it too) is to add events to the current span, which will show on the timeline as dots (red dots specifically, if we mark them as errors).

glide.yaml
yurishkuro: a fully functional mock tracer is not available directly in opentracing-go, so I replaced basictracer with it altogether.

btw, the protobuf dependency had to be in glide.yaml as otherwise it was not getting installed or even locked in glide.lock, even though basictracer depends on it. Not an issue now.

context.go
kriskowal: In YARPC, we should guarantee that there is a span associated with a response context. We do need tracing to expose an arbitrary string for applications like what @prashantv mentions for tools and logging. Something you can use to associate and search for the trace. It does not have to conform to the Zipkin scheme.

yurishkuro: @kriskowal OpenTracing cannot impose any specific restrictions on how a "trace identity" is represented in a string, because it's an implementation detail of the trace and they vary greatly. However, it is reasonable to require that `span.String()` returns _something_ that can be used to locate the trace in the respective UI. It's just that this "something" will be implementation specific. For example, Jaeger spans return a string like this: `73754768284bbd0f:73754768284bbd0f:0:0`. Note that the string will be unique for each span, but the first segment is a common trace id (implementation detail).

yurishkuro:
@prashantv any other comments?

context.go
kriskowal: It would be satisfying for spans to be Stringers without any requirements for the form of the string. This would reinforce the need for YARPC to return a context. We need a mechanism for YARPC to communicate back the span it created on your behalf. cc @bombela

outbound.go
kriskowal: Logging the error and logging the retry seem fine to me. That would work for speculative retry as well, just not an error.

kriskowal:
lgtm. please await stamp from @prashantv.

outbound.go
yurishkuro: If no objections, I would like to leave this for another PR, this one is already huge and been outstanding for almost 3 weeks. The current logic replicates the prior logic where a span was reported on every retry, so no worse than before.

Makefile
```diff
@@ -12,7 +12,7 @@ endif
 
 PATH := $(GOPATH)/bin:$(PATH)
 EXAMPLES=./examples/bench/server ./examples/bench/client ./examples/ping ./examples/thrift ./examples/hyperbahn/echo-server
-PKGS := . ./json ./hyperbahn ./thrift ./typed ./trace $(EXAMPLES)
+PKGS := . ./json ./hyperbahn ./thrift ./typed $(EXAMPLES)
```
prashantv: asdasd

channel.go
```diff
@@ -140,13 +129,21 @@ type Channel struct {
...
+	statsReporter StatsReporter
+	tracer        opentracing.Tracer
+	subChannels   *subChannelMap
+	timeNow       func() time.Time
+}
+
+// Tracer returns the OpenTracing Tracer for this channel.
+// If no tracer was provided in the configuration, returns opentracing.GlobalTracer().
+func (ccc channelConnectionCommon) Tracer() opentracing.Tracer {
+	if ccc.tracer != nil {
```
prashantv: shouldn't we only need to do this check once when creating the channel rather than everytime `Tracer` is called?

context_internal_test.go
```diff
@@ -0,0 +1,16 @@
+package tchannel
```
prashantv: add uber license

context_test.go
```diff
@@ -208,25 +194,12 @@ func TestContextWithHeadersAsContext(t *testing.T) {
...
-	ctx4, _ := NewContextBuilder(time.Second).
-		SetParentContext(ctx2).
-		SetExternalSpan(3, 2, 1, true).
 		Build()
-	span4 := NewSpan(3, 2, 1)
-	assert.Equal(t, &span4, CurrentSpan(ctx4), "external span used")
+	assert.Equal(t, "some value", ctx2.Value("some key"), "key/value propagated from parent ctx")
 
 	goroutines.VerifyNoLeaks(t, nil)
 }
```
prashantv: Can we add a test where you create a span outside of tchannel using opentracing, pass that context in, and call `CurrentSpan` and validate it matches?

channel.go
yurishkuro: I did it as late binding, in case the global tracer is set after the channel is created

prashantv: i don't understand -- you can't change `ccc.tracer` after the channel is created right?

inbound.go
```diff
@@ -62,7 +64,8 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 
 	call := new(InboundCall)
 	call.conn = c
-	ctx, cancel := newIncomingContext(call, callReq.TimeToLive, &callReq.Tracing)
+	// TODO used to propagate tracingDisabled
```
prashantv: can you explain this comment more?

channel.go
yurishkuro: correct, but `opentracing.GlobalTracer()` _can_ change (by default it returns Noop tracer, not a real one).

inbound.go
yurishkuro: think it was a note to myself, I'll remove it. The disabling is propagated and covered by the unit test.

channel.go
prashantv: Ah, makes sense. Can you update the comment to call that out, so we don't accidentally "optimize" this and break it. (ideally a unit test too).

json/handler.go
```diff
@@ -74,18 +75,24 @@ func verifyHandler(t reflect.Type) error {
...
 	if err := verifyHandler(hV.Type()); err != nil {
 		return nil, err
 	}
 	argType := hV.Type().In(1)
-	return &handler{hV, argType, argType.Kind() == reflect.Map}, nil
+	return &handler{
+		method:   method,
+		handler:  hV,
+		argType:  argType,
+		isArgMap: argType.Kind() == reflect.Map}, nil
```
prashantv: nit: put the `}, nil` on a separate new line

json/handler.go
```diff
@@ -74,18 +75,24 @@ func verifyHandler(t reflect.Type) error {
...
-func toHandler(f interface{}) (*handler, error) {
+func toHandler(method string, f interface{}) (*handler, error) {
 	hV := reflect.ValueOf(f)
 	if err := verifyHandler(hV.Type()); err != nil {
 		return nil, err
 	}
 	argType := hV.Type().In(1)
-	return &handler{hV, argType, argType.Kind() == reflect.Map}, nil
+	return &handler{
+		method:   method,
```
prashantv: is the method used for anything?

outbound.go
prashantv: Yep, happy to defer on this for a later PR.

tracing.go
```diff
@@ -22,32 +22,37 @@ package tchannel
...
+	"github.com/opentracing/opentracing-go/ext"
+	"golang.org/x/net/context"
 )
 
-// traceRng is a thread-safe random number generator for generating trace IDs.
-var traceRng = trand.NewSeeded()
+// ZipkinSpanFormat defines a name for OpenTracing carrier format that tracer may support.
+// It is used to extract zipkin-style trace/span IDs from the OpenTracing Span, which are
+// otherwise not exposed explicitly.
+const ZipkinSpanFormat = "zipkin-span-format"
```
prashantv: does this need to be exported?

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
-func (s *Span) EnableTracing(enabled bool) {
-	if enabled {
-		s.flags |= tracingFlagEnabled
-	} else {
-		s.flags &= ^tracingFlagEnabled
+// Flags returns flags bitmap. Interpretation of the bits is up to the tracing system.
+func (s Span) Flags() byte { return s.flags }
+
+// SetTraceID sets traceID
+func (s *Span) SetTraceID(traceID uint64) { s.traceID = traceID }
```
prashantv: I'd like to avoid having these in the public interface. Can we create a custom unexported type, 
```go
type injectableSpan Span
```

And have those functions defined on that `injectableSpan`. We can freely cast between them since they're the same underlying type, but the public interface should not have those functions exposed anywhere

tracing.go
yurishkuro: it doesn't, good catch

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
-func (s Span) NewChildSpan() *Span {
-	childSpan := &Span{
-		traceID:  s.traceID,
-		parentID: s.spanID,
-		flags:    s.flags,
+// initFromOpenTracing initializes Span fields from an OpenTracing Span,
+// assuming the tracing implementation supports Zipkin-style span IDs.
+func (s *Span) initFromOpenTracing(span opentracing.Span) {
+	if err := span.Tracer().Inject(span.Context(), ZipkinSpanFormat, s); err != nil {
+		s.initRandom()
```
prashantv: since the span itself isn't stored anywhere, does this mean that multiple calls to `CurrentSpan(ctx)` could return different spans? i think returning the empty span would probably be better in that case

tracing.go
yurishkuro: hm, possibly. The Jaeger Tracer internally has the same methods on some interface to inject span IDs into a carrier played by the `Span` struct. But there's no cross-dependency on the interface name, so as long as we cast to this `injectableSpan` before passing `Span` as a carrier, it should work. Duck typing is funny.

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+	span := response.span
+	if span == nil {
+		return headers
+	}
+	if headers == nil {
+		headers = make(map[string]string)
+	}
+	carrier := opentracing.TextMapCarrier(headers)
+	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, carrier); err == nil {
+		return headers
```
prashantv: this is the same as if there was an error. should we ignore the error using `_ = [...]` and put a comment for why we're ignoring it.

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+func (c *Connection) extractInboundSpan(callReq *callReq) opentracing.Span {
+	spanCtx, err := c.Tracer().Extract(ZipkinSpanFormat, &callReq.Tracing)
+	if spanCtx != nil {
+		span := c.Tracer().StartSpan("", ext.RPCServerOption(spanCtx))
+		ext.PeerService.Set(span, callReq.Headers[CallerName])
+		span.SetTag("as", callReq.Headers[ArgScheme])
+		setPeerHostPort(span, c.remotePeerInfo.HostPort)
+		return span
+	}
+	if err != opentracing.ErrUnsupportedFormat && err != opentracing.ErrSpanContextNotFound {
```
prashantv: is it possible to get an error but a non-nil `spanCtx`? if not, prefer checking the error first and returning, and reduce nesting in the `spanCtx != nil` case

tracing.go
yurishkuro: do you think we should log it?  In practice the only time `Inject` returns an error is if the carrier type does not match the Format argument. I wasn't sure if logger would be available from the json/thrift callers.

prashantv: If it happens once, it's likely to keep happening so logging might be too much. Should we try to do this once at creation time to see if it works, and log once if it doesn't?

Or maybe it makes more sense as metrics? I'm happy to put a TODO and defer this for another PR.

prashantv:
Still have to review tests but wanted to check something: can you run the benchmarks on `dev` (`make benchmark` should work), and then on this branch, and include that in the summary. The specific benchmarks we want to look at are: BenchmarkCallsSerial, BenchmarkCallsConcurrent*.

I expect this integration will require more allocs, but I want to make sure it won't degrade perf significantly.

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
-		s.EnableTracing(false)
+// extractInboundSpan attempts to create a new OpenTracing Span for inbound request
+// using only trace IDs stored in the frame's tracing field. It only works if the
+// tracer understand Zipkin-style trace IDs. If such attempt fails, another attempt
+// will be made from the higher level function ExtractInboundSpan() once the
+// application headers are read from the wire.
+func (c *Connection) extractInboundSpan(callReq *callReq) opentracing.Span {
+	spanCtx, err := c.Tracer().Extract(ZipkinSpanFormat, &callReq.Tracing)
+	if spanCtx != nil {
+		span := c.Tracer().StartSpan("", ext.RPCServerOption(spanCtx))
```
prashantv: should this span have the method name?

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
-		flags = tracingFlagEnabled
+// ExtractInboundSpan is a higher level version of extractInboundSpan().
+// If the lower-level attempt to create a span from incoming request was
+// successful (e.g. when then Tracer supports Zipkin-style trace IDs),
+// then the application headers are only used to read the Baggage and add
+// it to the existing span. Otherwise, the standard OpenTracing API supported
+// by all tracers is used to deserialize the tracing context from the
+// application headers and start a new server-side span.
+// Once the span is started, it is wrapped in a new Context, which is returned.
+func ExtractInboundSpan(ctx context.Context, call *InboundCall, headers map[string]string, tracer opentracing.Tracer) context.Context {
```
prashantv: shuld we delete the opentracing headers from the map? Users that print all the headers will see a difference after this is merged unless we remove them.

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+	var span = call.Response().span
+	if span != nil {
+		if headers != nil {
+			// extract SpanContext from headers, but do not start another span with it,
+			// just get the baggage and copy to the already created span
+			carrier := opentracing.TextMapCarrier(headers)
+			if sc, _ := tracer.Extract(opentracing.TextMap, carrier); sc != nil {
```
prashantv: is it OK to ignore errors here?

tracing.go
```diff
@@ -84,53 +87,155 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+					span.Context().SetBaggageItem(k, v)
+					return true
+				})
+			}
+		}
+	} else {
+		var parent opentracing.SpanContext
+		if headers != nil {
+			carrier := opentracing.TextMapCarrier(headers)
+			parent, _ = tracer.Extract(opentracing.TextMap, carrier)
```
prashantv: errors being ignored here?

tracing_test.go
```diff
@@ -21,314 +21,531 @@
...
-		assert.NotEqual(t, response.SpanID, nestedResponse.SpanID)
-	})
+// JSONHandler tests tracing over JSON encoding
+type JSONHandler struct {
+	traceHandler
 }
 
-func TestTraceReportingEnabled(t *testing.T) {
-	initialTime := time.Date(2015, 2, 1, 10, 10, 0, 0, time.UTC)
+func (h *JSONHandler) callJSON(ctx json.Context, req *TracingRequest) (*TracingResponse, error) {
```
prashantv: can we move json tests into the `json` package. if you want to share code, feel free to add a `tchannel/testutils/tracetest` package with helpful utilities for both, but this package should not be testing the json package.

tracing_test.go
```diff
@@ -21,314 +21,531 @@
...
+// ThriftHandler tests tracing over Thrift encoding
+type ThriftHandler struct {
+	traceHandler
+}
 
-		span := data.Span
-		data.Span = Span{}
-		state.call = data
-		state.span = span
+func (h *ThriftHandler) Call(ctx thrift.Context, arg *gen.Data) (*gen.Data, error) {
```
prashantv: same as earlier, thrift related tests should be in the `thrift` package, not in the `tchannel` package

yurishkuro: @prashantv these handlers are currently used in a single table-driven test that runs permutations of (noop, mock, jaeger) traces vs. (raw, json, thrift) encoding. I can certainly refactor it to have a single test engine and call it from different packages, but wouldn't it completely mess up code coverage? The core tracing functions are in the tchannel package, and they will not be nearly as thoroughly exercised without that main test. On the other hand, in order to exercise them thoroughly we need to run json or thrift server, which puts the dependency on json/thrift back into the main package.  Thoughts?

tracing.go
yurishkuro: tchannel code wouldn't know which headers are tracing and which are not, because different tracing implementation use different headers

prashantv: I think this issue affects yarpc-go as well. For TChannel and YARPC, my expectation is that when I make a call with 3 headers, the handler on the other side will receive those 3 headers (ignoring middleware which the user explicitly adds that may add more headers). We use prefixes in YARPC to differentiate the user headers vs YARPC-added headers to ensure this, and I imagine most OpenTracing tracer implementations will do the same.

Should the `Extract` have some option to delete the headers that it used to extract the tracing information?

cc: @kriskowal @breerly @abhinav

tracing_test.go
prashantv: My preference is to have the tchannel tests not rely on the Thrift package. There's a couple of reasons:
- Adds another dependency to the tchannel tests, which will make running `go test .` in the core tchannel package a little slower
- Package boundaries -- there may be a proto implementation that lives outside of the package, and they should be able to do what Thrift does and have similar tests without having to touch the tchannel package

My preference is:
 - Unit test the functions exposed in the tchannel package using fake spans and calling them directly
 - Move the thrift/json integration tests into their respective packages which will ensure they're hooked up correctly
 - Feel free to add `testutils/tracetest` or some package with common code if they share a lot of logic.

tracing.go
yurishkuro: > my expectation is that when I make a call with 3 headers, the handler on the other side will receive those 3 headers (ignoring middleware which the user explicitly adds that may add more headers)

I am not sure why such behavior needs to be a requirement; most RPC frameworks provide "headers" section without any guarantee that someone else wouldn't add data there. Also, "ignoring middleware" might as well apply to tracing headers, because if users don't want tracing in the headers they can initialize the channel with a noop tracer.

The only way to support this behavior with OpenTracing API is to do inject tracing headers to an intermediate map, and then insert all data from that map into the actual headers with a special prefix, controlled by tchannel. This will introduce an additional performance hit for map allocation/copying and string concatenation. And it wouldn't be optional, because tchannel in other languages would have to use the same prefix.

prashantv: Not a strong requirement but I think a reasonable expectation. YARPC folks: what do you think?

Intermediate map + tchannel prefix doesn't seem like it's worth it. I was thinking that `Extract` (or a similar new method) could delete the entries from the map as it loads them. Or it could return the keys that it used to load entries out of the map.

Are those approaches OK?

yurishkuro: It is very difficult to argue changes to the OpenTracing API for such a fringe use case. If we really want to support such behavior, it's something very specific to TChannel, not OpenTracing. 

OpenTracing inject/extract API is designed as a middleware: `inject` adds headers to the output map, `extract` finds its own headers in the (readonly) input map. If an RPC framework wants to reserve per-middleware headers (e.g. for name-spacing), then it needs to be handled by the RPC framework itself, imo.

yurishkuro:
Benchmark results
```
43:benchmark-before.log:99: BenchmarkCallsSerial             	   10000	    202991 ns/op	   27087 B/op	     482 allocs/op
45:benchmark-before.log:104:BenchmarkCallsConcurrentServer   	   10000	    203608 ns/op	   27104 B/op	     482 allocs/op
47:benchmark-before.log:109:BenchmarkCallsConcurrentClient   	   10000	    198513 ns/op	   27076 B/op	     482 allocs/op

1:benchmark-after.log:98:   BenchmarkCallsSerial             	   10000	    208078 ns/op	   27102 B/op	     482 allocs/op
3:benchmark-after.log:103:  BenchmarkCallsConcurrentServer   	   10000	    195314 ns/op	   27093 B/op	     482 allocs/op
5:benchmark-after.log:108:  BenchmarkCallsConcurrentClient   	   10000	    193117 ns/op	   27084 B/op	     482 allocs/op

```

context_test.go
prashantv: ping, is this done somewhere else?

json/handler.go
```diff
@@ -77,15 +78,16 @@ type handler struct {
 	handler  reflect.Value
 	argType  reflect.Type
 	isArgMap bool
+	tracer   func() opentracing.Tracer
 }
 
-func toHandler(f interface{}) (*handler, error) {
+func toHandler(method string, f interface{}) (*handler, error) {
```
prashantv: why is `method` required here?

json/tracing_test.go
```diff
@@ -0,0 +1,68 @@
...
+	err := json.CallPeer(jctx, peer, h.Ch.PeerInfo().ServiceName, "call", req, &response)
+	require.NoError(h.t, err)
+	return &response, err
+}
+
+func (h *JSONHandler) callJSON(ctx json.Context, req *TracingRequest) (*TracingResponse, error) {
+	return h.HandleCall(ctx, req,
+		func(ctx context.Context, req *TracingRequest) (*TracingResponse, error) {
+			jctx := ctx.(json.Context)
+			sc := h.Ch.Peers().GetOrAdd(h.Ch.PeerInfo().HostPort)
```
prashantv: nit: call this `peer`

json/tracing_test.go
```diff
@@ -0,0 +1,68 @@
...
+			childResp := new(TracingResponse)
```
prashantv: nit: to be consistent with above, let's do `var response TracingResponse`, or change the above to also be `response := new(TracingResponse)`

json/tracing_test.go
```diff
@@ -0,0 +1,68 @@
...
+			require.NoError(h.t, json.CallPeer(jctx, sc, h.Ch.PeerInfo().ServiceName, "call", req, childResp))
+			return childResp, nil
```
prashantv: the handler returns an error, but we're using `require.NoError`. Should we just return the error from the `CallPeer` instead?

Also be careful about using `require` when it's not from the main goroutine. It uses the `Fail` semantics which are only supported on the main test goroutine:
https://golang.org/pkg/testing/#T.FailNow
```
FailNow must be called from the goroutine running the test or benchmark function
```

context_test.go
yurishkuro: `context_internal_test.go:41:func TestCurrentSpan`

json/handler.go
yurishkuro: I thought I needed it, but ended up getting from elsewhere. I'll remove it.

testutils/testtracing/propagation.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	resp := &TracingResponse{Child: childResp}
+	resp.ObserveSpan(ctx)
+
+	return resp, nil
+}
+
+// TracerType is a convenient enum to indicate which type of tracer is being used in the test
+type TracerType string
```
prashantv: is this ever printed? if not, consider just having a custom int type and using `iota`

testutils/testtracing/propagation.go
```diff
@@ -0,0 +1,313 @@
...
+func (s *PropagationTestSuite) runTestCase(
+	t *testing.T,
+	tracer tracerChoice,
+	ch *tchannel.Channel,
+	test PropagationTestCase,
+	call TracingCall,
+) {
+	descr := fmt.Sprintf("test %+v with tracer %+v", test, tracer)
+	ch.Logger().Debug("")
+	ch.Logger().Debugf("======> STARTING %s", descr)
```
prashantv: do we need these logs now that the test development is complete and it works? Maybe just a simple "Starting tracing test: %v"? I'd like to avoid the ASCII art though

testutils/testtracing/propagation.go
```diff
@@ -0,0 +1,313 @@
...
+	ch.Logger().Infof("Received tracing response %+v", response)
+
+	// Spans are finished in inbound.doneSending() or outbound.doneReading(),
+	// which are called on different go-routines and may execute *after* the
+	// response has been received by the client. Give them a chance to run.
+	for i := 0; i < 100; i++ {
+		if spanCount := tracer.spansRecorded(); spanCount == test.ExpectedSpanCount {
+			break
+		}
+		time.Sleep(time.Millisecond) // max wait: 100ms
```
prashantv: use `testutils.Timeout(time.Millisecond)` since the Travis environment can be a lot slower, so we scale timeouts there.

testutils/testtracing/propagation.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	spanCount := tracer.spansRecorded()
+	ch.Logger().Debugf("end span count: %d", spanCount)
```
prashantv: remove this log, the `assert.Equal` below will print the number if it's wrong anyway

testutils/testtracing/propagation.go
```diff
@@ -0,0 +1,313 @@
...
+	for r, cnt := response, 0; r != nil || cnt <= test.ForwardCount; r, cnt = r.Child, cnt+1 {
+		require.NotNil(t, r, "Expecting response for forward=%d; %s", cnt, descr)
+		if !tracer.isFake {
+			if tracer.zipkinCompatible || s.Encoding.HeadersSupported {
+				assert.Equal(t, root.TraceID, r.TraceID, "traceID should be the same; %s", descr)
+			}
+			assert.Equal(t, test.ExpectedBaggage, r.Luggage, "baggage should propagate; %s", descr)
+		}
+	}
+	ch.Logger().Debugf("======> FINISHED %s", descr)
```
prashantv: make logs cleaner as above

testutils/testtracing/propagation.go
```diff
@@ -0,0 +1,313 @@
...
+type TracingCall func(ctx context.Context, req *TracingRequest) (*TracingResponse, error)
+
+// EncodingInfo describes the encoding used with tracing propagation test
+type EncodingInfo struct {
+	Format           tchannel.Format
+	HeadersSupported bool
+}
+
+// PropagationTestSuite is a collection of test cases for a certain encoding
+type PropagationTestSuite struct {
```
prashantv: this test is great, nice work!

tracing.go
```diff
@@ -84,53 +88,169 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+		return headers
+	}
+	if headers == nil {
+		headers = make(map[string]string)
 	}
+	carrier := opentracing.TextMapCarrier(headers)
+	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, carrier); err != nil {
+		// Something had to go seriously wrong for Inject to fail, usually a setup problem.
+		// A good Tracer implementation may also emit a metric.
+		response.log.Error("Failed to inject tracing span: " + err.Error())
```
prashantv: don't add errors to the log message, do
```go
response.log.WithFields(ErrField(err)).Error("Failed to inject tracing span")
```

tracing.go
```diff
@@ -84,53 +88,169 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+	if isTracingDisabled(ctx) {
+		ext.SamplingPriority.Set(span, 0)
 	}
-	if s.spanID == 0 {
-		childSpan.spanID = childSpan.traceID
+	ext.SpanKindRPCClient.Set(span)
+	ext.PeerService.Set(span, serviceName)
+	setPeerHostPort(span, c.remotePeerInfo.HostPort)
+	span.SetTag("as", call.callReq.Headers[ArgScheme])
+	injectable := injectableSpan{}
```
prashantv: nit: `var injectable injectableSpan` for default values (for consistency)

tracing.go
```diff
@@ -84,53 +88,169 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+// extractInboundSpan attempts to create a new OpenTracing Span for inbound request
+// using only trace IDs stored in the frame's tracing field. It only works if the
+// tracer understand Zipkin-style trace IDs. If such attempt fails, another attempt
+// will be made from the higher level function ExtractInboundSpan() once the
+// application headers are read from the wire.
+func (c *Connection) extractInboundSpan(callReq *callReq) opentracing.Span {
+	spanCtx, err := c.Tracer().Extract(zipkinSpanFormat, &callReq.Tracing)
+	if err != nil {
+		if err != opentracing.ErrUnsupportedFormat && err != opentracing.ErrSpanContextNotFound {
+			c.log.Error("Failed to extract Zipkin-style span: " + err.Error())
```
prashantv: same as above, don't append errors to the message but use log fields

tracing.go
```diff
@@ -84,53 +88,169 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+func (s *injectableSpan) initFromOpenTracing(span opentracing.Span) error {
+	return span.Tracer().Inject(span.Context(), zipkinSpanFormat, s)
+}
+
+// CurrentSpan extracts OpenTracing Span from the Context, and if found tries to
+// extract zipkin-style trace/span IDs from it using ZipkinSpanFormat carrier.
+// If there is no OpenTracing Span in the Context, an empty span is returned.
+func CurrentSpan(ctx context.Context) *Span {
+	if sp := opentracing.SpanFromContext(ctx); sp != nil {
+		injectable := injectableSpan{}
```
prashantv: nit: `var injectable injectableSpan`

tracing.go
```diff
@@ -84,53 +88,169 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+		// return empty span on error, instead of possibly a partially filled one
 	}
+	return &emptySpan
 }
 
-// TracingEnabled checks whether tracing is enabled for this context
-func (s Span) TracingEnabled() bool { return (s.flags & tracingFlagEnabled) == tracingFlagEnabled }
+// emptySpan is returned from CurrentSpan(ctx) when there is no OpenTracing
+// Span in ctx, to avoid returning nil.
+var emptySpan Span
```
prashantv: can you move this declaration to the top of the file

tracing_test.go
```diff
@@ -21,314 +21,120 @@
...
-			client := testutils.NewClient(t, opts)
-			defer client.Close()
-
-			for i := 0; i < tt.count; i++ {
-				ctx, cancel := NewContext(time.Second)
-				defer cancel()
-
-				_, _, _, err := raw.Call(ctx, client, hostPort, ch.PeerInfo().ServiceName, "t", nil, nil)
-				require.NoError(t, err, "raw.Call failed")
+			time.Sleep(time.Millisecond) // max wait: 100ms
```
prashantv: this test looks very similar to the one in the json package, is it worth having both?

tracing.go
abhinav: Normally I would expect things like the tracing middleware to be all but
transparent. It would inject a specific set of keys into outgoing application
headers and remove those same keys from incoming application headers, the
assumption there being that those keys are owned by the tracing middleware.

I'm not strongly opposed to it, but why do you think that the tracing
middleware should infect user-owned application headers instead of being
transparent?

prashantv:
Thanks for the benchmark numbers, good to see no regressions. Can you rebase off the latest dev as well?

tracing.go
yurishkuro: >  but why do you think that the tracing middleware should infect user-owned application headers instead of being transparent

I am not against it being transparent, my point is that it's not the business of OpenTracing API, which is transport-agnostic, to deal with hiding the headers, it's the business of the RPC framework that's using OpenTracing API to instrument itself (i.e. _this_ change). In fact, OpenTracing API supports a binary format for tracing context, so that we could transmit the whole thing (including baggage) as a single app header with the name known only to tchannel (e.g. `tchannel-opentracing-context={base64blob}`), and it can be easily removed from the app headers on the receiving side. But if we want to use "text map" format of encoding tracing context (as we do in these PRs, to make them readable on the wire), OpenTracing API returns it as a map, and the only way to distinguish them on the receiving side is by prefixing each key in that map with tchannel-specific header. This will require at least one extra map allocations (keys can be cached).

tracing_test.go
yurishkuro: yes, these are nearly identical tests, but they are using different encoding, and different encodings have different paths to invoke tracing functions from their call/handler files.

yurishkuro:
rebased off dev

tracing.go
```diff
@@ -84,53 +94,167 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+			// extract SpanContext from headers, but do not start another span with it,
+			// just get the baggage and copy to the already created span
+			carrier := tracingHeadersCarrier(headers)
+			if sc, err := tracer.Extract(opentracing.TextMap, carrier); err == nil {
+				sc.ForeachBaggageItem(func(k, v string) bool {
+					span.Context().SetBaggageItem(k, v)
+					return true
+				})
+			}
+			carrier.RemoveTracingKeys()
```
yurishkuro: @prashantv I've added the namespacing feature for tracing keys by using a custom `tracingHeadersCarrier` which uses a prefix for all tracing keys. The `RemoveTracingKeys()` call restores the application headers to a pre-injection state.

The prefixed and unprefixed keys are cached.

tracing_keys.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"strings"
+	"sync"
+)
+
+// tracingKeyPrefix is used to prefix all keys used by the OpenTracing Tracer to represent
+// its trace context and baggage. The prefixing is done in order to distinguish tracing
+// headers from the actual application headers and to hide the former from the user code.
+const tracingKeyPrefix = "$tracing$"
```
yurishkuro: @prashantv ok to use this prefix?

prashantv: Yep fine with me, @kriskowal @abhinav are you OK with this as well?

abhinav: Sure. @kriskowal?

kriskowal: :ok:

channel_test.go
```diff
@@ -112,3 +113,18 @@ func TestIsolatedSubChannelsDontSharePeers(t *testing.T) {
...
+		Tracer: tracer,
+	})
+	require.NoError(t, err)
+	defer ch.Close()
+	assert.Equal(t, tracer, ch.Tracer(), "expecting tracer passed at initialization")
+
+	ch, err = NewChannel("svc", &ChannelOptions{})
+	require.NoError(t, err)
+	defer ch.Close()
+	assert.EqualValues(t, opentracing.GlobalTracer(), ch.Tracer(), "expecting default tracer")
```
prashantv: can you change the opentracing global tracer, and make sure that it's returned when you call `ch.Tracer()` (basically testing that we're calling the function every time instead of caching it).

json/tracing_test.go
```diff
@@ -0,0 +1,68 @@
...
+	suite := &PropagationTestSuite{
+		Encoding: EncodingInfo{Format: tchannel.JSON, HeadersSupported: true},
+		Register: func(t *testing.T, ch *tchannel.Channel) TracingCall {
+			handler := &JSONHandler{TraceHandler: TraceHandler{Ch: ch}, t: t}
+			json.Register(ch, json.Handlers{"call": handler.callJSON}, handler.onError)
+			return handler.firstCall
+		},
+		TestCases: map[TracerType][]PropagationTestCase{
+			Noop: {
+				{2, true, "", 0},
```
prashantv: these args aren't very clear without keys, can we add the key names to the literals

tracing.go
prashantv: ping, this is just to be consistent with line 125

tracing.go
```diff
@@ -84,53 +94,167 @@ func (s Span) ParentID() uint64 { return s.parentID }
...
+		return headers
+	}
+	if headers == nil {
+		headers = make(map[string]string)
 	}
+	carrier := tracingHeadersCarrier(headers)
+	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, carrier); err != nil {
+		// Something had to go seriously wrong for Inject to fail, usually a setup problem.
+		// A good Tracer implementation may also emit a metric.
+		response.log.Error("Failed to inject tracing span: " + err.Error())
```
prashantv: can you use structured fields for errors, `log.WithFields(...).Error("Failed to inject tracing span.")` (the log message is constant and a full sentence ending with a period)

tracing.go
prashantv: ping

prashantv:
Code lgtm, there's a few minor comments left to address -- can you address those and then we can merge.

channel_test.go
yurishkuro: I see what you mean, but my concern with that is that there is actually a global variable behind GlobalTracer(), and changing a global variable in a unit test feels iffy. Probably no harm, but if any tests happen to run concurrently, it _could_ be a problem. Wdyt?

channel_test.go
```diff
@@ -115,16 +115,25 @@ func TestIsolatedSubChannelsDontSharePeers(t *testing.T) {
...
 	require.NoError(t, err)
 	defer ch.Close()
 	assert.EqualValues(t, opentracing.GlobalTracer(), ch.Tracer(), "expecting default tracer")
+
+	// because ch.Tracer() function is doing dynamic lookup, we can change global tracer
+	origTracer := opentracing.GlobalTracer()
+	defer func() {
+		// restore it on exit
+		opentracing.InitGlobalTracer(origTracer)
+	}()
```
prashantv: you can simplify this to:
```go
defer opentracing.InitGlobalTracer(origTracer)
```

channel_test.go
prashantv: Definitely agree it's a little iffy, but we don't run our test in parallel (none of them use the `t.Parallel`). If we do decide to use it, we'll have to be careful not to enable it on this test (or other tests which rely on global state). I think it's better to have the test and validation for now, thanks for adding the test.

PR #427, opened 2016-06-29
relay: Send ephemeral host:port on outbound relay connections

prashantv:
Currently, we send the listening host:port, which causes some services
to assume that we're Hyperbahn, and send traffic intended for Hyperbahn
over the connection we opened.

In general, destination services do not need to know our host:port so
hide it in all outbound connections from the relay.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -349,14 +349,18 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 	return p.Connect(ctx)
 }
 
-// getConnectionTimeout gets a connection, and uses the given timeout if a new
+// getConnectionRelay gets a connection, and uses the given timeout if a new
 // connection is required.
-func (p *Peer) getConnectionTimeout(timeout time.Duration) (*Connection, error) {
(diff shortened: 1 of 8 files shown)


akshayjshah:
lgtm

peer.go
```diff
@@ -349,14 +349,20 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
...
 		return conn, nil
 	}
 
-	ctx, cancel := NewContext(timeout)
+	// When the relay creates outbound connections, we don't want those services
+	// to ever connect back to us and send us traffic. We hide the host:port
+	// so that service instances on remote machines don't try to connect back
+	// and don't try to send Hyperbahn traffic on this connection.
+	cb := NewContextBuilder(timeout)
+	cb.hideListeningOnOutbound = true
```
akshayjshah: While I'm totally on-board with hiding this from users, we're literally going to be living with it for years (until the last un-upgraded tchannel-python service dies). Consider adding an unexported builder method for this.

PR #429, opened 2016-07-01
relay: Always set Zone/Pool on stats, even on errors

prashantv:
It's possible that a pool was selected, but there were no peers
in that pool, in which case the errors should be attributed to
the specified pool.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -259,6 +259,7 @@ func (r *Relayer) getDestination(f lazyCallReq, cs relay.CallStats) (*Connection
 
 	// Get the destination
 	selectedPeer, err := r.hosts.Get(f)
+	cs.SetPeer(selectedPeer)
 	if err == nil && selectedPeer.HostPort == "" {
 		err = errInvalidPeerForGroup(f.Service())
 	}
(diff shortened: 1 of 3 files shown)


relay.go
```diff
@@ -259,6 +259,7 @@ func (r *Relayer) getDestination(f lazyCallReq, cs relay.CallStats) (*Connection
 
 	// Get the destination
 	selectedPeer, err := r.hosts.Get(f)
+	cs.SetPeer(selectedPeer)
```
abhinav: isn't it likely that Peer will have a zero value if `err != nil`?

prashantv: Yep but it might have the pool set so we need to cash this even if there's an error.

If the routing table picks a pool, but then can't find a task within that pool, it will return an error but the peer will contain a zone/pool. We want stats tied to the pool, so we have to set them even if there's an error.

This interface needs work (maybe multiple returns), but we need to figure out the interface for peer management first.

abhinav: 👍

abhinav:
:shipit:

PR #430, opened 2016-07-06
Sync vendor back to GOPATH in CI

yurishkuro:
Previous `glide --debug install --cache --cache-gopath` does not work, it leaves HEAD in GOPATH, and only pins versions in `vendor/`, which is subsequently removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ get_thrift:
 install:
 	GOPATH=$(OLD_GOPATH) glide --debug install --cache --cache-gopath
 	rm -rf $(VENDOR_PATH)
-	mkdir -p $(VENDOR_PATH)/src
+	mkdir -p $(VENDOR_PATH)
 	mv vendor $(VENDOR_PATH)/src
 
 install_lint:
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,9 @@ get_thrift:
 # Note that glide itself is still executed against the original GOPATH.
 install:
 	GOPATH=$(OLD_GOPATH) glide --debug install --cache --cache-gopath
-	[ -d $(VENDOR_PATH) ] && rm -rf $(VENDOR_PATH) || true
+	rm -rf $(VENDOR_PATH)
 	mkdir -p $(VENDOR_PATH)/src
-	mv vendor/* $(VENDOR_PATH)/src/
-	rm -rf vendor
+	mv vendor $(VENDOR_PATH)/src
 
 install_lint:
 ifdef SHOULD_LINT
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,11 @@ PLATFORM := $(shell uname -s | tr '[:upper:]' '[:lower:]')
 ARCH := $(shell uname -m)
 THRIFT_REL := ./scripts/travis/thrift-release/$(PLATFORM)-$(ARCH)
 
+OLD_GOPATH := $(GOPATH)
+VENDOR_PATH := $(PWD)/.tmp/vendor
+
 export PATH := $(realpath $(THRIFT_REL)):$(PATH)
+export GOPATH := $(VENDOR_PATH):$(GOPATH)
 
 # Cross language test args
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -57,7 +58,19 @@ endif
 install_glide:
 	go get -u github.com/Masterminds/glide
 
-install_ci: install_glide get_thrift install
+install_deps_ci:
+	glide --debug install --cache --cache-gopath
+	rsync -acv vendor/ $GOPATH/src
+	rm -rf vendor
+ifdef SHOULD_LINT
```
prashantv: rather than duplicate this, let's make a new target `install_lint` that both `install` and `install_deps_ci` can depend on

Makefile
```diff
@@ -57,7 +58,19 @@ endif
 install_glide:
 	go get -u github.com/Masterminds/glide
 
-install_ci: install_glide get_thrift install
+install_deps_ci:
```
prashantv: add new rule to `.PHONY`

Makefile
```diff
@@ -57,7 +58,19 @@ endif
...
+	glide --debug install --cache --cache-gopath
+	rsync -acv vendor/ $GOPATH/src
```
prashantv: we probably also want `--delete` in the `rsync`

Makefile
```diff
@@ -57,7 +58,19 @@ endif
 install_glide:
 	go get -u github.com/Masterminds/glide
 
-install_ci: install_glide get_thrift install
+install_deps_ci:
+	glide --debug install --cache --cache-gopath
```
prashantv: add a comment above this target for what we do + why

yurishkuro:
@prashantv it worked with GOPATH

Makefile
```diff
@@ -44,21 +48,31 @@ setup:
...
 
+# We want to remove `vendor` dir because thrift-gen tests don't work with it.
+# However, glide install even with --cache-gopath option leaves GOPATH at HEAD,
+# not at the desired versions from glide.lock, which are only applied to `vendor`
+# dir. So we move `vendor` to a temp dir and prepend it to GOPATH.
+# Note that glide itself is still executed against the original GOPATH.
 install:
-	glide --debug install --cache --cache-gopath
+	GOPATH=$(OLD_GOPATH) glide --debug install --cache --cache-gopath
+	[ -d $(VENDOR_PATH) ] && rm -rf $(VENDOR_PATH) || true
```
prashantv: `rm -rf` should return a successful exit code even if the directory doesn't exist. can you simplify this to just `rm -rf $(VENDOR_PATH)`

Makefile
```diff
@@ -44,21 +48,31 @@ setup:
...
+	mkdir -p $(VENDOR_PATH)/src
+	mv vendor/* $(VENDOR_PATH)/src/
```
prashantv: instead of the `mv` followed by `rm -rf`, could this be:
```bash
mkdir -p $(VENDOR_PATH)
mv vendor $(VENDOR_PATH)/src
```

prashantv:
lgtm with minor nits, thanks for fixing this @yurishkuro

Makefile
```diff
@@ -44,21 +48,30 @@ setup:
...
+# However, glide install even with --cache-gopath option leaves GOPATH at HEAD,
+# not at the desired versions from glide.lock, which are only applied to `vendor`
+# dir. So we move `vendor` to a temp dir and prepend it to GOPATH.
+# Note that glide itself is still executed against the original GOPATH.
 install:
-	glide --debug install --cache --cache-gopath
-	rm -rf vendor
+	GOPATH=$(OLD_GOPATH) glide --debug install --cache --cache-gopath
+	rm -rf $(VENDOR_PATH)
+	mkdir -p $(VENDOR_PATH)/src
```
prashantv: you probably shouldn't create `src` -- otherwise the `mv` below might move to `$(VENDOR_PATH)/src/vendor`. easier to just `mkdir -p $(VENDOR_PATH)`

PR #435, opened 2016-07-13
Make the relay.CallStats interface non-fluent

akshayjshah:
For tests, it's convenient for `relay.CallStats` to have a fluent API.
However, this results in needless allocations along the forwarding path
in production.

Instead, make the call stats interface non-fluent, and wrap it in a
fluent API during tests.

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -19,6 +19,9 @@
 // THE SOFTWARE.
 
 // Package relay contains relaying interfaces for external use.
+//
+// These interfaces are currently unstable, and aren't covered by the API
+// backwards-compatibility guarantee.
 package relay
 
 // CallFrame is an interface that abstracts access to the call req frame.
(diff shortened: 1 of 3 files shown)


connection_test.go
```diff
@@ -284,7 +284,7 @@ func TestBadRequest(t *testing.T) {
 		assert.Equal(t, ErrCodeBadRequest, GetSystemErrorCode(err))
 
 		calls := relay.NewMockStats()
-		calls.Add(ts.ServiceName(), ts.ServiceName(), "Noone").Failed("bad-request").End()
+		calls.Add(ts.ServiceName(), ts.ServiceName(), "Noone").Fluent().Failed("bad-request").End()
```
prashantv: should `calls.Add` return the fluent interface by default? or a separate method to return the `Fluent` version?

relay/relay.go
```diff
@@ -19,6 +19,9 @@
 // THE SOFTWARE.
 
 // Package relay contains relaying interfaces for external use.
+//
+// These interfaces are currently unstable, and aren't covered by the API
```
prashantv: thanks for adding this. it's unrelated to this change, but if we don't already have this on Relay* fields in the channel options, can we also add it there.

connection_test.go
akshayjshah: `calls.Add` can't return the fluent version, because `relay.MockStats` needs to satisfy the `relay.CallStats` interface.

I don't mind adding a separate method to return the fluent version, but it seems less flexible without much benefit - is `AddFluent(...)` any better than `Add(...).Fluent()`?

relay/relay.go
akshayjshah: will do.

connection_test.go
prashantv: The `Add` method is not part of any relay interface, we only have a `Begin` which takes a call frame. Am I missing something?

akshayjshah: Nope. I'm just sleepy and hallucinating :/

prashantv:
lgtm

PR #437, opened 2016-07-14
Change go1.7 to go1.7rc1 in .travis.yml

### LINKED ISSUE #436: Go 1.7 will cause build to fail
Go 1.7 is not available in `gimme`, which has had the effect of defaulting those builds to `1.5.1`.

Looks like Travis just fixed https://github.com/travis-ci/gimme/issues/53 via https://github.com/travis-ci/travis-ci/issues/6299, which means this build will begin failing since `1.7` is not available in `gimme`.

The fix might be switching to `1.7rc1` in the interim, which is available in `gimme`.

prashantv:
Fixes #436

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ go:
   - 1.4
   - 1.5
   - 1.6
-  - 1.7
+  - 1.7rc1
   - tip
 install: make install_ci
 env:


breerly:
lgtm

This is what we're using in yarpc-go - https://github.com/yarpc/yarpc-go/blob/master/.travis.yml#L8

akshayjshah:
lgtm

breerly:
Looking  good:

```
4.51s$ eval "$(gimme 1.7rc1)"
go version go1.7rc1 linux/amd64
```

PR #315, opened 2016-04-04
Release v1.0.5

prashantv:
Merges #258, #259, #264, #269, #270, #271, #272, #268, #277, #278, #275, #279, #281, #283, #284, #285, #286, #290, #287, #274, #291, #276, #292, #295, #294, #297, #293, #296, #298, #299, #302, #303, #304, #305, #306, #307, #308, #309, #310, #311, #313 and #314 from dev into master for a 1.0.5 release.

PR #442, opened 2016-07-15
Release v1.0.7

prashantv:
This release includes the relay branch, which is not officially supported. The relay branch is a separate code path which is not activated unless `RelayHosts` (unsupported API) is set.

This release includes the following changes:
- Add CallOptions() to IncomingCall which can be used as the call option when making outbound calls to proxy all transport headers.
- Add tracing information to all error frames generated by the library.
- Add GetHandlers for getting all registered methods on a subchannel.
- Fix peer score not being calculated when adding a new outbound connections
- Expose the peer information for outbound calls.
- Support a separate connection timeout from the context timeout, useful for streaming calls where the stream timeout may be much longer than the connection timeout.


breerly:
Looks like this release ships a bunch of new functionality? 

Suggest renaming to the minor `1.1.0`.

Source:  http://semver.org/#spec-item-7 -

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

prashantv:
I think we had the same discussion last time (https://github.com/uber/tchannel-go/pull/315).

We're not following strict semver here. We want everyone using `1.0.6` to update to this version since it is completely backwards compatible, and includes internal features (exposing connection information for example), but is mostly bug fixes.

We'll be cutting a new minor `1.1.0` once we integrate opentracing, as that is significant new functionality in the library. Strict semver would require us to release a `2.x` release for that change -- I don't think we want to do that though.

prashantv:
Till we figure out a proper versioning story internally, this version number doesn't mean much other than "it's safe to upgrade" for internal customers. We need to figure out a story for package versioning internally before we figure out a solution for how to version this.

PR #443, opened 2016-07-16
relay: Move fakes into relaytest package

prashantv:
The production tchannel-go package depends on the relay package, which
has a dependency on the testing package. We do not want production
binaries to depend on testing (this also brings in testing flags into
prod binaries), so move all fakes into a separate relaytest package.

This is a pure refactoring change with no logic changes.

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -32,7 +32,7 @@ import (
 
 	. "github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/raw"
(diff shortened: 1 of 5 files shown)


prashantv:
We should cut a new release once this is merged.

abhinav:
lgtm

blampe:
You did it yaaaay!

blampe:
🚢 🚢 🚢

PR #446, opened 2016-07-16
Add check_no_test_deps to avoid test dependencies from production code

prashantv:
This will Ensure we don't accidentally bring in a test dependency into
the tchannel-go package (related to #443).


blampe:
Great idea :shipit:

PR #448, opened 2016-07-19
Emit system error as per metrics document

liweiqing:
Emit system error as per metrics document.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -197,3 +197,17 @@ func GetSystemErrorMessage(err error) string {
 
 	return err.Error()
 }
+
+// GetSystemErrorKey returns the metric key for the given error.  If the error is a
+// SystemError, we can get the metric key. Otherwise, treat it as an unexpected error
+func GetSystemErrorMetricKey(err error) string {
+	if err == nil {
+		return ErrCodeInvalid.MetricsKey()
+	}
+
+	if se, ok := err.(SystemError); ok {
+		return se.Code().MetricsKey()
+	}
+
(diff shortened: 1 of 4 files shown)


errors.go
```diff
@@ -197,3 +197,17 @@ func GetSystemErrorMessage(err error) string {
 
 	return err.Error()
 }
+
+// GetSystemErrorKey returns the metric key for the given error.  If the error is a
+// SystemError, we can get the metric key. Otherwise, treat it as an unexpected error
+func GetSystemErrorMetricKey(err error) string {
```
prashantv: can you unexport this, and add unit tests for this method

inbound.go
```diff
@@ -336,6 +337,7 @@ func (response *InboundCallResponse) SendSystemError(err error) error {
 	// Fail all future attempts to read fragments
 	response.state = reqResWriterComplete
 	response.systemError = true
+	response.errorType = GetSystemErrorMetricKey(err)
```
prashantv: this is not the only way a response may fail. This is only if the caller specifically wants to return an error.

however, there's errors that happen on the connection (e.g. timeout, cancellation, network failures) which would not go through this path.

We should unit test to make sure that the above scenarios work (timeout and network failure are the ones we tend to care about).

prashantv:
Thanks for the change @liweiqing 

This change will need a lot more unit tests to make sure it works in all the cases. Right now, the inbound path is not testing the metrics at all (only the outbound path is being tested).

prashantv:
Closing out stale change.

PR #449, opened 2016-07-19
Enable disabled test TestRelayHandlesClosedPeers

prashantv:
Test was disabled since the relay did not handle closed connections
properly. This is no longer the case, so we can enable the test.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -89,17 +89,23 @@ func TestRelay(t *testing.T) {
 	})
 }
 
-func DisabledTestRelayHandlesCrashedPeers(t *testing.T) {
-	withRelayedEcho(t, func(_, server, client *Channel, ts *testutils.TestServer) {
-		ctx, cancel := NewContext(time.Second)
+func TestRelayHandlesClosedPeers(t *testing.T) {
+	opts := serviceNameOpts("test").SetRelayOnly().
+		// The relay will fail to connect when the server is closed.
+		AddLogFilter("Failed to connect to relay host", 1)
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		ctx, cancel := NewContext(300 * time.Millisecond)
 		defer cancel()
+
+		testutils.RegisterEcho(ts.Server(), nil)
+		client := ts.NewClient(serviceNameOpts("client"))
+		client.Peers().Add(ts.HostPort())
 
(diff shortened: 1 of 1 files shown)


relay_test.go
```diff
@@ -89,17 +89,23 @@ func TestRelay(t *testing.T) {
 	})
 }
 
-func DisabledTestRelayHandlesCrashedPeers(t *testing.T) {
-	withRelayedEcho(t, func(_, server, client *Channel, ts *testutils.TestServer) {
-		ctx, cancel := NewContext(time.Second)
+func TestRelayHandlesClosedPeers(t *testing.T) {
+	opts := serviceNameOpts("test").SetRelayOnly().
```
breerly: Is this an accepted formatting choice? Feel like a chain, when it is multi line, should probably be:

```
Thing().
ThingTwo().
ThingThree()
```

relay_test.go
```diff
@@ -89,17 +89,23 @@ func TestRelay(t *testing.T) {
...
+		client := ts.NewClient(serviceNameOpts("client"))
+		client.Peers().Add(ts.HostPort())
+
 		sc := client.GetSubChannel("test")
 		_, _, _, err := raw.CallSC(ctx, sc, "echo", []byte("fake-header"), []byte("fake-body"))
 		require.NoError(t, err, "Relayed call failed.")
 
-		// Simulate a server crash.
-		server.Close()
+		ts.Server().Close()
```
breerly: Used to seeing cleanup in a `defer` statement at the top of the func - why do it down here?

breerly:
lgtm (w a couple Qs)

relay_test.go
prashantv: Chaining is pretty uncommon in Go generally, I tend to only use it in tests for options. Generally it's all on one line or each option on a separate line, but in this set of tests, the `SetRelayOnly` is always required, and all the tests have the same pattern of `serviceNameOpts("..").SetRelayOnly`, so I left it one line to match the rest of the tests.

relay_test.go
prashantv: In this case, we're closing the server before the test ends, since we have more logic to execute below (the `CallSC`). If we used defer, the server wouldn't close until the end of the test function.

defer is useful for cleanup, but in this case, the `Close` is actually important to what's being tested

PR #451, opened 2016-07-20
Expose number of connections registered in the channel

prashantv:
Fixes #398 and is required for the tests fixing #450 

cc @nomis52

PR #452, opened 2016-07-20
Handle host:port mismatches between peer.Connect and remote peer

### LINKED ISSUE #450: Better handling in peer.Connect for host:port mismatches
If we create an outbound connection to a peer `127.0.0.1:21300` but it identifies itself as `1.1.1.1:21300`, currently we don't add the connection to a peer under `127.0.0.1:21300`.

If another call is made to the same host:port, we don't find the previous connection (since it's under another peer) and end up creating a new connection.

This ends up causing a large number of connections to be made. Previously, we only added the peer to `127.0.0.1:21300` and we ignored the connection host:port sent in the init req.

prashantv:
Currently, if we conect to peer as "a:1" and it sends us an init req
header with host:port set to "b:2", we only add it under the peer "b:2".

After this change, the connection will be added under both peers so that
future calls to "a:1" will be able to use the same connection.

When this happens, we log the host:port mismatch under a debug log.
In future, we may want to emit metrics when this happens.

This information (outboundHostPort vs remoteHostPort) is also exposed
via introspection.

Fixes #450

Note: tests will fail till #451 is landed

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -512,7 +512,19 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string) (*Connection, e
 		return nil, err
 	}
 
-	return c, err
+	// It's possible that the connection we just created responds with a host:port
+	// that is not what we tried to connect to. E.g., we may have connected to
+	// 127.0.0.1:1234, but the returned host:port may be 10.0.0.1:1234.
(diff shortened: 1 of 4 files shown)


akshayjshah:
:shipit:

PR #453, opened 2016-07-20
check_test_deps: Disallow "mock" in addition to "test" in package deps

prashantv:


akshayjshah:
:shipit: works for me.

PR #456, opened 2016-07-20
Allow introspection calls on the "tchannel" service

prashantv:
We register meta endpoints on the "tchannel" service:
https://github.com/uber/tchannel/blob/master/docs/meta.md

We should register the Go-specific meta endpoints on the same
service name.

This is useful when trying to debug a service given only the host:port.
The introspection endpoints can be used to expose the service name
information.

### DIFF
--- a/introspection_test.go
+++ b/introspection_test.go
@@ -55,6 +55,16 @@ func TestIntrospection(t *testing.T) {
 			"includeGoStacks": true,
 		}, &resp)
 		require.NoError(t, err, "Call _gometa_runtime failed")
+
+		if !ts.HasRelay() {
+			// Try making the call on the "tchannel" service which is where meta handlers
+			// are registered. This will only work when we call it directly as the relay
+			// will not forward the tchannel service.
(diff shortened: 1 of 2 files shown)


introspection_test.go
```diff
@@ -55,6 +55,17 @@ func TestIntrospection(t *testing.T) {
...
+
+		if !ts.HasRelay() {
+			// Try making the call on the "tchannel" service which is where meta handlers
+			// are registered. This will only work when we call it directly as the relay
+			// will not forward the tchannel service.
+			err = json.CallPeer(ctx, peer, "tchannel", "_gometa_runtime", map[string]interface{}{
+				"includeGoStacks": true,
+			}, &resp)
+			require.NoError(t, err, "Call _gometa_runtime failed")
+
```
akshayjshah: Extra newline?

prashantv: Fixed, thanks

akshayjshah:
:shipit:

PR #458, opened 2016-07-20
Fix flaky TestRelayHandlesClosedPeers due to connection error log

prashantv:
Since we're closing the server, we'll cause errors about connections closing, failed to connect etc. Disable logging to avoid spurious test failures.

cc @akshayjshah


akshayjshah:
:shipit: (╯°□°)╯︵ ┻━┻

PR #463, opened 2016-07-21
relay: Support for handling unfragmented calls on the relay channel

prashantv:
This allows users to register methods directly on the relay channel
itself rather than forwarding methods. This is required for methods like
the hyperbahn advertise which need information about the caller's
host:port as speciifed in the connection to the relay.

Since this is only intended to be used for hyperbahn's ad and for
TChannel introspection requests, both of which are extremely small, we
are keeping the implementation simple and only handling calls that are
not fragmented. This requires no extra tracking or references between
the relay item and mex.

Fragmented calls are rejected with a BadRequest error code.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -395,7 +395,7 @@ func TestRelayHandleLocalCall(t *testing.T) {
 		testutils.RegisterEcho(s2, nil)
 
 		client := ts.NewClient(nil)
-		require.NoError(t, testutils.CallEcho(client, ts.HostPort(), "s2", nil), "CallEcho to relayed service failed")
+		testutils.AssertEcho(t, client, ts.HostPort(), "s2")
 
 		testutils.RegisterEcho(ts.Relay(), nil)
 		testutils.AssertEcho(t, client, ts.HostPort(), "relay")
--- a/relay_test.go
+++ b/relay_test.go
@@ -387,7 +387,7 @@ func TestRelayOutgoingConnectionsEphemeral(t *testing.T) {
 
 func TestRelayHandleLocalCall(t *testing.T) {
 	opts := testutils.NewOpts().SetRelayOnly().
-		SetRelayLocal([]string{"relay", "tchannel", "test"}).
+		SetRelayLocal("relay", "tchannel", "test").
 		// We make a call to "test" for an unknown method.
 		AddLogFilter("Couldn't find handler.", 1)
 	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
@@ -412,7 +412,7 @@ func TestRelayHandleLocalCall(t *testing.T) {
 
 func TestRelayHandleLargeLocalCall(t *testing.T) {
 	opts := testutils.NewOpts().SetRelayOnly().
-		SetRelayLocal([]string{"relay"}).
+		SetRelayLocal("relay").
 		AddLogFilter("Received fragmented callReq", 1).
 		// Expect 4 callReqContinues for 256 kb payload that we cannot relay.
 		AddLogFilter("Failed to relay frame.", 4)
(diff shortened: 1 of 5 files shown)


testutils/channel_opts.go
```diff
@@ -193,6 +193,13 @@ func (o *ChannelOpts) SetRelayHosts(rh relay.Hosts) *ChannelOpts {
 	return o
 }
 
+// SetRelayLocal sets the channel's relay local handlers for service names
+// that should be handled by the relay channel itself.
+func (o *ChannelOpts) SetRelayLocal(relayLocal []string) *ChannelOpts {
+	o.ChannelOptions.RelayLocalHandlers = relayLocal
+	return o
+}
```
akshayjshah: If we're aiming for a fluent API here, would it be better to take a variadic number of strings and append them to any existing relay-local handlers?

relay_test.go
```diff
@@ -384,3 +384,49 @@ func TestRelayOutgoingConnectionsEphemeral(t *testing.T) {
...
+	opts := testutils.NewOpts().SetRelayOnly().
+		SetRelayLocal([]string{"relay", "tchannel", "test"}).
+		// We make a call to "test" for an unknown method.
+		AddLogFilter("Couldn't find handler.", 1)
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		s2 := ts.NewServer(serviceNameOpts("s2"))
+		testutils.RegisterEcho(s2, nil)
+
+		client := ts.NewClient(nil)
+		require.NoError(t, testutils.CallEcho(client, ts.HostPort(), "s2", nil), "CallEcho to relayed service failed")
```
akshayjshah: Any reason not to use `AssertEcho` here too? I know it does a bit more than `CallEcho`, but none of the extra checks seem inappropriate here.

prashantv: No reason not to, updated.

testutils/channel_opts.go
prashantv: Great idea, done.

prashantv:
Updated

relay_test.go
```diff
@@ -384,3 +384,55 @@ func TestRelayOutgoingConnectionsEphemeral(t *testing.T) {
...
+			Arg2: testutils.RandBytes(128 * 1024),
+			Arg3: testutils.RandBytes(128 * 1024),
+		})
+		if assert.Equal(t, ErrCodeBadRequest, GetSystemErrorCode(err), "Expected BadRequest for large call to relay") {
+			assert.Contains(t, err.Error(), "cannot receive fragmented calls")
+		}
+
+		// We may get an error before the call is finished flushing.
+		// Do a ping to ensure everything has been flushed.
+		ctx, cancel := NewContext(time.Second)
```
akshayjshah: Just for consistency, should we use the helper function to scale this timeout in CI?

testutils/channel_opts.go
```diff
@@ -193,6 +193,13 @@ func (o *ChannelOpts) SetRelayHosts(rh relay.Hosts) *ChannelOpts {
 	return o
 }
 
+// SetRelayLocal sets the channel's relay local handlers for service names
+// that should be handled by the relay channel itself.
+func (o *ChannelOpts) SetRelayLocal(relayLocal ...string) *ChannelOpts {
+	o.ChannelOptions.RelayLocalHandlers = relayLocal
```
akshayjshah: Do you want to assign or append here? Either one is fine with me, but appending seems more consistent with the fluent API design.

akshayjshah:
:shipit: nits aside, looks good.

relay_test.go
prashantv: Any test that uses low timeouts (like under 100 milliseconds) uses the timeout function, but anything with a large timeout (like `time.Second`) hasn't been using it in tests (we have a whole bunch of these littered throughout the code). I'd rather keep this consistent with what the other tests do.

In future, I'd like to add a `TestServer.NewContext` which will automatically `defer` at the end of the test so that callers like this could inline if they wanted. Maybe a `NewDefaultContext` which will make a context with some timeout that is scaled automatically.

testutils/channel_opts.go
prashantv: I think the `Set` prefix on the function should replace. If it was `AddRelayLocal` it would make more sense to append, but I don't think we need `AddRelayLocal` right now, since we set all the service names we'd like handled locally at once.

PR #465, opened 2016-07-25
Pass context to Thrift post response callback

robskillington:
cc @prashantv

### DIFF
--- a/thrift/thrift_test.go
+++ b/thrift/thrift_test.go
@@ -237,14 +237,14 @@ func TestClientHostPort(t *testing.T) {
 }
 
 func TestRegisterPostResponseCB(t *testing.T) {
-	key := "key"
-	value := "value"
-	opts := setupOptions{contextFn: func() (Context, func()) {
-		ctx, cancel := NewContext(time.Second)
-		ctx = WithHeaders(context.WithValue(ctx, key, value), ctx.Headers())
-		return ctx, cancel
-	}}
-	withSetupOptions(t, opts, func(ctx Context, args testArgs) {
+	withSetup(t, func(ctx Context, args testArgs) {
+		ctxKey := "key"
+		ctxValue := "value"
+
+		args.server.SetContextFn(func(ctx context.Context, method string, headers map[string]string) Context {
+			return WithHeaders(context.WithValue(ctx, ctxKey, ctxValue), headers)
+		})
+
 		arg := &gen.Data{
 			B1: true,
 			S2: "str",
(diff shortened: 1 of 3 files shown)


thrift/thrift_test.go
```diff
@@ -250,8 +257,9 @@ func TestRegisterPostResponseCB(t *testing.T) {
 		}
 
 		called := make(chan struct{})
-		cb := func(method string, response thrift.TStruct) {
+		cb := func(reqCtx context.Context, method string, response thrift.TStruct) {
 			assert.Equal(t, "Call", method)
+			assert.Equal(t, ctx.Value(key), value)
```
prashantv: nit: `Equal` should have expected argument before the value it's testing, so this should be,
```go
assert.Equal(t, value, ctx.Value(key))
```

thrift/thrift_test.go
```diff
@@ -237,7 +237,14 @@ func TestClientHostPort(t *testing.T) {
 }
 
 func TestRegisterPostResponseCB(t *testing.T) {
-	withSetup(t, func(ctx Context, args testArgs) {
+	key := "key"
```
prashantv: nit: can we call these `ctxKey` and `ctxValue` so it's clearer in the rest of the test

thrift/thrift_test.go
prashantv: I don't think this test is testing what you're intending for it to test. So right now, you're using the `contextFn` to create the context that's passed in to the to the test function. So you're basically testing that your test logic creates the context using the function you specified, and that it has the value you set.

What you actually want to test is: if you call `SetContextFn` (https://godoc.org/github.com/uber/tchannel-go/thrift#Server.SetContextFn) to override the context creation to include an extra value, the context you receive inside of the `cb` has that same value.

You probably want your `contextFn` to be used to call `SetContextFn` on a `thrift.Server` right after creating it.

prashantv:
The library change looks good, but the test isn't testing the right behaviour.

thrift/thrift_test.go
robskillington: Yeah I just wanted to test its the same context being threaded through, don't really mind about where exactly it's created from.

thrift/thrift_test.go
robskillington: Sure thing.

thrift/thrift_test.go
prashantv: The issue is that `reqCtx` is not being tested -- you could pass `nil` as a parameter to the context in the Thrift code, and the test would still pass.

The test should make sure that `reqCtx` is the same context that is returned from the context creation function set by `SetContextFn`

robskillington:
Updated @prashantv with changes, was a lot easier to test with `SetContextFn` that I originally had believed.

prashantv:
lgtm, thanks @robskillington

PR #467, opened 2016-07-30
Add crossdock tests for tracing context propagation

yurishkuro:
This change adds the logic to run tracing interop tests via [crossdock](http://github.com/crossdock/crossdock). At the moment there are no travis changes to publish docker images, I would like to add those as separate PRs once the corresponding Python changes are available. In the mean time I can test the interop by pushing the images manually.

### DIFF
--- a/crossdock/behavior/trace/behavior.go
+++ b/crossdock/behavior/trace/behavior.go
@@ -1,0 +1,254 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package trace
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"time"
+
+	"github.com/uber/tchannel-go"
+
+	"github.com/crossdock/crossdock-go"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+	"github.com/uber/jaeger-client-go"
+	"github.com/uber/jaeger-client-go/utils"
+	"golang.org/x/net/context"
+)
+
+// Different parameter keys and values used by the system
+const (
+	BehaviorName = "trace"
+)
+
+// Behavior is an implementation of "trace behavior", that verifies
+// that a tracing context and baggage are properly propagated through
+// two level of servers.
+type Behavior struct {
+	ServerPort    string
+	Tracer        opentracing.Tracer
+	ServiceToHost func(string) string
+	ch            *tchannel.Channel
+	thriftCall    DownstreamCall
+	jsonCall      DownstreamCall
+}
+
+// DownstreamCall is an encoding-agnostic abstraction of calling a downstream service.
+type DownstreamCall func(ctx context.Context, target *Downstream) (*Response, error)
+
+// Register function adds JSON and Thrift handlers to the server channel ch
+func (b *Behavior) Register(ch *tchannel.Channel) {
+	b.registerThrift(ch)
+	b.registerJSON(ch)
+}
+
+// Run executes the trace behavior
+func (b *Behavior) Run(t crossdock.T) {
+	logParams(t)
+	sampled, err := strconv.ParseBool(t.Param(sampledParam))
+	if err != nil {
+		t.Fatalf("Malformed param %s: %s", sampledParam, err)
+	}
+	baggage := randomBaggage()
+
+	level1 := &Request{
+		ServerRole: RoleS1,
+	}
+	server1 := t.Param(server1NameParam)
+
+	level2 := &Downstream{
+		ServiceName: t.Param(server2NameParam),
+		ServerRole:  RoleS2,
+		HostPort: fmt.Sprintf("%s:%s",
+			b.serviceToHost(t.Param(server2NameParam)),
+			b.ServerPort,
+		),
+		Encoding: t.Param(server2EncodingParam),
+	}
+	level1.Downstream = level2
+
+	level3 := &Downstream{
+		ServiceName: t.Param(server3NameParam),
+		ServerRole:  RoleS3,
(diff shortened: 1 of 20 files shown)


crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,236 @@
...
+	b.registerThriftServer(ch)
+	b.registerJSONServer(ch)
+}
+
+// Run executes the trace behavior
+func (b *Behavior) Run(t crossdock.T) {
+	sampled, err := strconv.ParseBool(t.Param(sampledParam))
+	if err != nil {
+		t.Fatalf("Malformed param %s: %s", sampledParam, err)
+		return
```
abhinav: You don't need to return after logging a fatal. It stops execution of the behavior right away.

crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,236 @@
...
+	ServerPort    string
+	Tracer        opentracing.Tracer
+	ServiceToHost func(string) string
+	ch            *tchannel.Channel
+	thriftCall    DownstreamCall
+	jsonCall      DownstreamCall
+}
+
+// DownstreamCall is used in a few other structs here
+type DownstreamCall func(ctx context.Context, target *Downstream) (*Response, error)
```
abhinav: Nit: Doc doesn't say anything about what DownstreamCall is/what purpose it serves.

crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,236 @@
...
+			b.serviceToHost(t.Param(server3NameParam)),
+			b.ServerPort,
+		),
+		Encoding: t.Param(server3EncodingParam),
+	}
+	level2.Downstream = level3
+
+	resp, err := b.startTrace(level1, sampled, baggage)
+	if err != nil {
+		t.Errorf(err.Error())
```
abhinav: Consider including more information with the error message.

crossdock/behavior/trace/thrift.go
```diff
@@ -0,0 +1,113 @@
...
+	"encoding/json"
+
+	"github.com/uber/tchannel-go"
+	"github.com/uber/tchannel-go/thrift"
+	gen "github.com/uber/tchannel-go/thrift/gen-go/test"
+
+	"golang.org/x/net/context"
+)
+
+func (b *Behavior) registerThriftServer(ch *tchannel.Channel) {
```
prashantv: nit: since this also registers the caller, maybe just call this `registerThrift`. same for json

crossdock/behavior/trace/thrift.go
```diff
@@ -0,0 +1,113 @@
...
+	serviceClient := gen.NewTChanSimpleServiceClient(thriftClient)
+	tReq, err := requestToThrift(req)
+	if err != nil {
+		return nil, err
+	}
+
+	tctx := thrift.Wrap(ctx)
+	res, err := serviceClient.Call(tctx, tReq)
+	if err != nil {
+		panic(err.Error())
```
prashantv: why panic instead of just returning the error?

crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,236 @@
...
+		return
+	}
+
+	traceID := resp.Span.TraceID
+	if traceID == "" {
+		t.Errorf("Trace ID is empty in S1(%s)", server1)
```
abhinav: FYI, if you use the Assertions API, you'll get a stack trace in the error message in case of failure.

https://godoc.org/github.com/crossdock/crossdock-go#Assertions

crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,236 @@
...
+		return
+	}
+
+	success := validateTrace(t, level1.Downstream, resp, server1, 1, traceID, sampled, baggage)
+	if success {
+		t.Successf("trace checks out")
+		log.Println("PASS")
+	}
```
abhinav: Do you need to log a failure if success is false?

crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,236 @@
...
+) bool {
+	success := true
+	if traceID != resp.Span.TraceID {
+		t.Errorf("Trace ID mismatch in S%d(%s): expected %s, received %s",
+			level, service, traceID, resp.Span.TraceID)
+		success = false
+	}
+	if baggage != resp.Span.Baggage {
+		t.Errorf("Baggage mismatch in S%d(%s): expected %s, received %s",
+			level, service, baggage, resp.Span.Baggage)
```
abhinav: You can compress all this to,

```go
checks := crossdock.Checks(t)
checks.Equal(traceID, resp.Span.TraceID, "Trace ID mismatch in S%d(%s)", level, service)
checks.Equal(baggage, resp.Span.Baggage, "Baggage mismatch in S%d(%s)", level, service)
checks.Equal(sampled, resp.Span.Sampled, ...)
// ...
```

crossdock/client/client.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+// Listen initializes the server
+func (c *Client) listen() error {
+	c.setDefaultPort(&c.ClientHostPort, ":"+common.DefaultClientPortHTTP)
+	c.setDefaultPort(&c.ServerPort, common.DefaultServerPort)
+
+	c.mux = http.NewServeMux()
+	c.mux.Handle("/", crossdock.Handler(c.Behaviors, true))
```
abhinav: You're not putting anything else on the mux. Why not use `crossdock.Handler` directly with `http.Serve`?

    http.Serve(c.listener, crossdock.Handler(c.Behaviors, true))

crossdock/client/client.go
```diff
@@ -0,0 +1,84 @@
...
+	return c.listener.Close()
+}
+
+// URL returns a URL that the client can be accessed on
+func (c *Client) URL() string {
+	return fmt.Sprintf("http://%s/", c.ClientHostPort)
+}
+
+func (c *Client) setDefaultPort(port *string, defaultPort string) {
+	if *port == "" {
```
abhinav: What if `port` is nil?

crossdock/main.go
```diff
@@ -0,0 +1,73 @@
...
+			trace.BehaviorName: behavior.Run,
+		},
+	}
+
+	if err := client.Start(); err != nil {
+		panic(err.Error())
+	}
+
+	select {}
+	println("main exiting")
```
abhinav: This line will never be reached, right? Why add it?

crossdock/main.go
```diff
@@ -0,0 +1,73 @@
...
+func main() {
+	tracer, tCloser := initTracer()
+	defer tCloser.Close()
+
+	server := &server.Server{Tracer: tracer}
+	if err := server.Start(); err != nil {
+		panic(err.Error())
+	} else {
+		defer server.Close()
+	}
```
abhinav: Why is the defer gated behind the `else`?

crossdock/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+	client := &client.Client{
+		Behaviors: crossdock.Behaviors{
+			trace.BehaviorName: behavior.Run,
+		},
+	}
+
+	if err := client.Start(); err != nil {
+		panic(err.Error())
+	}
```
abhinav: `defer client.Close()`?

abhinav:
Overall approach seems fine. I suggest using the [`Assertions` API](https://godoc.org/github.com/crossdock/crossdock-go#Assertions) from crossdock-go for asserting conditions to get readable error messages with stack traces by default.

crossdock/client/client.go
yurishkuro: Using default mux creates problem in unit tests

crossdock/behavior/trace/behavior.go
```diff
@@ -0,0 +1,254 @@
...
+	target *Downstream,
+	resp *Response,
+	service string,
+	level int,
+	traceID string,
+	sampled bool,
+	baggage string,
+) bool {
+	service = fmt.Sprintf("S%d(%s)", level, service)
+	checks := crossdock.Assert(t)
```
abhinav: You called this `checks` but you're using `crossdock.Assert`. Assert and Checks have slightly different behavior: Assert logs even in case of success. This looks like you wanted logging in case of failure only.

abhinav:
LGTM

PR #472, opened 2016-08-01
mockhyperbahn: use tchannel relaying to implement in-memory forwarding

prashantv:
Now that we have inbuilt relaying and ability to handle calls on relays, we can implement a simple in-memory Hyperbahn in the mockhyperbahn package!

### DIFF
--- a/testutils/mockhyperbahn/advertise_table.go
+++ b/testutils/mockhyperbahn/advertise_table.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


testutils/mockhyperbahn/advertise_table_test.go
```diff
@@ -0,0 +1,71 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package mockhyperbahn
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/uber/tchannel-go/testutils"
```
abhinav: import group ordering

abhinav:
LGTM with nit

testutils/mockhyperbahn/advertise_table_test.go
prashantv: Fixed

abhinav:
:shipit:

testutils/mockhyperbahn/advertise_table.go
```diff
@@ -0,0 +1,69 @@
...
+		return
+	}
+
+	hostPorts = append(hostPorts, hostPort)
+	sort.Strings(hostPorts)
+	t.serviceMapping[svc] = hostPorts
+}
+
+func (t *advertisedTable) Get(call relay.CallFrame) (relay.Peer, error) {
+	peers := t.serviceMapping[string(call.Service())]
```
akshayjshah: Don't we need to hold the lock here?

akshayjshah:
otherwise lgtm

testutils/mockhyperbahn/advertise_table.go
prashantv: Good catch! Fixed

PR #473, opened 2016-08-04
Let relay tables check the remote process name

akshayjshah:
In order to build some internal TChannel relays, we need the relay table to behave differently depending on who the remote peer is. Since this use case is only for a short-term migration, we'll identify the remote peer by process name. This PR makes two small changes:

* It adds an efficient way for connections to check whether the remote peer matches one of a pre-configured list of process name prefixes. This is checked only once, when the connection is established.
* It adds a narrow `relay.Conn` interface to expose this new method, and passes the connection into the routing table's `Get` method.

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -748,3 +748,19 @@ func TestConnectTimeout(t *testing.T) {
 		close(testComplete)
 	})
 }
+
+func TestConnectionProcessPrefixes(t *testing.T) {
+	opts := testutils.NewOpts().NoRelay().SetProcessName("nodejs-hyperbahn")
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		clientOpts := testutils.NewOpts().NoRelay().SetProcessPrefixes("nod", "nodejs-hyperbahn", "", "hyperbahn")
+		client := ts.NewClient(clientOpts)
+
+		ctx, cancel := NewContextBuilder(testutils.Timeout(100 * time.Millisecond)).Build()
+		defer cancel()
+		conn, err := client.Connect(ctx, ts.HostPort())
+		require.NoError(t, err, "Unexpected error connecting to test server.")
+
+		matches := conn.RemoteProcessPrefixMatches()
+		assert.Equal(t, []bool{true, true, true, false}, matches, "Unexpected prefix matches.")
+	})
+}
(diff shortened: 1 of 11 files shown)


relay.go
```diff
@@ -170,10 +170,11 @@ type Relayer struct {
 	// It stores remappings for all response frames read on this connection.
 	inbound *relayItems
 
-	peers   *PeerList
-	conn    *Connection
-	logger  Logger
-	pending atomic.Uint32
+	peers     *PeerList
+	conn      *Connection
+	relayConn relay.Conn // cast and allocate once
```
prashantv: Since the `conn` is a pointer already, the conversion to an interface is pretty cheap (no allocs). I'd drop the field and pass it directly

connection_test.go
```diff
@@ -748,3 +748,19 @@ func TestConnectTimeout(t *testing.T) {
...
 	})
 }
+
+func TestConnectionProcessPrefixes(t *testing.T) {
+	opts := testutils.NewOpts().NoRelay().SetProcessName("nodejs-hyperbahn")
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		clientOpts := testutils.NewOpts().NoRelay().SetProcessPrefixes("nod", "nodejs-hyperbahn", "", "hyperbahn")
+		client := ts.NewClient(clientOpts)
+
+		ctx, cancel := NewContextBuilder(testutils.Timeout(100 * time.Millisecond)).Build()
```
prashantv: Don't need the builder here since we're not setting any custom context properties. Since we don't rely on timeout in the test, we can do what most of the tests do, `ctx, cancel := NewContext(time.Second)`

connection_test.go
```diff
@@ -748,3 +748,19 @@ func TestConnectTimeout(t *testing.T) {
 		close(testComplete)
 	})
 }
+
+func TestConnectionProcessPrefixes(t *testing.T) {
+	opts := testutils.NewOpts().NoRelay().SetProcessName("nodejs-hyperbahn")
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		clientOpts := testutils.NewOpts().NoRelay().SetProcessPrefixes("nod", "nodejs-hyperbahn", "", "hyperbahn")
```
prashantv: nit: don't need `NoRelay` when creating a client

relay.go
```diff
@@ -266,7 +268,7 @@ func (r *Relayer) getDestination(f lazyCallReq, cs relay.CallStats) (*Connection
 	}
 
 	// Get the destination
-	selectedPeer, err := r.hosts.Get(f)
+	selectedPeer, err := r.hosts.Get(f, r.relayConn)
```
prashantv: (only do this if it's not a lot of extra work) Can we add a test to make sure that when the relay calls `Get`, this is not nil.

connection_test.go
akshayjshah: yeah...the default feels strange here. Will change.

prashantv:
lgtm

PR #474, opened 2016-08-04
introspection: Only include other channels if requested

prashantv:
In our internal relaying system, we'll be creating a large number
of relays, and the OtherChannels just floods the output, so disable
it by default. If we're debugging issues where we think there's multiple
channels in a process, we can enable the IncludeOtherChannels at that
point.


akshayjshah:
:sadpanda: that we need this, but lgtm

PR #475, opened 2016-08-05
json: Fix incorrect error message on connect failures

prashantv:
JSON prepends context on failures, but there was no context if
the connection failed. Add "connect" as the context when starting
the call.


akshayjshah:
Amended PR title to disambiguate "incorrect message" from "severe error". Otherwise LGTM.

PR #476, opened 2016-08-05
Expose routing delegates through relay.CallFrame

akshayjshah:
To make forwarding decisions, implementations of `relay.Hosts` need access to the
`rd` transport header. Expose it via the `relay.CallFrame` interface.

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -27,7 +27,10 @@ import (
 	"time"
 )
 
-var _callerNameKeyBytes = []byte(CallerName)
+var (
+	_callerNameKeyBytes      = []byte(CallerName)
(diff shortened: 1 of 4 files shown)


relay_messages.go
```diff
@@ -115,6 +117,9 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 		if bytes.Equal(key, _callerNameKeyBytes) {
 			cr.caller = val
 		}
+		if bytes.Equal(key, _routingDelegateKeyBytes) {
```
prashantv: nit: suggest `else if` since key can't match both.

prashantv:
lgtm with minor comment

PR #477, opened 2016-08-05
relay: Add support for outgoing calls from a relay

prashantv:
If an ID is not found in the relay items, try to forward the frame
using the outbound exchange. If that fails, we return the original
error that caused handleNonCallReq to fail.


akshayjshah:
impressed_okay

:shipit:

PR #480, opened 2016-08-09
Restore the original Registrar interface

yurishkuro:
Extending Registrar interface in #426 caused compile errors in ringpop-go. This change restores the original Registrar interface and pulls the tracer from it by casting to TracerProvider

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -34,6 +34,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+// TracerProvider returns an OpenTracing Tracer
+type TracerProvider interface {
+	Tracer() opentracing.Tracer
+}
+
 // zipkinSpanFormat defines a name for OpenTracing carrier format that tracer may support.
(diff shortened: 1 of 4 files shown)


tracing.go
```diff
@@ -34,6 +34,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+// TracerProvider returns an OpenTracing Tracer
+type TracerProvider interface {
```
prashantv: Is there any advantages to having this be public? While the other option would duplicate the interface, I wonder if it's better to keep the tchannel API smaller

yurishkuro: well, the only reason is that it's used from json and thrift packages. But we could have a public helper method in tracing.go to extract the tracer from the registrar, wdyt?

prashantv: Separate public method seems a little better, let's do that.

prashantv:
lgtm

PR #481, opened 2016-08-09
Enable crossdock tests in Travis

yurishkuro:
Image is published to `tchannelhub/xdock-go` by user `uberbot`

@abhinav @breerly


.travis.yml
```diff
@@ -7,12 +8,16 @@ go:
   - 1.6
   - 1.7rc6
   - tip
-install: make install_ci
 env:
- # Set higher timeouts for Travis
- - TEST_TIMEOUT_SCALE=20
+  global:
+    - DOCKER_COMPOSE_VERSION=1.6.0
```
breerly: Might as well run the latest version, `1.8.0`.

abhinav:
LGTM @breerly?

breerly:
Looking

.travis.yml
```diff
@@ -1,18 +1,55 @@
...
- - make test_ci
- - make cover_ci
- - make lint
+- make test_ci
+- make cover_ci
+- make lint
+- make crossdock_ci
 
+after_success:
+- export REPO=tchannelhub/xdock-go
```
breerly: this looks like it breaks convention - I would expect `tchannel/tchannel-go`

breerly:
:shipit:

.travis.yml
yurishkuro: well, (a) the org `tchannel` is taken, (b) we used `xdock-go` / `xdock-py` image names in Jaeger clients as well because these images actually have nothing to do with tchannel or Jaeger, they are test images, so it seems more appropriate to name them with "crossdock".

breerly: Aha 👍

.travis.yml
```diff
@@ -1,18 +1,55 @@
+sudo: required
```
prashantv: Hmm, this will slow down our Travis runs (which already aren't super fast to begin with)

I'm assuming that's needed to run docker? Is there any way we can only require sudo for some specific parameters, and only run the docker tests for a specific version?

breerly: If I remember correctly, It was needed to upgrade docker to a custom version. Could be, with a little work to the Docker Compose pieces below, that this could be removed. I could be wrong though.

breerly: nm, looks like it is required

https://docs.travis-ci.com/user/docker/

breerly: We're probably better off running Crossdock for every version, unless there is a big time savings.

yurishkuro: we could do that, but we still want to publish the image only once after_success

prashantv: It looks like the run time went from ~3-4 minutes to ~5-7 minutes. I really want Travis to be < 5 minutes for a change, and we've already been creeping up slowly, this one jumps up significantly.

Since 99% of TChannel PRs don't touch tracing, I don't want every Travis run for each small change to get significantly slower without a large benefit.

breerly: Perhaps we could use Travis' matrix feature to parallelize the regular test suite with the Crossdock tests.

yurishkuro: that's what I am going in tchannel-python, as it already uses the matrix. Not quite sure how to do it in Go.

yurishkuro:
@prashantv @breerly ok, as usual with the makefile, quite a bit of nastiness, but the crossdock test now runs as a separate/parallel build step, and the image is only published from that step

.travis.yml
```diff
@@ -1,18 +1,58 @@
...
+dist: trusty
```
yurishkuro: issue in Travis - bottom of https://github.com/travis-ci/travis-ci/issues/5142

crossdock/docker-compose.yml
```diff
@@ -4,16 +4,16 @@ services:
     crossdock:
         image: crossdock/crossdock
         links:
-            - go
+            - go, python
```
breerly: I believe I trim whitespace, but you probably don't want a space here.

yurishkuro:
This is now using official python build.

@prashantv I had a look around, but didn't find a way of not requiring sudo in every build. Since I've converted the build to manually controlled matrix, we can divide it into smaller pieces controlled by env variable, i.e. run a test of a subset of packages.

prashantv:
Where we set `CROSSDOCK=true`, I think we can also set `sudo: true` there?

http://stackoverflow.com/questions/34668468/use-sudo-true-with-single-element-of-build-matrix-in-travis-ci

yurishkuro:
for some reason, when I tried that, all the main tests are failing. The
crossdock test did succeed though.

https://travis-ci.org/uber/tchannel-go/builds/151583187#L540

      === RUN   TestInitialAdvertiseFailedRetryTimeout

--- FAIL: TestInitialAdvertiseFailedRetryTimeout (0.02s)

	Error Trace:	advertise_test.go:91

			advertise_test.go:376

			advertise_test.go:92

	Error:		Not equal: 25 (expected)

			        != 0 (actual)

	Messages:	adHandler not retried correct number of times


On Thu, Aug 11, 2016 at 1:32 PM, Prashant Varanasi <notifications@github.com

prashantv:
Hmm, it almost seems like an incompatibility with where it's running. If you look at previous green runs, you'll see:
`This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.`
E.g., https://travis-ci.org/uber/tchannel-go/jobs/146264859

However, I don't see that in the latest failed runs. I wonder if we have to specify `sudo: false` at the top, and then override it?

yurishkuro:
@prashantv good call, I think it fixed it, the 1.5 step came in green

prashantv:
Glad it worked! lgtm. I'll squash and land this.

PR #482, opened 2016-08-10
relay: Expose RemoteHostPort() on relay.Conn

prashantv:
This also unexports RemoteProcessPrefixMatches from the Connection type.

This is required if we need to make relay decisions based on the host:port that is making the call.

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -33,8 +33,7 @@ type Conn interface {
 	// slices. Callers shouldn't mutate the returned slice.
 	RemoteProcessPrefixMatches() []bool
 
-	// RemotePeerInfo returns the peer info for the remote peer which contains
-	// the process name and host:port.
+	// RemoteHostPort returns the host:port of the remote peer.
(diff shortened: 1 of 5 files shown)


relay/relay.go
```diff
@@ -28,8 +28,14 @@ package relay
...
 type Conn interface {
 	// RemoteProcessPrefixMatches checks whether the remote peer's process name
-	// matches a preconfigured list of prefixes.
+	// matches a preconfigured list of prefixes specified in the connection
+	// options. It's the caller's responsibility to match indices between the two
+	// slices. Callers shouldn't mutate the returned slice.
 	RemoteProcessPrefixMatches() []bool
+
+	// RemotePeerInfo returns the peer info for the remote peer which contains
+	// the process name and host:port.
```
akshayjshah: s/RemotePeerInfo/RemoteHostPort, update comment to match.

akshayjshah:
apart from comment nit, :shipit:

PR #485, opened 2016-08-16
Don't send calls with 0 ttl even for low TTLs

### LINKED ISSUE #484: Don't send call req with 0 TTL
We currently check if the context has expired before sending calls here:
https://github.com/uber/tchannel-go/blob/dev/outbound.go#L58

This check tries to avoid sending frames with 0 TTL, but we're checking a `time.Duration` which is nanoseconds rather than milliseconds. Since tchannel uses millisecond granularity, we should instead check for < time.Millisecond.

prashantv:
Fixes #484


akshayjshah:
:shipit:

PR #486, opened 2016-08-17
Restore trace package to make glide happy

yurishkuro:
When doing `glide update` on other apps that depend on tchannel glide fails trying to look into the `trace` package that was removed in #426.


yurishkuro:
@prashantv

prashantv:
lgtm

PR #487, opened 2016-08-18
Make fields in test.thrift required, as they are in tchannel-python

yurishkuro:


prashantv:
lgtm -- surprised none of the tests needed changing for this.

yurishkuro:
Not sure how the struct was generated, but the fields were [not optional there](https://github.com/uber/tchannel-go/blob/dev/thrift/gen-go/test/ttypes.go#L23) to start with

```go
type Data struct {
	B1 bool   `thrift:"b1,1" db:"b1" json:"b1"`
	S2 string `thrift:"s2,2" db:"s2" json:"s2"`
	I3 int32  `thrift:"i3,3" db:"i3" json:"i3"`
}
```

I think they were implied as required in the absence of a qualifier.

PR #488, opened 2016-08-20
Parse host name as IPv4/IPv6 and set the respective PeerHostIPxx tags

yurishkuro:


tracing_internal_test.go
```diff
@@ -86,6 +88,41 @@ func TestSpanString(t *testing.T) {
...
+	span.Finish()
+	rawSpan = tracer.FinishedSpans()[1]
+	assert.Equal(t, "adhoc123", rawSpan.Tag(string(ext.PeerHostname)))
+	assert.Nil(t, rawSpan.Tag(string(ext.PeerPort)))
+
+	span = tracer.StartSpan("x")
+	setPeerHostPort(span, "10.20.30.40:321")
+	span.Finish()
+	rawSpan = tracer.FinishedSpans()[2]
+	ip := (((10<<8)|20)<<8|30)<<8 | 40
```
prashantv: nit: this may be easier to read as:
```
ip := 10<<24 | 20<<16 | 30<<8 | 40
```

tracing_internal_test.go
```diff
@@ -86,6 +88,41 @@ func TestSpanString(t *testing.T) {
 	assert.Equal(t, "TraceID=f,ParentID=0,SpanID=0", span.String())
 }
 
+func TestSetPeerHostPort(t *testing.T) {
+	tracer := mocktracer.New()
+
+	span := tracer.StartSpan("x")
```
prashantv: can you make this block a table test,
```
tests := []struct{
  hostPort string
  hostTag string
  wantHost interface{}
  wantPort int 
}{...}
```

tracing_internal_test.go
```diff
@@ -86,6 +88,41 @@ func TestSpanString(t *testing.T) {
 	assert.Equal(t, "TraceID=f,ParentID=0,SpanID=0", span.String())
 }
 
+func TestSetPeerHostPort(t *testing.T) {
```
prashantv: do we have tests for the following cases:
 - host:port with invalid port
 - no port (just host) -- something where `SplitHostPort` would fail

tracing_internal_test.go
```diff
@@ -86,6 +88,41 @@ func TestSpanString(t *testing.T) {
...
+	span.Finish()
+	rawSpan = tracer.FinishedSpans()[2]
+	ip := (((10<<8)|20)<<8|30)<<8 | 40
+	assert.Equal(t, uint32(ip), rawSpan.Tag(string(ext.PeerHostIPv4)))
+	assert.Equal(t, uint16(321), rawSpan.Tag(string(ext.PeerPort)))
+
+	span = tracer.StartSpan("x")
+	ipv6 := []byte{1, 2, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 16}
+	assert.Equal(t, 16, len(ipv6))
+	setPeerHostPort(span, fmt.Sprintf("[%s]:789", net.IP(ipv6)))
```
prashantv: nit: use `%v` instead of `%s` in the format string. `%s` should only be used for strings / byte slices

tracing_internal_test.go
```diff
@@ -86,6 +88,41 @@ func TestSpanString(t *testing.T) {
...
+	span = tracer.StartSpan("x")
+	setPeerHostPort(span, "10.20.30.40:321")
+	span.Finish()
+	rawSpan = tracer.FinishedSpans()[2]
+	ip := (((10<<8)|20)<<8|30)<<8 | 40
+	assert.Equal(t, uint32(ip), rawSpan.Tag(string(ext.PeerHostIPv4)))
+	assert.Equal(t, uint16(321), rawSpan.Tag(string(ext.PeerPort)))
+
+	span = tracer.StartSpan("x")
+	ipv6 := []byte{1, 2, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 16}
```
prashantv: can do:
```go
ipv6 := net.IP{1, 2, 3, [....]}
assert.Equal(t, net.IPv6len, len(ipv6))
```

then you can also get rid of the cast in the `setPeerHostPort` call

prashantv:
Overall looks good, just a few minor comments on the tests, and want to make sure we have all the cases covered.

tracing_internal_test.go
yurishkuro: I had one with `bad-port`, and added one without port.

tracing.go
```diff
@@ -252,10 +253,26 @@ func ExtractInboundSpan(ctx context.Context, call *InboundCall, headers map[stri
 
 func setPeerHostPort(span opentracing.Span, hostPort string) {
```
yurishkuro: @prashantv ideally, we would be getting `(host, port)` as proper types, it must be available somewhere in TChannel already. This method is purely internal, so it's rather wasteful to be parsing and converting host:port from string. It's called from the following places, where we have remote peer object, but it only exposes host:port as a string:

```
./tracing.go:157:	setPeerHostPort(span, c.remotePeerInfo.HostPort)
./tracing.go:209:	setPeerHostPort(span, c.remotePeerInfo.HostPort)
./tracing.go:248:		setPeerHostPort(span, call.RemotePeer().HostPort)
```

Another option is to cache it as `map[string]struct{host, hostTag, port}` in some LRU cache.

yurishkuro: PS: I opened an issue in OpenTracing (https://github.com/opentracing/opentracing.github.io/issues/116) to support a standard tag `peer.hostPort` so that the parsing logic could be consolidated in the tracer, instead of being duplicated across the RPC libraries instrumented with OT.

prashantv: I don't think we actually have this information within TChannel currently. We just have `HostPort` everywhere since that's what `net.Listen` takes, and `ln.Addr().String()` returns.

Instead of a new `peer.HostPort` tag, what about if OpenTracing had a helper to set the host/port tags based on similar logic to what you have in here?

prashantv:
lgtm

tracing.go
yurishkuro: That's the issue I linked above. I'm getting a bit of push back, but we can always go with a custom tag `peer.hostPort` (won't be portable with other OT tracers if the tag doesn't get standardized).

The good thing about that tag is that when we sample 1 in 1000 spans, setting that tag is no-op, while here we always try to parse the string even though the trace may not be sampled.

prashantv: Makes a lot of sense to avoid the parsing if we don't need it. If the `peer.hostPort` is not added, perhaps we should cache these values in the peer object so we don't need to reparse it constantly.

prashantv:
lgtm, I like the caching approach. Is something similar required for the local peer info?

yurishkuro:
No, the local peer is not producing any tags on the span. Generally the span producer is identified by the tracer automatically. If we want richer info like local peer port, we could add that instrumentation later.

PR #492, opened 2016-08-27
Reorder fields to make structs smaller

prashantv:
Used [maligned](https://github.com/mdempsky/maligned) to find structs that could be smaller. Pre-change:
```
  connection.go:145:17: struct of size 488 could be 480
  connection.go:186:20: struct of size 48 could be 40
  context.go:38:24: struct of size 56 could be 48
  context_builder.go:31:21: struct of size 96 could be 80
  fragmenting_reader.go:42:23: struct of size 56 could be 48
  introspection.go:137:29: struct of size 216 could be 208
  relay.go:49:16: struct of size 80 could be 72
```
Post-change:
```
  connection.go:145:17: struct of size 488 could be 480
  introspection.go:137:29: struct of size 216 could be 208
```

I only made changes when the structs are small and it matters -- e.g,
connection is already huge, so I'd rather keep the fields in a logical
order. Introspection struct doesn't matter to perf.

No changes except for field reorderings in this change.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -184,10 +184,10 @@ type Connection struct {
 	// remotePeerAddress is used as a cache for remote peer address parsed into individual
 	// components that can be used to set peer tags on OpenTracing Span.
 	remotePeerAddress struct {
+		port     uint16
 		ipv4     uint32
 		ipv6     string
 		hostname string
-		port     uint16
 	}
 }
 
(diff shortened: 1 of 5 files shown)


blampe:
Does this have a noticeable perf impact?

prashantv:
I was seeing ~1-2% differences, which is within the variance between runs so not sure.

Figured it's worth it since it was a pretty trivial change that doesn't make the code less clear, and less cache impact might help outside of benchmarking.

akshayjshah:
Sure, lgtm.

abhinav:
👍

PR #495, opened 2016-09-08
relay: Add more of the callReq frame to warn/error logs

prashantv:


relay.go
```diff
@@ -274,7 +274,12 @@ func (r *Relayer) canHandleNewCall() bool {
 
 func (r *Relayer) getDestination(f lazyCallReq, cs relay.CallStats) (*Connection, bool, error) {
 	if _, ok := r.outbound.Get(f.Header.ID); ok {
-		r.logger.WithFields(LogField{"id", f.Header.ID}).Warn("received duplicate callReq")
+		r.logger.WithFields(
+			LogField{"id", f.Header.ID},
+			LogField{"caller", string(f.Caller())},
+			LogField{"dest", string(f.Service())},
```
akshayjshah: Can we stick to caller/callee or source/dest? Mixing metaphors is a bit confusing.

akshayjshah:
o/w lgtm.

relay.go
prashantv: sure

PR #496, opened 2016-09-13
Support routing key transport header and call option

kriskowal:
We’ve encountered a need for a third way to address RPC, by service name, by the service name of a routing delegate, and now by a routing key, all of which now refer to traffic groups with the new traffic control system. This change allows us to thread a routing key through from TChannel call options into the corresponding transport header such that it can be examined by a relay.

r @prashantv

### DIFF
--- a/calloptions_test.go
+++ b/calloptions_test.go
@@ -30,6 +30,7 @@ func TestSetHeaders(t *testing.T) {
 	tests := []struct {
 		format          Format
 		routingDelegate string
+		routingKey      string
 		expectedHeaders transportHeaders
 	}{
 		{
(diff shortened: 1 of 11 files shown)


calloptions_test.go
```diff
@@ -49,12 +50,21 @@ func TestSetHeaders(t *testing.T) {
 				RoutingDelegate: "xpr",
 			},
 		},
+		{
+			format:     JSON,
+			routingKey: "canary",
```
breerly: Classy 💯

relay_messages.go
```diff
@@ -118,6 +119,8 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 			cr.caller = val
 		} else if bytes.Equal(key, _routingDelegateKeyBytes) {
 			cr.delegate = val
+		} else if bytes.Equal(key, _routingKeyKeyBytes) {
```
prashantv: Thanks for updating this as well. Can you also add some test cases for this in `relay_messages_test.go`

prashantv:
I didn't like the structure for the relay message tests so I refactored it a little and added a routing key test. Landing now.

Raynos:
This "rk" header is not documented in the protocol doc ( https://github.com/uber/tchannel/blob/master/docs/protocol.md ).

`sk` and `rd` are documented. Can we update the protocol with more information ?

PR #500, opened 2016-09-20
Add new connection lock to avoid creating more than one new connections

xichen2020:
cc @prashantv

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -346,6 +346,7 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 		return activeConn, nil
 	}
 
+	// Lock here to restrict new connection creation attempts to one goroutine
 	p.newConnLock.Lock()
 	defer p.newConnLock.Unlock()
 
--- a/peer.go
+++ b/peer.go
@@ -273,6 +273,7 @@ type Peer struct {
 	scCount uint32
 
 	// connections are mutable, and are protected by the mutex.
+	newConnLock         sync.Mutex
 	inboundConnections  []*Connection
 	outboundConnections []*Connection
 	chosenCount         atomic.Uint64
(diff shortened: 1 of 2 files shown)


peer.go
```diff
@@ -345,6 +346,14 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 		return activeConn, nil
 	}
 
+	p.newConnLock.Lock()
```
prashantv: Can you add a comment above this saying why we want to do this (to restrict pending connection attempts per peer)

peer_test.go
```diff
@@ -189,6 +189,48 @@ func TestPeerRemoveClosedConnection(t *testing.T) {
...
+
+	WithVerifiedServer(t, nil, func(ch *Channel, hostPort string) {
+		client := testutils.NewClient(t, nil)
+		defer client.Close()
+
+		var (
+			wg          sync.WaitGroup
+			lock        sync.Mutex
+			conn        *Connection
+			numRequests = 10
```
prashantv: nit: s/numRequests/numWorkers

or just concurrency.

peer_test.go
```diff
@@ -189,6 +189,48 @@ func TestPeerRemoveClosedConnection(t *testing.T) {
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				c, err := p.GetConnection(ctx)
+				require.NoError(t, err, "GetConnection failed")
+
+				lock.Lock()
+				defer lock.Unlock()
+
+				if conn == nil {
```
prashantv: This is a data race since multiple goroutines will be checking `conn`, suggest instead using a slice,
```go
conns := make([]*Connection, concurrency)
[....]
go func(i int) {
  [...]
  conns[i] = c
}
```

and then after the `wg.Wait`, verify that they're all the same

peer.go
xichen2020: Sure thing

peer_test.go
xichen2020: Can do

peer_test.go
xichen2020: hm they should be protected by the lock no?

peer_test.go
prashantv: I missed the lock, you're right. That's fine then!

peer.go
xichen2020: Done

peer_test.go
xichen2020: Done

prashantv:
lgtm, will squash and merge once Travis is green. Thanks @xichen2020!

PR #502, opened 2016-09-21
mockhb: Use TChannel peer selection to avoid unconnected peers

prashantv:
Instead of using a custom peer selection algorithm for mockHB, we can
use isolated subchannels and rely on tchannel's peer selection instead.
This will automatically avoid unconnected peers.


testutils/mockhyperbahn/hyperbahn_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/uber-go/atomic"
```
abhinav: wrong import group

testutils/mockhyperbahn/hyperbahn_test.go
```diff
@@ -118,3 +121,52 @@ func TestMockForwards(t *testing.T) {
...
+	require.True(t, moe1Called.Load(), "moe1 not called")
+	require.True(t, moe2Called.Load(), "moe2 not called")
+
+	// If moe2 is brought down, all calls should now be sent to moe1.
+	moe2.Close()
+
+	// Wait for the mock HB to have 0 connections to moe
+	for start := time.Now(); time.Since(start) < time.Second; time.Sleep(time.Millisecond) {
+		in, out := mockHB.Channel().Peers().GetOrAdd(moe2.PeerInfo().HostPort).NumConnections()
+		if in+out == 0 {
```
abhinav: Can't you use testutils.WaitFor here?

testutils/mockhyperbahn/hyperbahn_test.go
```diff
@@ -118,3 +121,52 @@ func TestMockForwards(t *testing.T) {
...
+func TestMockIgnoresDown(t *testing.T) {
+	mockHB, err := mockhyperbahn.New()
+	require.NoError(t, err, "Failed to set up mock hyperbahm")
+
+	var (
+		moe1Called atomic.Bool
+		moe2Called atomic.Bool
+	)
+
+	moe1 := newAdvertisedEchoServer(t, "moe", mockHB, func() { moe1Called.Store(true) })
```
abhinav: Is it worth tracking the counts rather than true/false

testutils/mockhyperbahn/hyperbahn.go
```diff
@@ -108,11 +123,23 @@ func (h *Mock) Configuration() hyperbahn.Configuration {
...
+	return h.ch.GetSubChannel(svc, tchannel.Isolated)
+}
+
 func (h *Mock) adHandler(ctx json.Context, req *hyperbahn.AdRequest) (*hyperbahn.AdResponse, error) {
+	callerHostPort := tchannel.CurrentCall(ctx).RemotePeer().HostPort
 	h.Lock()
 	for _, s := range req.Services {
 		h.advertised = append(h.advertised, s.Name)
-		h.table.addPeer(s.Name, tchannel.CurrentCall(ctx).RemotePeer().HostPort)
+		sc := h.ch.GetSubChannel(s.Name, tchannel.Isolated)
```
prashantv: Removed `h.getSC`, I wrote it thinking we'd use it in multiple places, but the table is a separate type.

testutils/mockhyperbahn/hyperbahn_test.go
prashantv: Done

testutils/mockhyperbahn/hyperbahn_test.go
prashantv: I just want to make sure load is dsitributed, it won't be 100% even because the peer selection is kind of random, so we can't assert any specific numbers

testutils/mockhyperbahn/hyperbahn_test.go
prashantv: Done

PR #504, opened 2016-09-24
relay: Selected peers should only be added to root peers

prashantv:
Previously, the relay was adding the peer returned from RelayHosts.Get
to the shared peer list, which could lead to unexpected results if the
channel's shared peer list is also used to talk to some other service.

We should never add peers to the shared peer list, if the user wants to
add the peer, they can add it to the shared peer list, but the relay
should only access peers through the root peer list.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -176,7 +176,7 @@ type Relayer struct {
 	// It stores remappings for all response frames read on this connection.
 	inbound *relayItems
 
-	peers   *PeerList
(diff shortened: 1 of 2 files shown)


prashantv:
Merging to unblock some other work, but @abhinav can you review when you get a chance.

abhinav:
Sounds reasonable. 👍

PR #506, opened 2016-09-29
Clone headers when injecting tracing data for outbound request

### LINKED ISSUE #505: Concurrent calls using the same context with headers could lead to a panic
The thrift client calls `InjectOutboundSpan` with the headers passed in to the context,
https://github.com/uber/tchannel-go/blob/dev/thrift/client.go#L69

This method will modify the passed in map (if it's not nil):
https://github.com/uber/tchannel-go/blob/dev/tracing.go#L178

If there's concurrent calls using the same context with the same map, they will both try to modify the map (causing a data race) which since go1.7 may also trigger the following panic:
```
fatal error: concurrent map writes
```

@yurishkuro

yurishkuro:
Fixes #505

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -166,21 +166,28 @@ func (c *Connection) startOutboundSpan(ctx context.Context, serviceName, methodN
 // when the outbound call is initiated. The tracing API is used to serialize the span
 // into the application `headers`, which will propagate tracing context to the server.
 // Returns modified headers containing serialized tracing context.
+//
+// Sometimes caller pass a shared instance of the `headers` map, so instead of modifying
+// it we clone it into the new map (assuming that Tracer actually injects some tracing keys).
 func InjectOutboundSpan(response *OutboundCallResponse, headers map[string]string) map[string]string {
 	span := response.span
 	if span == nil {
 		return headers
 	}
-	if headers == nil {
-		headers = make(map[string]string)
-	}
-	carrier := tracingHeadersCarrier(headers)
+	newHeaders := make(map[string]string)
+	carrier := tracingHeadersCarrier(newHeaders)
 	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, carrier); err != nil {
 		// Something had to go seriously wrong for Inject to fail, usually a setup problem.
 		// A good Tracer implementation may also emit a metric.
 		response.log.WithFields(ErrField(err)).Error("Failed to inject tracing span.")
 	}
-	return headers
+	if len(newHeaders) == 0 {
+		return headers
+	}
+	for k, v := range headers {
+		newHeaders[k] = v
+	}
+	return newHeaders
 }
 
 // extractInboundSpan attempts to create a new OpenTracing Span for inbound request
--- a/tracing.go
+++ b/tracing.go
@@ -182,7 +182,7 @@ func InjectOutboundSpan(response *OutboundCallResponse, headers map[string]strin
 		response.log.WithFields(ErrField(err)).Error("Failed to inject tracing span.")
 	}
 	if len(newHeaders) == 0 {
-		return headers
+		return headers // Tracer did not add any tracing headers, so return the original map
 	}
 	for k, v := range headers {
 		newHeaders[k] = v
--- a/tracing_test.go
+++ b/tracing_test.go
@@ -144,35 +144,36 @@ func TestTracingSpanAttributes(t *testing.T) {
 // made with the same shared map used as headers were causing panic due to
 // concurrent writes to the map when injecting tracing headers.
 func TestReusableHeaders(t *testing.T) {
-	tracer := mocktracer.New()
-
 	opts := &testutils.ChannelOpts{
-		ChannelOptions: ChannelOptions{Tracer: tracer},
-		DisableRelay:   true,
+		ChannelOptions: ChannelOptions{Tracer: mocktracer.New()},
 	}
 	WithVerifiedServer(t, opts, func(ch *Channel, hostPort string) {
 		jsonHandler := &JSONHandler{TraceHandler: testtracing.TraceHandler{Ch: ch}, t: t}
 		json.Register(ch, json.Handlers{"call": jsonHandler.callJSON}, jsonHandler.onError)
 
 		span := ch.Tracer().StartSpan("client")
+		traceID := span.(*mocktracer.MockSpan).SpanContext.TraceID // for validation
 		ctx := opentracing.ContextWithSpan(context.Background(), span)
 
-		ctx, cancel := NewContextBuilder(2 * time.Second).SetParentContext(ctx).Build()
+		sharedHeaders := map[string]string{"life": "42"}
+		ctx, cancel := NewContextBuilder(2 * time.Second).
+			SetHeaders(sharedHeaders).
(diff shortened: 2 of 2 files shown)


tracing.go
```diff
@@ -166,21 +166,28 @@ func (c *Connection) startOutboundSpan(ctx context.Context, serviceName, methodN
...
-	carrier := tracingHeadersCarrier(headers)
+	newHeaders := make(map[string]string)
+	carrier := tracingHeadersCarrier(newHeaders)
 	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, carrier); err != nil {
 		// Something had to go seriously wrong for Inject to fail, usually a setup problem.
 		// A good Tracer implementation may also emit a metric.
 		response.log.WithFields(ErrField(err)).Error("Failed to inject tracing span.")
 	}
-	return headers
+	if len(newHeaders) == 0 {
```
junchaowu: here is len(newHeaders) == 0 or len(headers) == 0 ?

yurishkuro: newHeaders - it's checking that tracer did not include any keys, thus the original map can be returned

prashantv:
Can we add a test that triggers a data race without the fix, that should be fixed by this change?

yurishkuro:
@prashantv is it ok to use `import "golang.org/x/sync/errgroup"`?

prashantv:
@yurishkuro we can but I don't think we need it, since we are in a test context and can `assert.NoError`
```go

ctx, cancel := // create context with a single headers map
defer cancel() 

var sync.WaitGroup
for i := 0; i < 5; i++ {
  wg.Add(1)
  go func() {
    defer wg.Done()
    for i := 0; i < 10; i++ {
      json.Call(ctx, ....)
    }
  }
}
wg.Wait()
```

That should be enough to trigger the race right?

tracing.go
```diff
@@ -166,21 +166,28 @@ func (c *Connection) startOutboundSpan(ctx context.Context, serviceName, methodN
...
 	if span == nil {
 		return headers
 	}
-	if headers == nil {
-		headers = make(map[string]string)
-	}
-	carrier := tracingHeadersCarrier(headers)
+	newHeaders := make(map[string]string)
+	carrier := tracingHeadersCarrier(newHeaders)
 	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, carrier); err != nil {
```
eculver: It's hard to see what's actually going on here because the `carrier` is actually the one operating on the headers map. It seems more reasonable for the tracer to always make a new headers map on `Inject` rather than try to understand all the implicit mutations that are going on. It's probably outside the scope of this PR, but I think it should be the responsibility of `Tracer().Inject()` to create the new map if it's going to make these assumptions.

yurishkuro: the point of Inject() is to affect the outgoing request's headers

tracing_test.go
```diff
@@ -138,3 +139,43 @@ func TestTracingSpanAttributes(t *testing.T) {
...
+
+// Per https://github.com/uber/tchannel-go/issues/505, concurrent client calls
+// made with the same shared map used as headers were causing panic due to
+// concurrent writes to the map when injecting tracing headers.
+func TestReusableHeaders(t *testing.T) {
+	tracer := mocktracer.New()
+
+	opts := &testutils.ChannelOpts{
+		ChannelOptions: ChannelOptions{Tracer: tracer},
+		DisableRelay:   true,
```
prashantv: do you need to disable the relay here?

yurishkuro:
@prashantv I did reproduce the panic, but thought doing `require.NoError` inside go-routines is not a good practice. Anyway, I've added the test.

tracing_test.go
```diff
@@ -138,3 +139,43 @@ func TestTracingSpanAttributes(t *testing.T) {
...
+
+		peer := ch.Peers().GetOrAdd(ch.PeerInfo().HostPort)
+		sharedHeaders := map[string]string{"life": "42"}
+
+		var wg sync.WaitGroup
+		for i := 0; i < 42; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				jsonCtx := json.WithHeaders(ctx, sharedHeaders)
```
prashantv: we should be able to create the context once at the top, i'd suggest
```go
ctx, cancel := NewContextBuilder(2 * time.Second).SetHeaders(sharedHeaders).Build()
```

tracing_test.go
```diff
@@ -138,3 +139,43 @@ func TestTracingSpanAttributes(t *testing.T) {
...
+	opts := &testutils.ChannelOpts{
+		ChannelOptions: ChannelOptions{Tracer: tracer},
+		DisableRelay:   true,
+	}
+	WithVerifiedServer(t, opts, func(ch *Channel, hostPort string) {
+		jsonHandler := &JSONHandler{TraceHandler: testtracing.TraceHandler{Ch: ch}, t: t}
+		json.Register(ch, json.Handlers{"call": jsonHandler.callJSON}, jsonHandler.onError)
+
+		span := ch.Tracer().StartSpan("client")
+		ctx := opentracing.ContextWithSpan(context.Background(), span)
```
prashantv: do we need to create a span, is that relevant to this test

tracing_test.go
```diff
@@ -138,3 +139,43 @@ func TestTracingSpanAttributes(t *testing.T) {
...
+
+		var wg sync.WaitGroup
+		for i := 0; i < 42; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				jsonCtx := json.WithHeaders(ctx, sharedHeaders)
+				var response testtracing.TracingResponse
+				require.NoError(t,
+					json.CallPeer(jsonCtx, peer, ch.PeerInfo().ServiceName,
```
prashantv: i think you can do `ch.ServiceName()` instead

tracing_test.go
```diff
@@ -138,3 +139,43 @@ func TestTracingSpanAttributes(t *testing.T) {
...
+						"call", &testtracing.TracingRequest{}, &response))
```
prashantv: can we add a message saying that "json.Call failed"

tracing_test.go
```diff
@@ -138,3 +139,43 @@ func TestTracingSpanAttributes(t *testing.T) {
...
 }
+
+// Per https://github.com/uber/tchannel-go/issues/505, concurrent client calls
+// made with the same shared map used as headers were causing panic due to
+// concurrent writes to the map when injecting tracing headers.
+func TestReusableHeaders(t *testing.T) {
+	tracer := mocktracer.New()
+
+	opts := &testutils.ChannelOpts{
+		ChannelOptions: ChannelOptions{Tracer: tracer},
```
prashantv: can we inline the `mocktracer.New` in here?

prashantv:
`require.NoError` inside of a goroutine is a bad idea, but `assert.NoError` is fine, so let's switch to that.

tracing_test.go
yurishkuro: I'll add validation, otherwise we don't know if tracing is even propagated / headers being modified

tracing.go
yurishkuro: added comment

prashantv:
lgtm, thanks for the quick fix @yurishkuro

PR #507, opened 2016-09-29
Remove call to Github API to get current branch name

breerly:
Now that Travis has shipped travis-ci/travis-ci#1633, we don't need to make a call to the Github API to get the current branch name from both PUSH and PR builds.


prashantv:
Nice! lgtm

PR #511, opened 2016-10-04
relay: Increase max tombstones and TTL

prashantv:
If we receive a response frame after the TTL period, we see an info
log "received frame [...] for outbound message that no longer exists".

We're seeing this often in production, so increase the TTL and allow
storing more tombstones in memory which should reduce the likelihood
of hitting this issue.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -33,10 +33,10 @@ import (
 
 // _maxRelayTombs is the maximum number of tombs we'll accumulate in a single
 // relayItems.
-const _maxRelayTombs = 1e4
+const _maxRelayTombs = 3e4
 
 // _relayTombTTL is the length of time we'll keep a tomb before GC'ing it.
-const _relayTombTTL = time.Second
+const _relayTombTTL = 3 * time.Second
 
 var (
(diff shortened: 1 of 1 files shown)


blampe:
LGTM. Why do you think we're seeing these frequently?

prashantv:
The events that would cause this are:
 - Receive a call request, forward it on to some remote service
 - Call's timeout is hit so we Entomb the relay (we still keep track of it, but we know it's timed out)
 - Call's entomb TTL is hit, so we remove the relay completely. At this point, we no longer know anything about this call ID.
 - The remote service responds, we get a `callRes` for an ID we no longer recognize. This causes the log message from earlier.

So I think we're seeing this because the remote service has not implemented timeouts, and is busy, so they're taking > 1 second to respond in extreme cases. A lot of these calls seem to be to a Java process, so I checked tchannel-java, and it doesn't seem like it implements timeouts when it receives a call, only on the outbound side when making a call. In this case, the Java process would be receiving an inbound call, and if they don't implement timeouts, we'd expect a delayed response from them.

I'm hoping the higher tombstone TTL will reduce the logs although it won't fix the core issue.

PR #516, opened 2016-10-10
Set DiffServ (QoS) bit on Outbound Connections

m4ttr:
The QoS option is configurable using DiffServ names AF11, etc.
DiffServ names correspond with rfc4594 recommended values,
that are also implemented with iptables.
A ToS bit is Configurable using DefaultConnectionOptions.TosPriority
configuration parameter on create server.

The QoS feature was implmeented with a dual stack network in mind.
setsockopt will throw an error if the wrong address family flag
is called.
- Adding Configuration Parameter TosPriority to ConnectonOptions.
- dscpName to interger value in tos.go.
- Test dscp value mapping is compliant with RFC4594 recommened values.
- TestTosPriority in connection_test.go to verify the outgoing socket
  has the correct ToS Bit set.
- Helper function IsTosPriority to determine if bit matches


tos.go
```diff
@@ -0,0 +1,70 @@
...
+
+package tchannel
+
+import "errors"
+
+var tosFieldMap map[string]int
+var errNotRecognizedTosName = errors.New("Unknown Tos Name")
+
+func init() {
+	tosFieldMap = map[string]int{
```
prashantv: Should we create a custom type rather than a string? And that way we can set the bit directly? It might even live in a subpackage,
```go
package tos

type TOS int

const (
  DF ToS = 0x00
  CS0 ToS = 0x00
  CS1 ToS = 0x08
  [...]
)
```

and then the connection option becomes `TOSPriority tos.TOS`

tos_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Two return value shim
+func makeAsOne(a, b interface{}) []interface{} {
+	return []interface{}{a, b}
+}
+
+func TestAssuredFowardValues(t *testing.T) {
```
prashantv: typically tests in Go use test tables, so this would look like:
```go

tests := []struct{
  str string
  want int
}{
  {"AF11", 10},
  {"AF12", 12},
  [...]
}

for _, tt := range tests {
  got, err := GetTosField(tt.str)
  require.NoError(t, err, "Failed to get ToS field for %v", tt.str)
  assert.Equal(t, tt.want, got, "Unexpected ToS value for %v", tt.str)
}
```

connection_test.go
```diff
@@ -765,3 +765,33 @@ func TestConnectTimeout(t *testing.T) {
...
+	opts1 := testutils.NewOpts().SetServiceName("s1").SetTosPriority("LOWDELAY").NoRelay()
+	opts2 := testutils.NewOpts().SetServiceName("s2").SetTosPriority("LOWDELAY").NoRelay()
+	testutils.WithTestServer(t, opts1, func(ts *testutils.TestServer) {
+		ch2 := ts.NewServer(opts2)
+		hp2 := ch2.PeerInfo().HostPort
+		defer ch2.Close()
+
+		ts.Register(raw.Wrap(newTestHandler(t)), "echo")
+		ch2.Register(raw.Wrap(newTestHandler(t)), "echo")
+		outbound, err := ts.Server().BeginCall(ctx, hp2, "s2", "echo", nil)
```
prashantv: since the call is being made from s1 -> s2, I think s2 doesn't need the tos priority set right?

connection_test.go
```diff
@@ -765,3 +765,33 @@ func TestConnectTimeout(t *testing.T) {
...
+		raw.WriteArgs(outbound, []byte("arg2"), []byte("arg3"))
+		require.NoError(t, err)
+		_, outboundNetConn := OutboundConnection(outbound)
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func(call *OutboundCall) {
```
prashantv: you've already written the arguments after the `BeginCall`, don't think you need any of the synchronization + goroutine

connection_test.go
```diff
@@ -765,3 +765,33 @@ func TestConnectTimeout(t *testing.T) {
...
+	testutils.WithTestServer(t, opts1, func(ts *testutils.TestServer) {
+		ch2 := ts.NewServer(opts2)
+		hp2 := ch2.PeerInfo().HostPort
+		defer ch2.Close()
+
+		ts.Register(raw.Wrap(newTestHandler(t)), "echo")
+		ch2.Register(raw.Wrap(newTestHandler(t)), "echo")
+		outbound, err := ts.Server().BeginCall(ctx, hp2, "s2", "echo", nil)
+		raw.WriteArgs(outbound, []byte("arg2"), []byte("arg3"))
+		require.NoError(t, err)
```
prashantv: check for error immediately after the `BeginCall`

connection_test.go
```diff
@@ -765,3 +765,33 @@ func TestConnectTimeout(t *testing.T) {
...
+	opts2 := testutils.NewOpts().SetServiceName("s2").SetTosPriority("LOWDELAY").NoRelay()
+	testutils.WithTestServer(t, opts1, func(ts *testutils.TestServer) {
+		ch2 := ts.NewServer(opts2)
+		hp2 := ch2.PeerInfo().HostPort
+		defer ch2.Close()
+
+		ts.Register(raw.Wrap(newTestHandler(t)), "echo")
+		ch2.Register(raw.Wrap(newTestHandler(t)), "echo")
+		outbound, err := ts.Server().BeginCall(ctx, hp2, "s2", "echo", nil)
+		raw.WriteArgs(outbound, []byte("arg2"), []byte("arg3"))
```
prashantv: don't ignore errors from `raw.WriteArgs`

connection_test.go
```diff
@@ -765,3 +765,33 @@ func TestConnectTimeout(t *testing.T) {
...
+		wg.Add(1)
+		go func(call *OutboundCall) {
+			defer wg.Done()
+			raw.WriteArgs(call, []byte("arg2"), []byte("arg3"))
+		}(outbound)
+		connTosPriority, err := IsTosPriority(outboundNetConn, "LOWDELAY")
+		require.NoError(t, err)
+		assert.Equal(t, connTosPriority, true)
+		wg.Wait()
+	})
```
prashantv: suggest completing the call (raw.ReadResponse I think) so that when the channel is closed, it doesn't wait for a timeout

tos.go
m4ttr: This was how I originally wrote the package, (used uint8, due to the size of the IP header) I ran into problem trying to lookup say, "CS0" by name.  If you have a go trick, let me know.  

Second, when I return the value, and I need it as a regular int to pass to .SetTOS(), without creating a custom type I dont have to convert.

prashantv: A custom int8 is perfect. You wouldn't need to look it up by name, since you'd expect the user to pass in the `tos.TOS` type. You can then "convert" to int, by doing `int(tos)`, since the underlying type is an integer. No conversions necessary

connection_test.go
m4ttr: yea it doesn't I should have removed this sooner

connection_test.go
m4ttr: Out of mostly being naive, the tested passed once I called write arg in the goroutine.

connection_test.go
m4ttr: I see only ReadArgs but doesn't appear to close the connection

tos/tos.go
```diff
@@ -0,0 +1,84 @@
...
+	// AF41 Assured Forward 33
+	AF41 ToS = 0x22
+	// AF42 Assured Forward 41
+	AF42 ToS = 0x24
+	// AF43 Assured Forward 42
+	AF43 ToS = 0x26
+	// EF Expedited Forwarding (RFC 3246)
+	EF ToS = 0x2e
+	// LOWDELAY 10
+	LOWDELAY ToS = 0x10
```
prashantv: In Go, constants typically aren't capitalized. Acronyms still should be, so all the above are fine, but these should just be `tos.LowDelay` `tps.Throughput` etc

connection.go
```diff
@@ -127,6 +130,11 @@ type ConnectionOptions struct {
 	// A static set of prefixes to check against the remote process name when
 	// setting up a connection. Check matches with RemoteProcessPrefixMatches().
 	CheckedProcessPrefixes []string
+
+	// QoS class name to be marked on outbound connection. Supply a DiffServ
+	// Compliant name, AF11 .. etc,  or "LOWDELAY", "THROUGHPUT", "RELIABILTIY"
```
prashantv: rather than quoting the `"LOWDELAY"` and `"THROUGHPUT"`, I think we should just mention `tos.LowDelay` and `tos.Throughput`

connection.go
```diff
@@ -243,6 +251,25 @@ func (ch *Channel) newOutboundConnection(timeout time.Duration, hostPort string,
 		return nil, err
 	}
 
+	opts := &ch.connectionOptions
+	tosPriority := opts.TosPriority
```
prashantv: We can simplify, as well as inline the scope of the variables here:
```go
if tosPriority := opts.TosPriority; tosPriority > 0 {
  [...]
}
```

connection.go
```diff
@@ -243,6 +251,25 @@ func (ch *Channel) newOutboundConnection(timeout time.Duration, hostPort string,
...
+
+	if tosPriority > 0 {
+		// Verify we are using a TCP socket
```
prashantv: Can we move this into a separate function that takes the connection + `tosPriority`?

connection_test.go
```diff
@@ -765,3 +766,35 @@ func TestConnectTimeout(t *testing.T) {
 		close(testComplete)
 	})
 }
+
+func TestTosPriority(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	opts1 := testutils.NewOpts().SetServiceName("s1").SetTosPriority(tos.LOWDELAY).NoRelay()
+	opts2 := testutils.NewOpts().SetServiceName("s2").NoRelay()
```
prashantv: `NoRelay` should only be required for options passed to `WithTestServer`, can remove it here

connection_test.go
```diff
@@ -765,3 +766,35 @@ func TestConnectTimeout(t *testing.T) {
 		close(testComplete)
 	})
 }
+
+func TestTosPriority(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	opts1 := testutils.NewOpts().SetServiceName("s1").SetTosPriority(tos.LOWDELAY).NoRelay()
```
prashantv: Couple of things:
 - you don't call "s1" so don't need to set the name
 - you don't actually need 2 listening servers here, only one's being called
 - we should add a comment for why relay is disabled here
 - no need to close, since servers created through the `TestServer` will automatically be closed
 - add context if it makes sense to assert/require.NoError
 - why does the WriteArgs need to happen in a goroutine?

I think this can actually be simplified to:
```go
// Relay is disabled because [....]
opts := testutils.NewOpts().SetServiceName("s1").SetTosPriority(..).NoRelay()
testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
  ts.Register(raw.Wrap(...))

  client := ts.NewClient(tnil)
  outbound, err := client.BeginCall(ctx, ts.HostPort(), "s1", "echo", nil)
  require.NoError(t, err, "BeginCall failed")

  _, outboundNetConn := OutboundConnection(outbound)
  // verify conn TOS here.

  // Complete the call
  raw.WriteArgs(....)
})
```

m4ttr: Some of these questions are hard for me to answer, while running the tests with and without my TestTosPriority sometimes other tests fail.  It become hard to determine signal from noise when reading the other failed tests.  To get something working i've copied from other tests.  Mostly naive on my part.   This feedback here helps.  I send a new diff shortly

tos/tos_jsonenums.go
```diff
@@ -0,0 +1,116 @@
...
+			interface{}(Lowdelay).(fmt.Stringer).String():    Lowdelay,
+			interface{}(Throughput).(fmt.Stringer).String():  Throughput,
+			interface{}(Reliability).(fmt.Stringer).String(): Reliability,
+			interface{}(Lowcost).(fmt.Stringer).String():     Lowcost,
+		}
+	}
+}
+
+// MarshalJSON is generated so ToS satisfies json.Marshaler.
+func (r ToS) MarshalJSON() ([]byte, error) {
```
prashantv: Suggest using `MarshalText` and `UnmarshalText` which works with both JSON and YAML
https://golang.org/pkg/encoding/#TextMarshaler

tos/tos_jsonenums.go
```diff
@@ -0,0 +1,116 @@
...
+		Throughput:  "Throughput",
+		Reliability: "Reliability",
+		Lowcost:     "Lowcost",
+	}
+)
+
+func init() {
+	var v ToS
+	if _, ok := interface{}(v).(fmt.Stringer); ok {
+		_ToSNameToValue = map[string]ToS{
```
prashantv: I assume we want to build up a map from string -> ToS here rather than hard coding it? If so, I think all you really want is:
```go
func init() {
  _tosNameToValue = make(map[string]ToS, len(_tosValueToName))
  for tos, tosString := range _tosValueToName {
    _tosNameToValue[tosString] = tos
  }
}
```

tos/tos_jsonenums.go
```diff
@@ -0,0 +1,116 @@
...
+
+package tos
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+var (
+	_ToSNameToValue = map[string]ToS{
```
prashantv: nit, suggest calling this `_tosNameToValue` and below `_tosValueToName` (the caps + underscore is a little confusing as to whether it's exported or not)

tos/tos_jsonenums.go
```diff
@@ -0,0 +1,116 @@
...
+			interface{}(Throughput).(fmt.Stringer).String():  Throughput,
+			interface{}(Reliability).(fmt.Stringer).String(): Reliability,
+			interface{}(Lowcost).(fmt.Stringer).String():     Lowcost,
+		}
+	}
+}
+
+// MarshalJSON is generated so ToS satisfies json.Marshaler.
+func (r ToS) MarshalJSON() ([]byte, error) {
+	if s, ok := interface{}(r).(fmt.Stringer); ok {
```
prashantv: why do you need this case? You know statically whether the `ToS` has a string method or not. You can just make this
```go
func (r ToS) MarshalText() ([]byte, error) {
  return json.Marshal(r.String()), nil
}
```

Since we know the value range is going to be a valid string.

utils_for_test.go
```diff
@@ -27,7 +27,10 @@ import (
 	"net"
 	"time"
 
+	"github.com/uber/tchannel-go/tos"
```
prashantv: nit: blank line after this to separate internal imports vs external imports

utils_for_test.go
```diff
@@ -72,6 +75,20 @@ func OutboundConnection(call *OutboundCall) (*Connection, net.Conn) {
 	return conn, conn.conn
 }
 
+// Return true if the connection matches the passed DiffServ Name.
+func IsTosPriority(c net.Conn, tosPriority tos.ToS) (bool, error) {
+	var connTosPriority int
+	var err error
+
+	if c.RemoteAddr().(*net.TCPAddr).IP.To16() != nil && c.RemoteAddr().(*net.TCPAddr).IP.To4() == nil {
```
billf: you can use a complex statement here to help out, e.g

```go
func IsTosPriority(c net.Conn, tosPriority tos.ToS) (bool, error) {
  var tosResolver func () (int, error)
  switch ip := c.RemoteAddr().(*net.TCPAddr).IP; ip {
  case ip.To16() != nil && ip.To4() == nil:
    tosResolver = ipv6.NewConn(c).TrafficClass
  case ip.To4() != nil:
    tosResolver = ipv4.NewConn(c).TOS
  default:
    return tosPriority == 0, nil
  }
  connTosPriority, err := tosResolver()
  return (connTosPriority == int(tosPriority)), err
}
```

excuse any syntax errors, i wrote this in the comment editor.

tos/tos.go
```diff
@@ -0,0 +1,77 @@
...
+	EF ToS = 0x2e
+	// Lowdelay 10
+	Lowdelay ToS = 0x10
+	// Throughput 8
+	Throughput ToS = 0x08
+	// Reliability  4
+	Reliability ToS = 0x04
+	// Lowcost 2
+	Lowcost ToS = 0x02
+)
```
billf: should there be a constant for `None ToS = 0x0` or `Unmarked ToS = 0x0` for the purposes of clearing or checking if a connection is unmarked

connection_test.go
```diff
@@ -765,3 +767,42 @@ func TestConnectTimeout(t *testing.T) {
...
+
+	var TosBit tos.ToS
+	err := TosBit.UnmarshalText([]byte("Lowdelay"))
+	if assert.NoError(t, err, "JSON unmarshal failed") {
+		assert.Equal(t, expected, TosBit, "JSON config mismatch")
+	}
+}
+
+func TestMarshalTosPriority(t *testing.T) {
+	expected := "\"Lowdelay\""
```
billf: can avoid the escaping (and i prefer to make these const in a simple test like this)

```go
const expected =`"Lowdelay"`
```

connection_test.go
```diff
@@ -765,3 +767,42 @@ func TestConnectTimeout(t *testing.T) {
...
+
+	v, err := json.Marshal(tos.Lowdelay)
+	if assert.NoError(t, err, "Marshal failed") {
```
billf: as the error is truly unexpected:

```go
require.NoError(t, err, "Marshal failed")
assert.Equal(t, expected, string(v), "Marshal mismatch")
```

tos/tos.go
m4ttr: I thought of this, but in go declared and undefiend bits are 0 anyways.

utils_for_test.go
m4ttr: Thanks for this, pattern is much cleaner.

schallert:
LGTM

tos/tos_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestMarshal(t *testing.T) {
+	for tos, wantMarshalled := range _tosValueToName {
+		marshalled, err := tos.MarshalText()
+		require.NoError(t, err, "Failed to marshal %v", tos)
+		assert.Equal(t, wantMarshalled, string(marshalled))
+
+		var got ToS
+		err = got.UnmarshalText(marshalled)
+		require.NoError(t, err, "Failed to unmarhsal %v", string(marshalled))
```
billf: typo `unmarhsal`

tos/tos_string.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// MarshalText implements TextMarshaler from encoding
+func (r ToS) MarshalText() ([]byte, error) {
+	return []byte(_tosValueToName[r]), nil
+}
+
+// UnmarshalText implements TextUnMarshaler from encoding
+func (r *ToS) UnmarshalText(data []byte) error {
+	v, ok := _tosNameToValue[string(data)]
```
billf: nit: can be shrunk down by two lines

```go
if v, ok := _tosNameToValue[string(data)]; ok {
  *r = v
  return nil 
}
return fmt.Errorf("invalid ToS %q", string(data))
```

prashantv: 👍

tos/tos_test.go
prashantv: 👍

PR #523, opened 2016-10-18
relay: Improve error when handling calls on inactive connections

prashantv:
Currently, we use a "channel-closed" message which is not correct since
the channel may be active, but the connection could be closing (e.g.,
because it hit a network error).

Return an error frame to the remote side so they see a declined rather than
a timeout.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -261,8 +261,11 @@ func (r *Relayer) Receive(f *Frame, fType frameType) {
 }
 
 func (r *Relayer) canHandleNewCall() (bool, connectionState) {
-	var canHandle bool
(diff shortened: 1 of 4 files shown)


relay.go
```diff
@@ -260,16 +260,18 @@ func (r *Relayer) Receive(f *Frame, fType frameType) {
 	}
 }
 
-func (r *Relayer) canHandleNewCall() bool {
+func (r *Relayer) canHandleNewCall() (bool, connectionState) {
 	var canHandle bool
+	var curState connectionState
```
abhinav: nit: var group

PR #524, opened 2016-10-19
relay: Add RateLimitDropError to drop traffic instead of relaying it

eklitzke:
This adds a new error that can be used to force the relay to drop packets. This should cause the client to timeout.

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -99,3 +99,15 @@ type Peer struct {
 	// to, which is also useful for stats.
 	Zone string
 }
+
+// RateLimitDropError is the error that should be returned from
+// RelayHosts.Get if the request should be dropped silently.
+// This is bit of a hack, because rate limiting of this nature isn't part of
+// the actual TChannel protocol.
+// The relayer will record that it has dropped the packet, but *won't* notify
+// the client.
+type RateLimitDropError struct{}
+
+func (e RateLimitDropError) Error() string {
+	return "frame dropped silently due to rate limiting"
+}
--- a/relay_test.go
+++ b/relay_test.go
@@ -541,7 +541,7 @@ func TestRelayRejectsDuringClose(t *testing.T) {
 	})
 }
 
-func TestSilentDrop(t *testing.T) {
+func TestRelayRateLimitDrop(t *testing.T) {
 	getHost := func(call relay.CallFrame, conn relay.Conn) (relay.Peer, error) {
 		return relay.Peer{}, relay.RateLimitDropError{}
 	}
@@ -549,11 +549,10 @@ func TestSilentDrop(t *testing.T) {
 	opts := testutils.NewOpts().
 		SetRelayOnly().
 		SetRelayHosts(hostsFunc(getHost))
-
 	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
-		gotCall := make(chan struct{})
+		var gotCall bool
 		testutils.RegisterEcho(ts.Server(), func() {
-			close(gotCall)
+			gotCall = true
 		})
 
 		client := ts.NewClient(nil)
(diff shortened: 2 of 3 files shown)


relay_test.go
```diff
@@ -490,6 +490,29 @@ func TestRelayConnection(t *testing.T) {
...
+	}
+
+	// Note: we cannot use WithTestServer since we override the RelayHosts.
+	opts := testutils.NewOpts().
+		SetServiceName("relay").
+		SetRelayHosts(hostsFunc(getHost)).
+		SetProcessPrefixes("nod", "nodejs-hyperbahn", "", "hyperbahn")
+	relay := testutils.NewServer(t, opts)
+	defer relay.Close()
+
```
prashantv: can we create a server, register echo, and ensure that it doesn't actually receive a call?

relay/relay.go
```diff
@@ -99,3 +99,15 @@ type Peer struct {
...
 	Zone string
 }
+
+// RateLimitDropError is an error that a Relayer will get when it should drop
+// requests due to hitting a rate limit. This is kind of a hack, because rate
+// limiting of this nature isn't part of the actual TChannel protocol.
+//
+// When the Relayer gets an error like this when calling .Get() on an outbound
+// it will record that it has dropped the packet, but *won't* notify the client.
+type RateLimitDropError struct{}
```
prashantv: Should this be an interface?

relay_test.go
```diff
@@ -490,6 +490,29 @@ func TestRelayConnection(t *testing.T) {
...
+	// Create a client that is listening so we can set the expected host:port.
+	clientOpts := testutils.NewOpts().SetProcessName("nodejs-hyperbahn")
```
prashantv: don't need to set process prefixes or process names in this test, it's unrelated to silent drop

relay_test.go
```diff
@@ -490,6 +490,29 @@ func TestRelayConnection(t *testing.T) {
 	assert.Contains(t, err.Error(), errTest.Error(), "Unexpected error")
 }
 
+func TestSilentDrop(t *testing.T) {
```
prashantv: let's name this `TestRelayRateLimitDrop` (the tests are all in the same package, so the names should describe which component they're testing)

relay/relay.go
eklitzke: I'm not sure. What's the argument for making it an interface?

prashantv: A good discussion of errors here:
http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

An interface allows Muttley to have different errors (maybe there's different reasons for the error, one because of hitting some service specific limit, one because it hit a global Muttley limit), and they implement the same interface. tchannel doesn't need to know the different reasons or what the error message for each of those should be, but it just checks whether the returned error satisfied the interface

eklitzke: I don't see how that holds  here. All of the other errors that are used in this path are using the SystemError interface, but that appears to be only used for errors that correspond to TChannel protocol errors, i.e. things defined by the TChannel protocol. The TChannel protocol does not have a soft drop error.

eklitzke:
We already have a custom pool name that our relay uses for dropped stats.

prashantv:
Yes, but we are not verifying that TChannel is actually incrementing stats for the correct pool name when it drops the frame right now

relay/relay.go
prashantv: There are multiple reasons for dropping traffic:
 - We are over the pool limit and are dropping traffic from the table
 - We are over some global limit for Muttley and are dropping traffic
 - Similarly, based on runtime stats, we determine that we want to drop traffic

Having a single error means that if we wanted to add more errors in future, we're stuck breaking the public API. With an interface, Muttley could return different errors for the above cases, but still have them be dropped.

prashantv:
Note: a TChannel call consists of multiple calls, so it's possible that the call has a `callReqContinue` afterwards. If we drop the frame without recording anything, we'll get `errUnknownID` when handling the consequent frames:
https://github.com/uber/tchannel-go/blob/dev/relay.go#L208

This will currently cause a log when we go to `forwardPeerFrame`. We might need to fix that as well (can do that separately).

prashantv:
@eklitzke I updated the test, can you take a look and make sure the changes look fine to you?

Issue #529, opened 2016-11-01
relay: Cap TTL to some configurable max


prashantv:
Allow a configurable TTL when creating a relay that is the maximum TTL that muttley supports. Hyperbahn currently uses a 2 minute cap on TTL, we should have something similar.

All calls with a TTL higher than this value should be capped (Muttley will change the `ttl` field).

We may want to track when we capped a TTL in stats so we can reach out to service owners and have them reduce their timeouts.

akshayjshah:
Closed in #530.

PR #530, opened 2016-11-02
Clamp relay timeouts to a configurable max

akshayjshah:
To prevent accumulating an unbounded number of relay items, clamp timeouts to a configurable max. Default to a max of 2 minutes, which matches Hyperbahn's current behavior.

Addresses #529.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -46,7 +46,8 @@ var (
 )
 
 const (
-	ephemeralHostPort = "0.0.0.0:0"
+	ephemeralHostPort      = "0.0.0.0:0"
+	defaultRelayMaxTimeout = 2 * time.Minute
 )
 
 // ChannelOptions are used to control parameters on a create a TChannel
@@ -71,6 +72,10 @@ type ChannelOptions struct {
 	// The list of service names that should be handled locally by this channel.
 	// This is an unstable API - breaking changes are likely.
 	RelayLocalHandlers []string
+
+	// The maximum allowable timeout for relayed calls (longer timeouts are
+	// clamped to this value). Passing zero uses the default of 2m.
+	RelayMaxTimeout time.Duration
 
 	// The reporter to use for reporting stats for this channel.
 	StatsReporter StatsReporter
@@ -121,6 +126,7 @@ type Channel struct {
 	connectionOptions ConnectionOptions
 	peers             *PeerList
 	relayHosts        relay.Hosts
+	relayMaxTimeout   time.Duration
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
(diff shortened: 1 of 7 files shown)


relay.go
```diff
@@ -379,6 +382,9 @@ func (r *Relayer) handleCallReq(f lazyCallReq) error {
 	origID := f.Header.ID
 	destinationID := remoteConn.NextMessageID()
 	ttl := f.TTL()
+	if ttl > r.maxTimeout {
+		ttl = r.maxTimeout
```
prashantv: Do we also want to rewrite the ttl in the frame? So that the caller also receives the shorter timeout. If so, we should add a unit test for that case.

akshayjshah: Went back and forth about that - what does Hyperbahn do?

akshayjshah: Resolved offline - HB *does* rewrite the TTL in the in-flight frame. We'll do the same here.

akshayjshah: Fixed.

relay_messages.go
```diff
@@ -167,6 +172,19 @@ func (f lazyCallReq) TTL() time.Duration {
 	return time.Duration(ttl) * time.Millisecond
 }
 
+func (f lazyCallReq) SetTTL(d time.Duration) error {
+	if d < 0 {
+		return errTTLNegative
+	}
+	millis := d / time.Millisecond
+	if millis > math.MaxUint32 {
+		return errTTLOverflow
```
prashantv: these error cases can be caught at channel creation time since it points to a strange configuration error. Should we instead do this check once when creating the `Relayer`, log errors, and then store the `maxTimeTTL` which is either the user-configured value, or 2 mins if not configured or the value passed in is invalid.

Once we do that, this method doesn't need to return an error which simplifies some of the caller logic and removes logging in this path.

channel.go
```diff
@@ -191,11 +202,20 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		relayStats = opts.RelayStats
 	}
 
+	maxMillis := opts.RelayMaxTimeout / time.Millisecond
```
akshayjshah: Sounds good.

channel.go
```diff
@@ -191,11 +202,20 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
...
+	if opts.RelayMaxTimeout == 0 {
+		opts.RelayMaxTimeout = defaultRelayMaxTimeout
+	} else if opts.RelayMaxTimeout < 0 || maxMillis > math.MaxUint32 {
+		logger.WithFields(
+			LogField{"configuredMaxTimeout", opts.RelayMaxTimeout},
+			LogField{"defaultMaxTimeout", defaultRelayMaxTimeout},
+		).Warn("Configured RelayMaxTimeout is invalid, using default instead.")
+		opts.RelayMaxTimeout = defaultRelayMaxTimeout
```
akshayjshah: Considered it, but I really don't like storing durations as anything other than a `time.Duration`. Even in the hot path, it doesn't seem worth it to save 0.3 ns.

relay.go
```diff
@@ -579,3 +589,18 @@ func determinesCallSuccess(f *Frame) (succeeded bool, failMsg string) {
 		return false, ""
 	}
 }
+
+func validateRelayMaxTimeout(d time.Duration, logger Logger) time.Duration {
+	maxMillis := d / time.Millisecond
+	if maxMillis >= 1 && maxMillis <= math.MaxUint32 {
```
prashantv: same as `> 0` which might be easier to read

PR #533, opened 2016-11-09
[WIP] Add dispatcher abstraction w/ panic handling

nathanjordan:


prashantv:
This is a good start @nathanjordan 

Some things to think about:
 - should dispatcher be an option that allows user to override it?
 - We should probably log at least an error when this happens

PR #535, opened 2016-11-14
relay: Refactor hosts/stats interface

prashantv:
This refactor does not make any logical changes to the code -- although
it does allow for relay to now forward frames across channels by
returning a peer from a separate channel.

The new interface combines the peer selection and stats interface into a
single RelayCall interface that returns a *Peer rather than a
relay.Peer.

Some advantages of this approach:
 - Avoids a map lookup for converting host:port to *Peer
 - Allows returning peers from separate channels
   This allows our Hyperbahn emulation code to significantly reduce the
   number of channels we will need.
 - No more passing around Zone/Pool in tchannel which does not need to
   know about these concepts.


relay_api.go
```diff
@@ -0,0 +1,54 @@
...
+	Start(relay.CallFrame, relay.Conn) (RelayCall, error)
+}
+
+// RelayCall abstracts away peer selection, stats, and any other business
+// logic from the underlying relay implementation. A RelayCall may not
+// have a destination if there was an error during peer selection
+// (which should be returned from start).
+type RelayCall interface {
+	// Destination returns the selected peer (if there was no error from Start).
+	Destination() (peer *Peer, ok bool)
```
akshayjshah: Why do we need the additional boolean? Is there a situation where `peer == nil` but `ok == true`?

benchmark/real_relay.go
```diff
@@ -34,14 +36,14 @@ type fixedHosts struct {
 	pickI atomic.Int32
 }
 
-func (fh *fixedHosts) Get(call relay.CallFrame, _ relay.Conn) (relay.Peer, error) {
-	peers := fh.hosts[string(call.Service())]
+func (fh *fixedHosts) Get(cf relay.CallFrame, conn relay.Conn) (string, error) {
```
prashantv: ✅

connection.go
```diff
@@ -309,7 +309,7 @@ func (ch *Channel) newConnection(conn net.Conn, outboundHP string, initialState
 	c.inbound.onAdded = c.onExchangeAdded
 	c.outbound.onAdded = c.onExchangeAdded
 
-	if ch.relayHosts != nil {
+	if ch.relayHost != nil {
```
prashantv: ✅

relay/noop.go
```diff
@@ -19,21 +19,3 @@
...
-func (n noopStats) Begin(_ CallFrame) CallStats {
-	return noopCallStats{}
-}
-
-type noopCallStats struct{}
-
-func (n noopCallStats) Succeeded()      {}
-func (n noopCallStats) Failed(_ string) {}
-func (n noopCallStats) SetPeer(_ Peer)  {}
-func (n noopCallStats) End()            {}
```
prashantv: ✅ 

wasn't sure whether i'd need noop implementations, but we can add them later if required.

relay_api.go
prashantv: Nope, if `ok`, then `peer != nil`. Just added this to make it more explicit that `peer` can be `nil`.

PR #539, opened 2016-11-23
Always call the Thrift post response callback if the ctx fn is called

robskillington:
cc @prashantv @xichen2020 @martin-mao @cw9

This fixes several small memory leaks for M3DB where the post response callback was not getting called and hence we weren't marking resources as able to be closed and flushed.

### DIFF
--- a/thrift/server.go
+++ b/thrift/server.go
@@ -134,9 +134,7 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
 	thriftProtocolPool.Put(wp)
 
 	if handler.postResponseCB != nil {
-		defer func() {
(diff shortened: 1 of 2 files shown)


thrift/server.go
```diff
@@ -133,6 +133,12 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
 	success, resp, err := handler.server.Handle(ctx, method, wp.protocol)
 	thriftProtocolPool.Put(wp)
 
+	if handler.postResponseCB != nil {
+		defer func() {
```
prashantv: nit: can simplify to
```go
defer handler.postResponseCB(ctx, method, resp)
```

PR #541, opened 2016-11-29
Use net.DialContext on go1.7 when creating outbound connections

### LINKED ISSUE #540: Allow cancellation during TCP connection dial
This seems to be related but not duplicate of #124.

`Connection.newOutboundConnection(...)` should use `net.DialContext` (instead of `net.DialTimeout`) to enable passing a context to the `net.Dialer` and allow cancellation while establishing a new TCP connection. This currently is prone to causing goroutine leaks if the calling goroutine exits before the dial operation returns.

https://github.com/uber/tchannel-go/blob/dev/connection.go#L228

dorianperkins:
Fixes #540

### DIFF
--- a/dial_16.go
+++ b/dial_16.go
@@ -23,8 +23,9 @@
 package tchannel
 
 import (
-	"context"
 	"net"
+
+	"golang.org/x/net/context"
 )
 
 func dialContext(ctx context.Context, hostPort string) (net.Conn, error) {
--- a/dial_16.go
+++ b/dial_16.go
@@ -1,0 +1,33 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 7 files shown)


dial_16.go
```diff
@@ -0,0 +1,33 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// +build !go1.7
```
breerly: Won't this also fire for 1.6 and 1.8?

prashantv: The way go's build tags work is that the "1.7" tag exists on 1.7+, so this will only happen on versions < 1.7. So 1.6 and lower will hit this (which is what we want), but 1.7 and higher will not

prashantv:
Tests are failing on go1.6 and lower, we should probably skip the test on anything under 1.7.

dial_17_test.go
```diff
@@ -0,0 +1,65 @@
...
+		return
+	}
+
+	if strings.Contains(err.Error(), "network is unreachable") {
+		t.Skipf("Skipping test, as network interface may not be available")
+	}
+
+	d := time.Since(started)
+	assert.Equal(t, ErrCodeCancelled, GetSystemErrorCode(err), "Ping expected to fail with timeout")
+	assert.True(t, d <= 2*timeoutPeriod, "Timeout should take less than %v, took %v", timeoutPeriod, d)
```
prashantv: nit: the check and message do not align:
 - we use `2 * timeoutPeriod` but the message just says `timeoutPeriod`
 - we do ` <= ` but message implies less.

We should probably do:
```go
assert.True(t, d < 2*timeoutPeriod, "Timeout should take less than %v, took %v", 2*timeoutPeriod, d)
```

dial_17_test.go
```diff
@@ -0,0 +1,65 @@
...
+	if !assert.Error(t, err, "Ping to blackhole address should fail") {
+		return
+	}
+
+	if strings.Contains(err.Error(), "network is unreachable") {
+		t.Skipf("Skipping test, as network interface may not be available")
+	}
+
+	d := time.Since(started)
+	assert.Equal(t, ErrCodeCancelled, GetSystemErrorCode(err), "Ping expected to fail with timeout")
```
prashantv: context error message is wrong

PR #542, opened 2016-11-30
relay: Don't attempt concurrent connections to the same peer

prashantv:
We avoid concurrent connections on peer.GetConnection, but the relay
path was missing the same checks.

Related PR: https://github.com/uber/tchannel-go/pull/500

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -719,3 +719,48 @@ func TestRelayThroughSeparateRelay(t *testing.T) {
 		assert.Equal(t, 1, numConns(introspected.RootPeers[serverHP]), "Expected 1 connection from relay2 to server")
 	})
 }
+
+func TestRelayConcurrentNewConnectionAttempts(t *testing.T) {
+	opts := testutils.NewOpts().SetRelayOnly()
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		// Create a server that is slow to accept connections by using
+		// a frame relay to slow down the initial message.
+		slowServer := testutils.NewServer(t, serviceNameOpts("slow-server"))
+		defer slowServer.Close()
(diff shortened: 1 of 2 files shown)


relay_test.go
```diff
@@ -718,3 +718,48 @@ func TestRelayThroughSeparateRelay(t *testing.T) {
...
+		// a frame relay to slow down the initial message.
+		slowServer := testutils.NewServer(t, serviceNameOpts("slow-server"))
+		defer slowServer.Close()
+		testutils.RegisterEcho(slowServer, nil)
+
+		var delayed bool
+		relayFunc := func(outgoing bool, f *Frame) *Frame {
+			if !delayed {
+				time.Sleep(testutils.Timeout(50 * time.Millisecond))
+				delayed = true
```
akshayjshah: The relay function runs in two concurrent goroutines - should `delayed` be atomic?

relay_test.go
```diff
@@ -718,3 +718,48 @@ func TestRelayThroughSeparateRelay(t *testing.T) {
...
+		defer close()
+		ts.RelayHost().Add("slow-server", slowHP)
+
+		// Make concurrent calls to trigger concurrent getConnectionRelay calls.
+		var wg sync.WaitGroup
+		for i := 0; i < 5; i++ {
+			wg.Add(1)
+			// Create client and get dest host:port in the main goroutine to avoid races.
+			client := ts.NewClient(nil)
+			relayHostPort := ts.HostPort()
```
akshayjshah: Do we need to check this on each loop iteration, or can we do it once outside the loop?

prashantv: Don't need to check this on each iteration, but since it's not expensive, I'd prefer it here so it's closer to use, and is next to the comment explaining why we don't just inline the call.

Issue #546, opened 2016-12-06
thrift.Wrap should not replace existing headers


prashantv:
Currently, calling `thrift.Wrap` will replace any existing headers.

If the user has something like:
```go
func f(ctx thrift.Context) {
  vctx := context.WithValue(ctx, "a", "b")
  ctx = thrift.Wrap(vctx)
}
```

the resulting `ctx` drops headers.

PR #547, opened 2016-12-06
Wrap(ctx) should not drop existing headers

prashantv:
Fixes https://github.com/uber/tchannel-go/issues/546


prashantv:
@billf @akshayjshah can you take a look at this?

PR #548, opened 2016-12-09
Clean up Channel lifecycle, add ID per channel

prashantv:
Add a unique channel ID, and have an onClosed method for when a channel
is actually closed. This is useful for:
 - Stopping related objects that are no longer needed on shutdown
 - Raising an event when a channel is closed


prashantv:
Good idea, I'll do that as a follow on, and get it ready based off this.

prashantv:
Actually turns out connection IDs are already in the introspection output, so I'll add channel IDs in this PR.

Issue #519, opened 2016-10-13
ContextWithHeaders implementation is not thread-safe


jacobgreenleaf:
The standard library context is thread-safe, but the structs that implement `ContextWithHeaders` are not because multiple responses will overwrite the internal map for response headers. This causes a data race or undefined behavior, especially confusing since v1.0.5 where a `ContextWithHeaders` can be passed as a `context.Context`. 

It's still a useful pattern to be able to re-use context values across goroutines, for example sharing common request headers and deadlines. 

The current solution for this would be for example:

```
var client thrift.TChanClient

ctx := context.Background()
tctx := thrift.WithHeaders(ctx, map[string]string{
    "User-Agent": "Googlebot",
})

go client.Call(thrift.WithHeaders(tctx, tctx.Headers()), ...)
go client.Call(thrift.WithHeaders(tctx, tctx.Headers()), ...)
```

Suggest adding a utility method for callers who don't care about the response headers

```
go client.Call(tctx.Clone(), ...)
go client.Call(tctx.Clone(), ...)
```

PR #549, opened 2016-12-09
Add ctx.Child method for easier parallel sub-request calls

prashantv:
Since our context is not goroutine safe (SetResponesHeaders mutates the
headers container), users need to create a copy when making sub-requests
in parallel. Child makes this much easier for users.

Fixes https://github.com/uber/tchannel-go/issues/519


prashantv:
Open to other method names too, `Clone` might be misinterpreted as a disconnected context, but this still retains the Parent->Child relationship. Maybe `SubContext` or `Child`?

prashantv:
Yeah I think I prefer `Child`, renamed.

PR #550, opened 2016-12-12
relay: Remove "remote process match" checks

prashantv:
Our internal relays have been refactored to no longer require the remote
process checking, so we can remove this very specific functionality.


prashantv:
@akshayjshah Can you take a look at this change -- just a little bit of cleanup.

PR #556, opened 2017-01-05
ListenIP: Prefer non-docker interfaces

### LINKED ISSUE #554: ListenIP should try to avoid the docker interface
On machines that have `docker0` and `eth0`, `ListenIP` is currently selecting the docker interface's IP address. From Go's perspective, the interface is up, non-local, and has an IPv4 address:
```
iface {Index:4 MTU:1500 Name:docker0 HardwareAddr:02:42:9c:52:fc:86 Flags:up|broadcast|multicast}
  addr 172.17.0.1/16
```

This leads to a service running on the host hardware selecting an IP that can't be reached from other machines. It seems like the Docker interfaces have a fixed prefix for MAC addresses:
https://docs.docker.com/v1.7/articles/networking/

```
The MAC address is generated using the IP address allocated to the container to avoid ARP collisions, using a range from 02:42:ac:11:00:00 to 02:42:ac:11:ff:ff.
```

So we could score MAC addresses that start with `02:42:ac:11` a little lower.

prashantv:
Fixes #554

### DIFF
--- a/localip.go
+++ b/localip.go
@@ -21,9 +21,15 @@
 package tchannel
 
 import (
+	"bytes"
 	"errors"
 	"net"
 )
+
+// Docker MAC addresses are 02:42:ac:11:00:00 to 02:42:ac:11:ff:ff.
+// See: https://docs.docker.com/v1.7/articles/networking/
+// We want to prefix match on the first 4 bytes.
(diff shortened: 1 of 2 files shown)


prashantv:
Yep, we tend to update the changelog when doing a release (not sure why we decided on that, but I think it was so we could summarize multiple PRs in one line if it made sense), so will update at that point.

PR #558, opened 2017-01-06
Rewrite thrift-gen test to work without removing vendor

prashantv:
thrift-gen tests did not work with the vendor directory, since the
TChannel code required generated code to satisfy an interface which used
types exposed via vendor, but the generated code was in a temporary
directory, so the types it implemented did not reference the same vendor
directory.

This change creates a fake GOPATH in a temporary directory, symlinks
TChannel and the vendor directory, and creates temporary directories
inside of the this TChannel directory so that TChannel and the generated
code share the same references to vendor.


thrift/thrift-gen/compile_test.go
```diff
@@ -30,14 +30,71 @@ import (
...
+	goPath, err := ioutil.TempDir("", "thrift-gen")
+	require.NoError(t, err, "TempDir failed")
+
+	// Create $GOPATH/src/github.com/uber/tchannel-go and symlink everything.
+	// And then create a dummy directory for all the test output.
+	tchannelDir := getTChannelDir(goPath)
+	require.NoError(t, os.MkdirAll(tchannelDir, 0755), "MkDirAll failed")
+
+	// Symlink everything in the real GOPATH.
+	realTChannelDir := getTChannelDir(os.Getenv("GOPATH"))
```
abhinav: What about when `GOPATH="$path1:$path2"`? Also, aren't tests run from inside
the directory containing the test file? Is it too hacky to walk up the
directory tree and find the tchannel-go directory?

thrift/thrift-gen/compile_test.go
```diff
@@ -30,14 +30,71 @@ import (
...
+	require.NoError(t, err, "Failed to read real tchannel-go dir")
+
+	for _, f := range realDirContents {
+		realPath := filepath.Join(realTChannelDir, f.Name())
+		err := os.Symlink(realPath, filepath.Join(tchannelDir, filepath.Base(f.Name())))
+		require.NoError(t, err, "Failed to symlink %v", f.Name())
+	}
+
+	_testGoPath = goPath
+	os.Setenv("GOPATH", goPath)
```
abhinav: Do you not need to restore the old GOPATH?

thrift/thrift-gen/compile_test.go
```diff
@@ -231,13 +288,10 @@ func runBuildTest(t *testing.T, thriftFile string) error {
...
 func runTest(t *testing.T, opts processOptions, extraChecks func(string) error) error {
-	tempDir, err := ioutil.TempDir("", "thrift-gen")
-	if err != nil {
-		return err
-	}
+	tempDir := getOutputDir(t)
 
 	// Generate code from the Thrift file.
-	*packagePrefix = "../"
+	*packagePrefix = strings.TrimPrefix(tempDir, _testGoPath+"/src/") + "/"
```
abhinav: Looks like package prefix will always be `github.com/uber/tchannel-go/$rand`
where `$rand` was generated by `getOutputDir`. Might be cleaner to just have
`getOutputDir` return the directory and the import path.

PR #564, opened 2017-01-18
Ensure reader is empty before calling Close

### LINKED ISSUE #563: Handle empty frames that return 0 bytes in thrift/JSON
The TChannel spec allows the following:
- Frame with full Thrift body
- Frame with empty bytes (8 byte `call res continue` with just "00" in the fragmentation payload).

This can cause issues when we're decoding responses, e.g.,
- We pass a reader to the Thrift decoder
- Thrift decoder consumes all required bytes
- We try to close the reader

At this point, the reader returns an error `errExpectedMoreArguments`, since the last TChannel frame has the `hasMoreFragments` bit set.

To avoid this error, instead of closing the reader, we should ensure that there's no bytes left explicitly by issuing a further `Read` call and expect `0, io.EOF`.

prashantv:
This triggers an extra Read which will avoid
https://github.com/uber/tchannel-go/issues/563

This is only an issue if the remote side is sending empty TChannel
frames, which should only happen during streaming.

Fixes #563 

cc @willhug @akshayjshah

### DIFF
--- a/arguments_test.go
+++ b/arguments_test.go
@@ -23,7 +23,10 @@ package tchannel
 import (
 	"bytes"
 	"io"
+	"strings"
 	"testing"
+
+	"io/ioutil"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -70,6 +73,15 @@ func TestJSONInputOutput(t *testing.T) {
 	assert.True(t, reader.closed)
 	assert.Equal(t, "Foo", outObj.Name)
 	assert.Equal(t, 20756, outObj.Value)
+}
+
+func TestReadNotEmpty(t *testing.T) {
+	// Note: The contents need to be larger than the default buffer size of bufio.NewReader.
+	r := bytes.NewReader([]byte("{}" + strings.Repeat("{}\n", 10000)))
+
+	var data map[string]interface{}
+	reader := NewArgReader(ioutil.NopCloser(r), nil)
+	require.Error(t, reader.ReadJSON(&data), "Read should fail due to extra bytes")
 }
 
 func BenchmarkArgReaderWriter(b *testing.B) {
--- a/arguments_test.go
+++ b/arguments_test.go
@@ -23,10 +23,9 @@ package tchannel
 import (
 	"bytes"
 	"io"
+	"io/ioutil"
 	"strings"
 	"testing"
-
-	"io/ioutil"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
--- a/internal/argreader/empty.go
+++ b/internal/argreader/empty.go
@@ -23,15 +23,25 @@ package argreader
 import (
 	"fmt"
 	"io"
+	"sync"
 )
+
+var _bufPool = sync.Pool{
+	New: func() interface{} {
+		b := make([]byte, 128)
+		return &b
+	},
+}
 
 // EnsureEmpty ensures that the specified reader is empty. If the reader is
 // not empty, it returns an error with the specified stage in the message.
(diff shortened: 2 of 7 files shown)


arguments_test.go
```diff
@@ -72,6 +75,15 @@ func TestJSONInputOutput(t *testing.T) {
 	assert.Equal(t, 20756, outObj.Value)
 }
 
+func TestReadNotEmpty(t *testing.T) {
+	// Note: The contents need to be larger than the default buffer size of bufio.NewReader.
+	r := bytes.NewReader([]byte("{}" + strings.Repeat("{}\n", 10000)))
```
willhug: so the "{}" is the 'end' of the thrift request right?  Which is why anything after the first '{}' causes an error?

internal/argreader/empty_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+package argreader
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uber/tchannel-go/testutils/testreader"
```
willhug: import ordering

internal/argreader/empty.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+// EnsureEmpty ensures that the specified reader is empty. If the reader is
+// not empty, it returns an error with the specified stage in the message.
+func EnsureEmpty(r io.Reader, stage string) error {
+	buf := make([]byte, 128)
```
abhinav: should this just be,

    var buff [128]byte
    n, err := r.Read(buff[:])

arguments_test.go
```diff
@@ -23,8 +23,11 @@ package tchannel
 import (
 	"bytes"
 	"io"
+	"strings"
 	"testing"
 
+	"io/ioutil"
```
prashantv: 👍

arguments_test.go
```diff
@@ -72,6 +75,15 @@ func TestJSONInputOutput(t *testing.T) {
 	assert.Equal(t, 20756, outObj.Value)
 }
 
+func TestReadNotEmpty(t *testing.T) {
+	// Note: The contents need to be larger than the default buffer size of bufio.NewReader.
```
prashantv: Nope, no constant we can use here.

arguments_test.go
prashantv: in this case it's using JSON, but yep.

internal/argreader/empty_test.go
prashantv: 👍

internal/argreader/empty.go
prashantv: It doesn't really matter, `buf` escapes either way.
```go
 31   var buf [128]byte
 32   n, err := r.Read(buf[:])
```

```
> go build  -gcflags=-m
# github.com/uber/tchannel-go/internal/argreader
./empty.go:32: buf escapes to heap
./empty.go:31: moved to heap: buf
```

Decided to instead pool the byte slices instead.

arguments_test.go
willhug: can we make a constant?

prashantv: I don't really think it adds much value to have a constant that's only used in a single test, and doesn't represent anything (e.g., it's counting the number of repeats for the string, not a number of bytes). It's pretty arbitrary.

willhug: ztkau2p

internal/argreader/empty.go
abhinav: 👍

PR #567, opened 2017-01-19
Clarify error messages in for fragmenting reader errors

### LINKED ISSUE #562: errExpectedMoreArguments clarification
https://github.com/uber/tchannel-go/blob/dev/fragmenting_reader.go#L38

The human msg is not really clear. Is it missing arguments? Or not enough arguments? Err variable itselfs says "Expected More Arguments". A simple renaming might really help. Debugging when you get this strange error msg is confusing.

prashantv:
Fixes #562

### DIFF
--- a/fragmenting_reader.go
+++ b/fragmenting_reader.go
@@ -34,8 +34,8 @@ var (
 	errChunkExceedsFragmentSize = errors.New("peer chunk size exceeds remaining data in fragment")
 	errAlreadyReadingArgument   = errors.New("already reading argument")
 	errNotReadingArgument       = errors.New("not reading argument")
-	errMoreDataInArgument       = errors.New("more data available in argument")
-	errExpectedMoreArguments    = errors.New("more arguments in message")
+	errMoreDataInArgument       = errors.New("closed argument reader when there is more data available to read")
+	errExpectedMoreArguments    = errors.New("closed argument reader when there may be more data available to read")
 	errNoMoreFragments          = errors.New("no more fragments")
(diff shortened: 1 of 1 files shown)


fragmenting_reader.go
```diff
@@ -34,8 +34,8 @@ var (
 	errChunkExceedsFragmentSize = errors.New("peer chunk size exceeds remaining data in fragment")
 	errAlreadyReadingArgument   = errors.New("already reading argument")
 	errNotReadingArgument       = errors.New("not reading argument")
-	errMoreDataInArgument       = errors.New("more data available in argument")
-	errExpectedMoreArguments    = errors.New("more arguments in message")
+	errMoreDataInArgument       = errors.New("closed argument reader when there is more data available to read")
+	errExpectedMoreArguments    = errors.New("closed argument reader when there may be more data available to read")
```
bombela: Why is this "may be more" and not "more data was expected"?

prashantv: The TChannel protocol allows frames with 0 bytes of argument data, which caused this bug:
https://github.com/uber/tchannel-go/issues/563

This was actually the issue that caused the error to be returned, but the user was seeing a correct result as well. The Thrift decoder completed reading all bytes, and called Close. However, `Close` sees that the TChannel stream is not yet complete, and returns the error. Trying to read more data causes you to get 0 bytes (so there was actually nothing to read), but it could have had more bytes. So I'm keeping the message as "may be more" since we don't really know at that point.

PR #568, opened 2017-01-19
Refactor connection options propagation

prashantv:
As we add more connection options, we keep duplicating them in
the connection struct. Instead, we can store the connection options
struct as a field in the connection.

This is some prefactoring for the active health checking work.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -39,6 +39,20 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	// CurrentProtocolVersion is the current version of the TChannel protocol
+	// supported by this stack
+	CurrentProtocolVersion = 0x02
+
+	// DefaultConnectTimeout is the default timeout used by net.Dial, if no timeout
+	// is specified in the context.
+	DefaultConnectTimeout = 5 * time.Second
+
+	// defaultConnectionBufferSize is the default size for the connection's
+	// read and write channels.
+	defaultConnectionBufferSize = 512
+)
+
 // PeerVersion contains version related information for a specific peer.
 // These values are extracted from the init headers.
 type PeerVersion struct {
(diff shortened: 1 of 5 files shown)


connection.go
```diff
@@ -309,13 +310,12 @@ func (ch *Channel) newConnection(conn net.Conn, outboundHP string, initialState
 
 		connID:          connID,
 		conn:            conn,
-		framePool:       framePool,
+		opts:            opts,
 		state:           initialState,
-		sendCh:          make(chan *Frame, sendBufferSize),
+		sendCh:          make(chan *Frame, opts.RecvBufferSize),
```
witriew: opts.SendBufferSize?

connection.go
```diff
@@ -238,6 +243,22 @@ func getTimeout(ctx context.Context) time.Duration {
...
+	if co.ChecksumType == ChecksumTypeNone {
+		co.ChecksumType = ChecksumTypeCrc32
+	}
+	if co.FramePool == nil {
+		co.FramePool = DefaultFramePool
+	}
+	if co.SendBufferSize <= 0 {
+		co.SendBufferSize = defaultConnectionBufferSize
+	}
+	if co.RecvBufferSize <= 0 {
```
witriew: Is this (RecvBufferSize) actually used anywhere?

connection.go
```diff
@@ -238,6 +243,22 @@ func getTimeout(ctx context.Context) time.Duration {
...
+		co.RecvBufferSize = defaultConnectionBufferSize
```
witriew: Is it actually useful to distinguish between Recv and Send buffer sizes?  Will they ever actually be different?

connection.go
prashantv: Good catch, marking as deprecated since we no longer use it.

connection.go
prashantv: Apparently the recv size is never used

connection.go
prashantv: 👍

connection.go
```diff
@@ -126,7 +132,7 @@ type ConnectionOptions struct {
 	// The frame pool, allowing better management of frame buffers. Defaults to using raw heap.
 	FramePool FramePool
 
-	// The size of receive channel buffers. Defaults to 512.
+	// NOTE: This is deprecated and not used for anything.
 	RecvBufferSize int
```
billf: would it be useful to change the type to reduce allocation size or is the concern about backwards compat?

prashantv: Yeah keeping it around for backwards compatibility.

PR #571, opened 2017-01-19
Publicize channel root peers

kriskowal:
This change reveals the root peer list on the channel’s public interface. This allows a library, say YARPC, to provide alternate peer list implementations and manage the lifecycle of retained peers without affecting unrelated TChannel proper peer lists.

Test coverage existed for RootPeers as the method was previously exposed as public but only for the purpose of tests. Clever!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 =========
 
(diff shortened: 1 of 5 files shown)


kriskowal:
This looks like a rebase-and-merge repository. c/d?

PR #572, opened 2017-01-21
Add OnPeerStatusChanged option to TChannel options

kriskowal:
Such that YARPC can monitor peer status changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+# (unreleased)
+
+* Add OnPeerStatusChanged option to NewChannel to receive notifications when
+  TChannel drops a connection and potentially other state change notifications.
+  Accepts a function and sends the affected peer.  The event handler function
+  must then call methods of the peer to inspect the new state.
+
 # v1.2.3
 
 * Improve error messages when an argument reader is closed without
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,9 @@ else
 endif
 
 install_glide:
-	GOPATH=$(OLD_GOPATH) go get -u github.com/Masterminds/glide
+	# all we want is: GOPATH=$(OLD_GOPATH) go get -u github.com/Masterminds/glide
+	# but have to pin to 0.12.3 due to https://github.com/Masterminds/glide/issues/745
+	GOPATH=$(OLD_GOPATH) go get -u github.com/Masterminds/glide && cd $(OLD_GOPATH)/src/github.com/Masterminds/glide && git checkout v0.12.3 && go install
 
 install_ci: install_glide install_lint get_thrift install
 	GOPATH=$(OLD_GOPATH) go get -u github.com/mattn/goveralls
--- a/channel.go
+++ b/channel.go
@@ -62,6 +62,10 @@ type ChannelOptions struct {
 	// unstable API - breaking changes are likely.
 	RelayHost RelayHost
 
+	// OnPeerStatusChanged receives notifications when a peer has become
+	// available or unavailable.
+	OnPeerStatusChanged func(*Peer)
+
 	// The list of service names that should be handled locally by this channel.
 	// This is an unstable API - breaking changes are likely.
 	RelayLocalHandlers []string
@@ -114,13 +118,14 @@ const (
 type Channel struct {
 	channelConnectionCommon
 
-	chID              uint32
-	createdStack      string
-	commonStatsTags   map[string]string
-	connectionOptions ConnectionOptions
-	peers             *PeerList
-	relayHost         RelayHost
-	relayMaxTimeout   time.Duration
+	chID                uint32
+	createdStack        string
+	commonStatsTags     map[string]string
+	connectionOptions   ConnectionOptions
+	peers               *PeerList
+	relayHost           RelayHost
+	relayMaxTimeout     time.Duration
+	onPeerStatusChanged func(*Peer)
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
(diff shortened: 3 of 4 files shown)


channel.go
```diff
@@ -62,6 +62,10 @@ type ChannelOptions struct {
 	// unstable API - breaking changes are likely.
 	RelayHost RelayHost
 
+	// OnPeerStatusChanged receives notifications when a peer has become
+	// available or unavailable.
+	OnPeerStatusChanged func(*Peer)
```
breerly: @kriskowal have you written a POC from yarpc's POV that uses this code? Might make sense to do that before landing, since making changes to this now is really cheap compared to after it's been released.

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+# (unreleased)
+
+* Add OnPeerStatusChanged option to NewChannel to receive notifications when
```
breerly: Add OnPeerStatusChanged option to Channel (not NewChannel, right?)

peer_test.go
```diff
@@ -183,20 +187,26 @@ func TestPeerRemoveClosedConnection(t *testing.T) {
 		_, outConns := p.NumConnections()
 		assert.Equal(t, 1, outConns, "Expected 1 remaining outgoing connection")
 
+		select {
+		case <-serverPeerClosed:
+		case <-time.After(time.Second):
+			t.Errorf("Timeout out waiting for server peer closed notification")
```
breerly: Perhaps consider a separate testcase instead of expanding this one.

channel.go
kriskowal: https://github.com/yarpc/yarpc-go/pull/705

Makefile
```diff
@@ -66,7 +66,9 @@ else
 endif
 
 install_glide:
-	GOPATH=$(OLD_GOPATH) go get -u github.com/Masterminds/glide
+	# all we want is: GOPATH=$(OLD_GOPATH) go get -u github.com/Masterminds/glide
+	# but have to pin to 0.12.3 due to https://github.com/Masterminds/glide/issues/745
+	GOPATH=$(OLD_GOPATH) go get -u github.com/Masterminds/glide && cd $(OLD_GOPATH)/src/github.com/Masterminds/glide && git checkout v0.12.3 && go install
```
prashantv: need to rebase?

channel.go
```diff
@@ -62,6 +62,10 @@ type ChannelOptions struct {
 	// unstable API - breaking changes are likely.
 	RelayHost RelayHost
 
+	// OnPeerStatusChanged receives notifications when a peer has become
+	// available or unavailable.
```
prashantv: I think we should expand the comment here with more details -- I this method should be called for more events than a peer becoming "available" or "unavailable". Basically any change to the state of the peer (new connections, closed connections) should trigger this right?

CHANGELOG.md
prashantv: I think the README shouldn't get into specifics about when the method is called or how it should be handled. That should be part of the API documentation. Ideally this should be 1-2 lines at most

channel.go
```diff
@@ -621,6 +628,9 @@ func (ch *Channel) getMinConnectionState() connectionState {
 func (ch *Channel) connectionCloseStateChange(c *Connection) {
 	ch.removeClosedConn(c)
 	if peer, ok := ch.rootPeers().Get(c.remotePeerInfo.HostPort); ok {
+		if ch.onPeerStatusChanged != nil {
+			ch.onPeerStatusChanged(peer)
```
prashantv: This is only called on `connectionCloseStateChange` -- we'll also want to deal with new incoming connections.

peer_test.go
```diff
@@ -183,20 +187,26 @@ func TestPeerRemoveClosedConnection(t *testing.T) {
...
+		}
+
 		c, err := p.GetConnection(ctx)
 		require.NoError(t, err, "GetConnection failed")
 		assert.Equal(t, c2, c, "Expected second active connection")
-	})
-}
 
-func TestPeerConnectCancelled(t *testing.T) {
```
prashantv: Why are we removing this test?

peer_test.go
```diff
@@ -183,20 +187,26 @@ func TestPeerRemoveClosedConnection(t *testing.T) {
 		_, outConns := p.NumConnections()
 		assert.Equal(t, 1, outConns, "Expected 1 remaining outgoing connection")
 
+		select {
+		case <-serverPeerClosed:
```
prashantv: Check the host:port of the returned peer

channel.go
prashantv: This should be the last thing we do, since the peer itself hasn't yet processed the `connectionCloseStateChange`, and so if you query the connections, it may still contain a closed connection.

We should have caught this in our unit test by having the test query the peer properties in the callback and asserting on the number of connections.

kriskowal:
I've figured out that this issue is not blocking *basic* peer chooser support in YARPC (the only kind of peer chooser we have). The current peer choosers will approach fully connected to all provided peers, and do nothing with peer availability notifications.

In the future, we will need better peer lists for peer to peer RPC or peer to mesh (e.g., Hyperbahn) so a peer list will attempt to maintain at most *n* peers from the provided peers.

In the Hyperbahn case, we have a large static list of all known router tasks and we are supposed to find one that is available, advertise to it with an ephemeral request, and maintain that outbound connection until we have at least *m* *preferred* inbound connections from Hyperbahn peers. This is a fancier peer list than what we have prototyped in YARPC today, and we may never need a Hyperbahn peer chooser in YARPC since we use a routing sidecar proxy now.

For the peer to peer case, we don't want every caller task to fully connect to every service task. Solving this problem might mean implementing something similar to partial affinity behind the peer provider, also bypassing the need for this complication.

prashantv:
Superceded by #624.

PR #573, opened 2017-01-23
Pin glide to version 0.12.3

kriskowal:
To unbreak dev branch.

Context https://github.com/Masterminds/glide/issues/745


breerly:
Aye, can confirm 0.12.3 is the one you want. Glide has been notoriously bad about breaking changes between minor, and even patch releases :(

PR #574, opened 2017-01-25
thrift-gen: Use namespace to generate the output filename

prashantv:
The namespace was not being used everywhere in #559 -- extend the test
to create services so that all the tchan-*.go files are generated and
we can ensure it's created in the right directory.

### DIFF
--- a/testutils/data.go
+++ b/testutils/data.go
@@ -37,8 +37,8 @@ var (
 
 func checkCacheSize(n int) {
 	// Start with a reasonably large cache.
-	if n < 8 {
-		n = 8
+	if n < 1024 {
+		n = 1024
 	}
 
 	randMut.RLock()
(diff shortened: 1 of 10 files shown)


testutils/data.go
```diff
@@ -37,8 +37,8 @@ var (
 
 func checkCacheSize(n int) {
 	// Start with a reasonably large cache.
-	if n < 8 {
-		n = 8
+	if n < 1024 {
```
prashantv: Since we cache a bunch of random bytes and reuse them, we were likely to return the same bytes when using something like `RandBytes(10)` where we have a small cache we select from. This caused thrift-gen tests to sometimes share the same directory name, making it harder to debug failures.

thrift/thrift-gen/main.go
```diff
@@ -180,14 +180,19 @@ func parseFile(inputFile string) (map[string]parseState, error) {
 	return allParsed, setExtends(allParsed)
 }
 
+func defaultPackageName(fullPath string) string {
+	_, filename := filepath.Split(fullPath)
```
abhinav: Why not https://golang.org/pkg/path/filepath/#Base

prashantv: 👍

PR #583, opened 2017-02-13
Update the peer scores when SetStrategy is called

### LINKED ISSUE #580: PeerList.SetStrategy should rescore peers
If `peers.Add(hostPort)` was already called, then `PeerList.SetStrategy(newStrat)` is subsequently called, the existing scores for the peers are wrong.

All the scores should be updated when `.SetStrategy()` is called.

alxn:
Fixes #580.

I took a heavy locking approach to this change, as I think that it's not safe to allow any access to any peer while the change is occurring, i.e. until all peers have been updated, the data is in an inconsistent state, and peers should not be compared to each other.

It's also worth noting that `.scoreCalculator` was not being protected, so there could have been a race with setting that, vs using it.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -83,7 +83,7 @@ func (l *PeerList) SetStrategy(sc ScoreCalculator) {
 
 	l.scoreCalculator = sc
 	for _, ps := range l.peersByHostPort {
-		l.updatePeerScore(ps, ps.score)
+		l.updatePeer(ps, ps.score)
 	}
 }
 
@@ -227,22 +227,22 @@ func (l *PeerList) exists(hostPort string) (*peerScore, uint64, bool) {
 	return ps, score, ok
 }
 
-// updatePeer is called when there is a change that may cause the peer's score to change.
+// onPeerChange is called when there is a change that may cause the peer's score to change.
 // The new score is calculated, and the peer heap is updated with the new score if the score changes.
-func (l *PeerList) updatePeer(p *Peer) {
+func (l *PeerList) onPeerChange(p *Peer) {
 	ps, psScore, ok := l.exists(p.hostPort)
 	if !ok {
 		return
 	}
 
 	l.Lock()
-	l.updatePeerScore(ps, psScore)
+	l.updatePeer(ps, psScore)
 	l.Unlock()
 }
 
-// Try to update the score of the peer given the existing score.
+// updatePeer is called to update the score of the peer given the existing score.
 // Note that a Write lock must be held to call this function.
-func (l *PeerList) updatePeerScore(ps *peerScore, psScore uint64) {
+func (l *PeerList) updatePeer(ps *peerScore, psScore uint64) {
 	newScore := l.scoreCalculator.GetScore(ps.Peer)
 	if newScore == psScore {
 		return
--- a/peer.go
+++ b/peer.go
@@ -76,9 +76,20 @@ func newPeerList(root *RootPeerList) *PeerList {
 	}
 }
 
-// SetStrategy sets customized peer selection stratedgy.
+// SetStrategy sets customized peer selection strategy.
 func (l *PeerList) SetStrategy(sc ScoreCalculator) {
+	l.Lock()
+	defer l.Unlock()
 	l.scoreCalculator = sc
+	for _, ps := range l.peersByHostPort {
+		newScore := l.scoreCalculator.GetScore(ps.Peer)
+		if newScore == ps.score {
+			continue
+		}
+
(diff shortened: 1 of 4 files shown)


alxn:
@prashantv fyi.

peer.go
```diff
@@ -76,9 +76,20 @@ func newPeerList(root *RootPeerList) *PeerList {
 	}
 }
 
-// SetStrategy sets customized peer selection stratedgy.
+// SetStrategy sets customized peer selection strategy.
 func (l *PeerList) SetStrategy(sc ScoreCalculator) {
+	l.Lock()
+	defer l.Unlock()
```
prashantv: nit: add a blank line after the `Unlock`

peer.go
```diff
@@ -76,9 +76,20 @@ func newPeerList(root *RootPeerList) *PeerList {
...
 	l.scoreCalculator = sc
```
prashantv: we use `l.scoreCalculator` in `updatePeer` without getting a lock. If we're protecting the write with a lock, we probably need to read the value under a lock too.

peer.go
```diff
@@ -76,9 +76,20 @@ func newPeerList(root *RootPeerList) *PeerList {
...
+	for _, ps := range l.peersByHostPort {
+		newScore := l.scoreCalculator.GetScore(ps.Peer)
+		if newScore == ps.score {
+			continue
+		}
+
+		ps.score = newScore
```
prashantv: since this is the same logic as in `updatePeer`, let's move this into a private function (Which sets the score, and updates the peer heap).

peer.go
alxn: Ah that's true, we could have probably just added:

```lang=go
        l.RLock()
        newScore := l.scoreCalculator.GetScore(p)
        l.RUnlock()
```

but... sharing the code will mean that we need to take an outerlock.

peer.go
alxn: I did think about this, but was frustrated by `updatePeer()` already being used!

peer.go
alxn: hmm, it's [there](https://github.com/uber/tchannel-go/pull/583/commits/aa93e59c2ced3b4ed79de3e5e0274e194891be84#diff-32e1ab53c69bf3272bd9e4b51b9bb105R83)

peer.go
prashantv: It currently gets an rlock to get the current score and to check whether the peer exists. Perhaps that same function should return the score calculator?

peer.go
prashantv: feel free to rename it! Maybe it can be called "onPeerChange" or something

peer.go
alxn: You mean `.exists()` should also return the score calculator?

Well that help with the common function, since that function assumes that an outer lock is held, since the lock is held for a *long* time in `SetStrategy()`.

alxn:
@prashantv Do you have any more comments over here.

@0legg do you want to take a look?

0legg:
@alxn I don't think I'm right person to ask for review =)

peer.go

peer.go
```diff
@@ -221,23 +227,29 @@ func (l *PeerList) exists(hostPort string) (*peerScore, uint64, bool) {
...
+// onPeerChange is called when there is a change that may cause the peer's score to change.
 // The new score is calculated, and the peer heap is updated with the new score if the score changes.
-func (l *PeerList) updatePeer(p *Peer) {
+func (l *PeerList) onPeerChange(p *Peer) {
 	ps, psScore, ok := l.exists(p.hostPort)
 	if !ok {
 		return
 	}
 
-	newScore := l.scoreCalculator.GetScore(p)
```
prashantv: This makes the behaviour of `onPeerChange` always get a write lock on the peer list, even if the score doesn't change. I don't think we want this behaviour.

Can we revert back to calculating the score, and only doing `updatePeer` if the score has changed.

Within `updatePeer`, it's worth checking the score as well (Since it's possible it changed between the read lock and the write lock).

alxn: @prashantv Remember that for the `SetStrategy()` case, I want to hold the write lock for the entire time, hence the reason that some lock must be held before calling `.updatePeer()`. I'll push a change.

alxn: @prashantv doing a double calculation nukes `TestPeerSelection()` ...:

```go
assert.EqualValues(t, 5, count.Load(),
	"Expect 5 exchange updates: peer add, init mex, new conn, ping, pong")
```

I don't want to change that to "9", which is what I got...

PR #590, opened 2017-02-23
Override inbound request handler

kriskowal:
The default inbound request handler delegates to the subchannel for a service name. YARPC needs the ability to dispatch or forward requests with arbitrary service names. To do so, it needs to bypass the subchannel router. This introduces an option to the Channel constructor to thread an alternate root handler.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -85,6 +85,10 @@ type ChannelOptions struct {
 	// Tracer is an OpenTracing Tracer used to manage distributed tracing spans.
 	// If not set, opentracing.GlobalTracer() is used.
 	Tracer opentracing.Tracer
+
+	// Handler is an alternate handler for all inbound requests, overriding the
+	// default handler that delegates to a subchannel.
+	Handler Handler
 }
 
 // ChannelState is the state of a channel.
@@ -126,6 +130,7 @@ type Channel struct {
 	peers             *PeerList
 	relayHost         RelayHost
 	relayMaxTimeout   time.Duration
+	handler           Handler
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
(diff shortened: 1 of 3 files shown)


subchannel_test.go
```diff
@@ -252,3 +253,29 @@ func TestGetSubchannelOptionsOnNew(t *testing.T) {
 	require.NoError(t, err, "Should get peer")
 	assert.Equal(t, want, peer, "Unexpected peer")
 }
+
+func TestHandlerWithoutSubChannel(t *testing.T) {
+	ch, err := NewChannel("svc", &ChannelOptions{
```
prashantv: Few suggests for this test:
 - use `testutils` to reduce the boilerplate in this test. You might want to add the new handler option to the `ChannelOptions` as well, so you can pass it in.
 - test the root channel name + a few subchannels
 - if you register an echo handler, you can use `testutils.AssertEcho`

```go
handler := HandlerFunc(...)
opts := testutils.NewOpts().[...]
testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
  client := ts.NewClient(nil)
  for _, svc := range []string{ts.ServiceName(), "misc", "other"} {
    testutils.AssertEcho(t, client, ts.HostPort(), svc)
  }
})
```

channel.go
```diff
@@ -217,6 +222,12 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	}
 	ch.peers = newRootPeerList(ch).newChild()
 
+	if opts.Handler != nil {
+		ch.handler = opts.Handler
```
prashantv: If a handler is replaced, should we disallow `Register` on the root channel (it should probably panic?)

PR #600, opened 2017-02-28
Improve connection logging

### LINKED ISSUE #599: Log tchannel version headers when a new connection is created
We currently log the remote process's name and hostports with all connection logs. On the initial "connection created" log, we should also add fields with the TChannel version:
https://github.com/uber/tchannel-go/blob/dev/connection.go#L254

prashantv:
Our health checks currently use a new TCP connection and we have a large
number of info logs for each health check (which is a very short lived
connection).

Improve the info logs to contain more information, and reduce the level
of less relevant logs to Debug.

Fixes #599

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -244,15 +244,23 @@ func (ch *Channel) newConnection(conn net.Conn, initialID uint32, outboundHP str
 	log := ch.log.WithFields(LogFields{
 		{"connID", connID},
 		{"localAddr", conn.LocalAddr()},
-		{"outboundHP", outboundHP},
 		{"remoteAddr", conn.RemoteAddr()},
 		{"remoteHostPort", remotePeer.HostPort},
 		{"remoteIsEphemeral", remotePeer.IsEphemeral},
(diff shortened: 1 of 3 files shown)


connection.go
```diff
@@ -512,7 +534,10 @@ func (c *Connection) protocolError(id uint32, err error) error {
 	sysErr := NewWrappedSystemError(ErrCodeProtocol, err)
 	c.SendSystemError(id, Span{}, sysErr)
 	// Don't close the connection until the error has been sent.
-	c.Close()
+	c.close(LogFields{
```
billf: is there a reason to create the `LogFields{}` container at the call site rather than having the compiler do whatever it feels like with:

```go
c.close(
  LogField{"reason", "protocol error"},
  ErrField(err),
)
```

syntactic sugar in case another `LogField{}` is created in-place?

prashantv: No reason, we should definitely just do what you suggested, updated.

PR #604, opened 2017-03-02
relay: Remove relay.Conn and use Connection directly

prashantv:
relay.Conn was used to expose relay-specific methods that we did not
want on the Connection (see #482). We were able to refactor some code
to remove this logic (#550).

We'd like to access more information about the connection a call started
on, so let's instead expose the Connection object.

### DIFF
--- a/relay/relaytest/func_host.go
+++ b/relay/relaytest/func_host.go
@@ -1,14 +1,14 @@
 package relaytest
 
 import (
-	tchannel "github.com/uber/tchannel-go"
+	"github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/relay"
 )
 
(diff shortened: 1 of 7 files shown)


relay/relaytest/func_host.go
```diff
@@ -18,15 +18,15 @@ type hostFuncPeer struct {
 }
 
 // HostFunc wraps a given function to implement tchannel.RelayHost.
-func HostFunc(fn func(cf relay.CallFrame, conn relay.Conn) (string, error)) tchannel.RelayHost {
+func HostFunc(fn func(cf relay.CallFrame, _ *tchannel.Connection) (string, error)) tchannel.RelayHost {
```
prashantv: 👍

PR #607, opened 2017-03-10
Added a peer.GetNew() to only return previously unselected peers.

alxn:
We pass in a `prevSelected` map, so this API will enforce that we honor it.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -45,6 +45,9 @@ var (
 	// ErrPeerNotFound indicates that the specified peer was not found.
 	ErrPeerNotFound = errors.New("peer not found")
 
+	// ErrNoNewPeers indicates that no previously unselected peer is available.
+	ErrNoNewPeers = errors.New("no new peer available")
+
 	peerRng = trand.NewSeeded()
 )
 
@@ -116,8 +119,8 @@ func (l *PeerList) Add(hostPort string) *Peer {
 	return p
 }
 
-// GetNew returns a new, previously un-selected peer from the peer list, or nil,
-// if no new un-selected peer can be found.
+// GetNew returns a new, previously unselected peer from the peer list, or nil,
+// if no new unselected peer can be found.
 func (l *PeerList) GetNew(prevSelected map[string]struct{}) (*Peer, error) {
 	l.Lock()
 	defer l.Unlock()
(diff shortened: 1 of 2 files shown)


peer.go
```diff
@@ -116,11 +116,12 @@ func (l *PeerList) Add(hostPort string) *Peer {
 	return p
 }
 
-// Get returns a peer from the peer list, or nil if none can be found.
-func (l *PeerList) Get(prevSelected map[string]struct{}) (*Peer, error) {
+// GetNew returns a new, previously un-selected peer from the peer list, or nil,
+// if no new un-selected peer can be found.
```
prashantv: instead of returning `nil, nil,` I think we should return `nil, ErrNoNewPeers` or something

alxn: @prashantv I'd wondered that, though I wasn't sure it was really an "error" because returning `nil` for the peer in that case is the expected behavior... Then the client has to do weirder error handling to actually "deal" with that `ErrNoNewPeers`, and then should they *still* check for `nil` ?

prashantv: We should guarantee that we either return a nil  peer with an error, or a non-nil peer with no error. That way the caller only checks the error, and not check the peer.

alxn: @prashantv So they'll have to do that awkward split error handling (like `EAGAIN` etc)

```go
peer, err := pl.GetNew(prev)
if err != nil {
  if err != ErrNoNewPeers {
    // log / wrap err
    return ...
  } else {
    // Do something different with the knowledge.
  }
}
// Use the peer without checking for nil.

```

alxn: @prashantv pushing from the 🚆

peer.go
```diff
@@ -131,9 +135,23 @@ func (l *PeerList) Get(prevSelected map[string]struct{}) (*Peer, error) {
...
+		return nil, ErrNoNewPeers
+	}
+	return peer, nil
+}
+
+// Get returns a peer from the peer list, or nil if none can be found,
+// will avoid previously selected peers if possible.
+func (l *PeerList) Get(prevSelected map[string]struct{}) (*Peer, error) {
+	peer, err := l.GetNew(prevSelected)
+	if err != nil && err != ErrNoNewPeers {
```
prashantv: `Get` should never return `nil, nil`
I think we want the logic to look like:
```go
if err == ErrNoNewPeers {
  peer = l.choosePeer(nil, false /* avoidHost */)
} else if err != nil {
  return nil, err
}

// This should only happen when we run `choosePeer`, but check outside
// of the if to be safe.
if peer == nil {
  return nil, ErrNoPeers
}

return peer, nil
```

alxn: @prashantv ah nice spot, it could race without the lock and the peer it would have returned got removed.

alxn:
@prashantv looking good over here.

Issue #608, opened 2017-03-16
Lock down version of Apache Thrift to either 0.9.3 or 0.10.0


peter-edge:
Apache Thrift changed the outputted data structure from `map[T]bool` to `map[T]struct{}` for `set<T>` from version 0.9.3 to 0.10.0. This repository does not lock down the version of thrift used in glide, and does not enforce the version used to generate code.

https://github.com/yarpc/yarpc-go/issues/783

peter-edge:
@prashantv The only guidance is `This version is based off apache/thrift@43fb34d with a couple of fixes`, we need a specific version of github.com/apache/thrift that should be added to users' glide.yaml that matches https://github.com/uber/tchannel-go/releases/tag/thrift-v1.0.0-dev, which version is this?

dcelasun:
Not sure if this issue is still relevant, but since Thrift 0.10.0, `set<T>` is [represented with a slice](https://github.com/apache/thrift/commit/43fb34df2871b69c2f34dc3bb353e65cbc9f8692) `[]T` (just like lists) and duplicate elements [return an error](https://github.com/apache/thrift/commit/43fb34df2871b69c2f34dc3bb353e65cbc9f8692#diff-36ef31e670e3df20e286373dfb7ecc23R3229) at runtime.

prashantv:
TChannel works either, and the `glide.yaml` specifies the following constraint:
```
- package: github.com/apache/thrift
  version: '>=0.9.3, <0.11.0'
```

I don't think TChannel needs to be pinned to a specific version, the APIs it uses are available in both, and so there's no need to make it more restrictive.

PR #609, opened 2017-03-17
Lock to Apache Thrift 0.10.0

peter-edge:
This solves https://github.com/uber/tchannel-go/issues/608.

So here's the situation:

There's two ways to solve the immediate issue, both of which have drawbacks:

**Option 0.9.3:** Retroactively lock tchannel-go to 0.9.3, use Docker for `make thrift_gen` in a container that has 0.9.3 installed, update Travis CI to use 0.9.3, update docs and enforce downstream consumers (including all consumers for the last few months) to use 0.9.3.

I started this option at https://github.com/uber/tchannel-go/compare/docker-thrift-gen. This has the benefit of not needing to update tchannel-go/thrift/thrift-gen to output `map[T]struct{}` instead of `map[T]bool`, making this technically the non-breaking change between the two options. However, installed versions of thrift on most machines (especially mac) will be 0.10.0 now, and users will have to figure out how to compile their thrift files using 0.9.3 themselves, or we have instructions on how to do this in a docker container. Given that neither tchannel-go or yarpc-go was doing this themselves, I think this is the more complicated option there, especially as the current flow for users to upgrade tchannel-go is to just upgrade this repo and expect everything to work. Right now, if you upgrade this repo and use thrift-gen, if 0.10.0 is installed (which is what will happen on 99% of machines), this will break.

**Option 0.10.0:** Lock tchannel-go to 0.10.0, continue to do `make thrift_gen` locally for now (although it would probably be better to use Docker anyways), update Travis CI to use 0.10.0, update tchannel-go/thrift/thrift-gen to output `map[T]struct{}` instead of `map[T]bool`.

This is what this PR does. This is a breaking change technically, which will make most people groan. However, existing users on the current version of tchannel-go are not affected until they regenerate their thrift files, which they're going to do at some point anyways (like when I tried to do this for yarpc-go, for example), and when they do regenerate their thrift files, they will see something break, update tchannel-go, and then nothing breaks again. If we go with option 0.9.3, they will not be able to fix the break unless they change how they do their thrift generation.

**Analysis**

Either option is breaking in my opinion, and after this, we need to do a better job of locking down thrift versions, and if we could, maintain statically-compiled thrift releases (if this was trivial, I'm sure someone would have done this, but just saying). Between the two options, as a consumer (in yarpc-go), it was much easier for me to deal with option 0.10.0. Let's fix this going forward, and (in my opinion) choose the less painful of two bad options.

**Post PR Tasks**

- Use Docker for `make thrift_gen` anyways
- Figure out how to do Apache Thrift releases (which I think is the idea behind https://github.com/prashantv/thrift/releases/tag/p0.0.1) and use them again in Travis
- I put together https://github.com/yarpc/yarpc-go/pull/785 that uses this branch right now, so get this in (or whatever we decide) afterwards
- Note that this also updated jaeger-client-go, which is being used at version 2 for most of Uber now. Locking this to version 1 makes yarpc-go hell at this point, but there is a crossdock test that fails with python because trace IDs changed from uint64s to strings, so after this PR, we need to get tchannel-python updated, then uncomment the test line I commented out.


peter-edge:
https://github.com/uber/tchannel-go/issues/608

peter-edge:
Note I put together https://github.com/yarpc/yarpc-go/pull/785 that uses this branch right now.

PR #615, opened 2017-03-23
Enforce apache thrift 0.9.3 in glide and in thrift generation

peter-edge:
This sets apache thrift to 0.9 in glide, and also makes sure we always use the released thrift in the tchannel-go releases for all generation.

https://github.com/uber/tchannel-go/issues/608

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,8 @@
-hash: 96546aa202f8964e8cc0c8acea5452cc9c6d79259ea4f9aaff03ffbf13c918a4
-updated: 2017-03-23T09:37:24.551738786+01:00
+hash: 111bdf9797d23d4b9f8833a9780e382dd53153af70ad286f99815cea4c2fdf4b
+updated: 2016-08-05T18:02:17.171302481-04:00
 imports:
 - name: github.com/apache/thrift
-  version: 53dd39833a08ce33582e5ff31fa18bb4735d6731
+  version: 0e9fed1e12ed066865e46c6903782b2ef95f4650
   subpackages:
   - lib/go/thrift
 - name: github.com/bmizerany/perks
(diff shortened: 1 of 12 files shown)


glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/uber/tchannel-go
 import:
 - package: github.com/apache/thrift
-  version: master
+  version: ~0.9
```
prashantv: should we do `^0.9.3`?

glide.lock
```diff
@@ -1,33 +1,36 @@
-hash: 111bdf9797d23d4b9f8833a9780e382dd53153af70ad286f99815cea4c2fdf4b
-updated: 2016-08-05T18:02:17.171302481-04:00
+hash: 96546aa202f8964e8cc0c8acea5452cc9c6d79259ea4f9aaff03ffbf13c918a4
```
prashantv: Can we do the glide change in a separate PR -- since it upgrades a lot of other deps too. Let's make this change just affect the thrift compiler

glide.yaml
breerly: No, `~0.9` keeps us in the 0.9 minor.

prashantv: I don't think we support `0.9.1` or `0.9.2` though in tchannel

peter-edge:
What is the problem exactly? Do you want to lock to a micro?

On Thu, Mar 23, 2017 at 4:46 PM, Prashant Varanasi <notifications@github.com

glide.yaml
peter-edge: Note I don't think apache is going to be doing any updating, like I can set this to 0.9.3 if we don't support 0.9.1 or 0.9.2, but I don't think there will be a 0.9.4.

breerly:
That depends: is the Thrift project releasing new functionality or breaking changes in patch releases (the z in x.y.z)?

If we can't trust them to do semver, then we might have to fully pin... that wouldn't be great tho.

peter-edge:
I reverted that lock from this PR entirely.

My tl;dr:

1. Locking to either ~-0.9 or 0.9.3 is strictly an improvement over what we
have now, so IMO we should merge and merge and merge
2. In practice, it probably won't matter, as I don't think there will be
another 0.9 release

PR #621, opened 2017-04-19
Make frame parsing more resilient to bad inputs

prashantv:
Currently, if we recieve a frame that is invalid (E.g., the size is
smaller than the header size which is the minimum), then we end up with
a huge payload size which causes an out-of-bounds panic trying to read
the payload.

Add full unit tests for parsing of the frame, as well as a quick test to
ensure TChannel will not panic on bad frames.

### DIFF
--- a/frame_test.go
+++ b/frame_test.go
@@ -22,10 +22,13 @@ package tchannel
 
 import (
 	"bytes"
+	"encoding/binary"
 	"encoding/json"
 	"io"
+	"math"
 	"testing"
 	"testing/iotest"
+	"testing/quick"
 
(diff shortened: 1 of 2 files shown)


frame_test.go
```diff
@@ -142,3 +145,139 @@ func TestMessageType(t *testing.T) {
...
+			decoded = false
+		}
+	}()
+	frame.ReadIn(bytes.NewReader(bs))
+	return true
+}
+
+func TestQuickFrameReadIn(t *testing.T) {
+	// Try to read any set of bytes as a frame.
+	err := quick.Check(frameReadIn, &quick.Config{MaxCount: 10000})
```
billf: Nice.

PR #624, opened 2017-05-22
Add hook for peer connection status changes

kriskowal:
This change adds a channel option OnPeerStatusChanged.  This is a callback that will receive a notification for peer connection status change events: whenever any connection opens or closes.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -57,7 +57,9 @@ type ChannelOptions struct {
 	// The name of the process, for logging and reporting to peers
 	ProcessName string
 
-	// OnPeerStatusChanged
+	// OnPeerStatusChanged is an optional callback that receives a notification
+	// whenever the channel establishes a usable connection to a peer, or loses
+	// a connection to a peer.
 	OnPeerStatusChanged func(*Peer)
 
 	// The logger to use for this channel
@@ -123,14 +125,15 @@ const (
 type Channel struct {
 	channelConnectionCommon
 
-	chID              uint32
-	createdStack      string
-	commonStatsTags   map[string]string
-	connectionOptions ConnectionOptions
-	peers             *PeerList
-	relayHost         RelayHost
-	relayMaxTimeout   time.Duration
-	handler           Handler
+	chID                uint32
+	createdStack        string
+	commonStatsTags     map[string]string
+	connectionOptions   ConnectionOptions
+	peers               *PeerList
+	relayHost           RelayHost
+	relayMaxTimeout     time.Duration
+	handler             Handler
+	onPeerStatusChanged func(*Peer)
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
@@ -220,7 +223,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		relayHost:         opts.RelayHost,
 		relayMaxTimeout:   validateRelayMaxTimeout(opts.RelayMaxTimeout, logger),
 	}
-	ch.peers = newRootPeerList(ch).newChild()
+	ch.peers = newRootPeerList(ch, opts.OnPeerStatusChanged).newChild()
 
 	if opts.Handler != nil {
 		ch.handler = opts.Handler
@@ -740,7 +743,7 @@ func (ch *Channel) State() ChannelState {
 // Close starts a graceful Close for the channel. This does not happen immediately:
 // 1. This call closes the Listener and starts closing connections.
 // 2. When all incoming connections are drained, the connection blocks new outgoing calls.
-// 3. When all connections are drainged, the channel's state is updated to Closed.
+// 3. When all connections are drained, the channel's state is updated to Closed.
 func (ch *Channel) Close() {
 	ch.Logger().Info("Channel.Close called.")
 	var connections []*Connection
--- a/connection_test.go
+++ b/connection_test.go
@@ -885,7 +885,7 @@ func TestPeerStatusChangeClient(t *testing.T) {
 		// Induce the creation of a connection from client to server.
 		client := ts.NewClient(copts)
 		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
-		assert.Equal(t, 1, <-changes, "first connection")
+		assert.Equal(t, 1, <-changes, "event for first connection")
 
 		// Re-use
 		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
@@ -897,15 +897,15 @@ func TestPeerStatusChangeClient(t *testing.T) {
 		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
 		defer cancel()
 		p.Connect(ctx)
-		assert.Equal(t, 2, <-changes, "second connection")
+		assert.Equal(t, 2, <-changes, "event for second connection")
 
 		// Induce the destruction of a connection from the server to the client.
 		server.Close()
-		assert.Equal(t, 1, <-changes, "first disconnection")
-		assert.Equal(t, 0, <-changes, "second disconnection")
+		assert.Equal(t, 1, <-changes, "event for first disconnection")
+		assert.Equal(t, 0, <-changes, "event for second disconnection")
 
 		client.Close()
-		assert.Equal(t, 0, len(changes), "unexpected peer status changes")
+		assert.Len(t, changes, 0, "unexpected peer status changes")
 	})
 }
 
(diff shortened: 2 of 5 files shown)


channel.go
```diff
@@ -57,7 +57,9 @@ type ChannelOptions struct {
 	// The name of the process, for logging and reporting to peers
 	ProcessName string
 
-	// OnPeerStatusChanged
+	// OnPeerStatusChanged is an optional callback that receives a notification
```
prashantv: Want to discuss whether we want to report other states:
 - peer that received a new connection (especially if it's from 0 -> 1)
 - peer that went from N -> N-1 where N > 2.

I think we want the former, but not the latter.

kriskowal: I’m hacking out the YARPC connection manager for TChannel and so far it looks like receiving a notification when the last connection dies is necessary and sufficient. Since I’m calling peer.Connect(), I have a direct line on when the connection count goes form 0 to 1.

The decrement case would be useful if we sought to ensure a pool of redundant connections, which I might in a later pass.

kriskowal:
I’ve taken a step back and amended this change such that the channel emits a notification for any gained or lost connection. This makes the code a little more obvious, with the dual added/removed notification call sites nearby in one file. It also supports the general case so we can go on to implement a >1 minimum concurrent connections manager in YARPC.

channel.go
```diff
@@ -57,7 +57,9 @@ type ChannelOptions struct {
...
+	// after a peer's status changes. Peer status changes include after the
+	// last connection to a peer closes and the peer is collected.
```
prashantv: Can we update this comment?

connection_test.go
```diff
@@ -868,3 +868,30 @@ func TestTosPriority(t *testing.T) {
...
+		server := ts.Server()
+		testutils.RegisterEcho(server, nil)
+		changes := make(chan int, 2)
+
+		copts := testutils.NewOpts().SetOnPeerStatusChanged(func(p *Peer) {
+			i, o := p.NumConnections()
+			changes <- i + o
+		})
+
+		// Induce the creation of a connection from client to server.
```
prashantv: can we ensure that we get a notification on the server for all clients by:
 - setting the `SetOnPeerStatusChanged` on the server
 - using a for loop with multiple clients, and ensuring that each new client triggers the event

```go
for i := 0; i < 5; i++ {
  client := ts.NEwClient(copts)
  testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
  assert.Equal(t, i, <-changes, "connected")

  // reuse connection
  testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
  assert.Len(t, 0, changes, "reusing connection should not trigger any events")
}
```
we should also verify that there are no queued events after we assert echo a second time.

we'll have to disable the delay as an option.

connection_test.go
```diff
@@ -868,3 +868,30 @@ func TestTosPriority(t *testing.T) {
...
+		client := ts.NewClient(copts)
+		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+		assert.Equal(t, 1, <-changes, "connected")
+
+		// Re-use
+		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+
+		// Induce the destruction of a connection from the server to the client.
+		server.Close()
```
prashantv: Unless explicitly specified, tests are run twice in TChannel -- one without a relay, and one with a relay. In this case, I think you want to run the test without a relay since you need to check logic for the number of connections.

Just closing the client should be enough to trigger changes once you disable the relay, you shouldn't have to close the server as well.

peer.go
```diff
@@ -318,7 +318,7 @@ func newPeerScore(p *Peer, score uint64) *peerScore {
 type Peer struct {
 	sync.RWMutex
 
-	channel             Connectable
```
prashantv: Rather than changing this reference to use `*Channel` and reaching into the `*Channel` from the peer, I think we should encapsulate the logic within the peer itself.

The `newPeer` constructor should take a `onPeerConnChange func(*Peer)`, and same for the root peer list. that way, it's not coupled to the channel, but uses a more DI style approach.

channel.go
```diff
@@ -740,7 +744,7 @@ func (ch *Channel) State() ChannelState {
 // Close starts a graceful Close for the channel. This does not happen immediately:
 // 1. This call closes the Listener and starts closing connections.
 // 2. When all incoming connections are drained, the connection blocks new outgoing calls.
-// 3. When all connections are drainged, the channel's state is updated to Closed.
+// 3. When all connections are drained, the channel's state is updated to Closed.
```
prashantv: thanks for fixing the typo!

peer.go
```diff
@@ -517,6 +522,10 @@ func (p *Peer) connectionCloseStateChange(changed *Connection) {
 
 	if found {
 		p.onClosedConnRemoved(p)
+		// Inform third parties that a peer lost a connection.
+		if p.channel.onPeerStatusChanged != nil {
```
prashantv: Optionally: once we do a DI approach, we may want to set a dummy noop function if this is `nil` so we can avoid the nil checks.

kriskowal: My intuition is that, if in the majority of cases, calling the function is avoidable, the nil check is cheaper. However, going forward on the assumption that most use of TChannel will be through YARPC, the nil check is overhead for every change. So, I concluded that using a noopStatusChanged is the way to go.

connection_test.go
```diff
@@ -868,3 +868,64 @@ func TestTosPriority(t *testing.T) {
...
+
+			// Open
+			testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+			assert.Equal(t, 1, <-changes, "opened")
+
+			// Re-use
+			testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+			assert.Equal(t, 0, len(changes), "re-used")
+
+			// Close
```
prashantv: can we move the `Close` tests to a separate loop, would like to see `NumConnections` go `> 1` and ensure we're still raising events. Right now, the tests are only testing the no connections <-> has connections case, rather than event on every connection.

connection_test.go
```diff
@@ -868,3 +868,64 @@ func TestTosPriority(t *testing.T) {
...
+		for i := 0; i < 5; i++ {
+			client := ts.NewClient(copts)
+
+			// Open
+			testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+			assert.Equal(t, 1, <-changes, "opened")
+
+			// Re-use
+			testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+			assert.Equal(t, 0, len(changes), "re-used")
```
prashantv: i think you can use `assert.Len(t, 0, changes, "no new events on reused connection")`

connection_test.go
```diff
@@ -868,3 +868,74 @@ func TestTosPriority(t *testing.T) {
...
+		assert.Equal(t, 1, <-changes, "event for first connection")
+
+		// Re-use
+		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+
+		// Induce the creation of a second connection from client to server.
+		pl := client.RootPeers()
+		p := pl.GetOrAdd(ts.HostPort())
+		ctx := context.Background()
+		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
```
prashantv: nit: no need to do `context.Background()` as a separate call, can inline it in the `WithTimeout` call

connection_test.go
```diff
@@ -868,3 +868,74 @@ func TestTosPriority(t *testing.T) {
...
+		client := ts.NewClient(copts)
+		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+		assert.Equal(t, 1, <-changes, "event for first connection")
+
+		// Re-use
+		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+
+		// Induce the creation of a second connection from client to server.
+		pl := client.RootPeers()
+		p := pl.GetOrAdd(ts.HostPort())
```
prashantv: if we don't need `pl` anywhere else, suggest inlining: `client.RootPeers().GetOrAdd(..)`

connection_test.go
```diff
@@ -868,3 +868,74 @@ func TestTosPriority(t *testing.T) {
 		require.NoError(t, err, "Failed to write to outbound conn")
 	})
 }
+
+func TestPeerStatusChangeClient(t *testing.T) {
+	sopts := testutils.NewOpts().NoRelay()
```
prashantv: can you add a comment above the `NoRelay` with the reason for why we don't enable the relay

connection_test.go
```diff
@@ -868,3 +868,74 @@ func TestTosPriority(t *testing.T) {
...
+		p := pl.GetOrAdd(ts.HostPort())
+		ctx := context.Background()
+		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
+		defer cancel()
+		p.Connect(ctx)
+		assert.Equal(t, 2, <-changes, "event for second connection")
+
+		// Induce the destruction of a connection from the server to the client.
+		server.Close()
+		assert.Equal(t, 1, <-changes, "event for first disconnection")
```
prashantv: i think this test might be flaky since it's possible that by the time the first even it raised, both connections have been closed, so `NumConnections` will be 0. can we run with `-count 1000` to check if it's flaky?

kriskowal: Totes flake. Moment.

PR #628, opened 2017-06-02
Back to -dev

kriskowal:


version.go
```diff
@@ -23,4 +23,4 @@ package tchannel
 // VersionInfo identifies the version of the TChannel library.
 // Due to lack of proper package management, this version string will
 // be maintained manually.
-const VersionInfo = "1.6.0"
```
prashantv: normally we bump the version

kriskowal: Brilliant. Pushed. I made the bold but uncertain assumption that the next version will be a minor, not patch.

prashantv: I think that's a great assumption based on the past!

PR #629, opened 2017-06-13
glide up to upgrade jaeger

yulunli:
The jaeger client version is too old, causing upgrade issues with a bunch other services


yulunli:
I see. Thanks @prashantv

PR #631, opened 2017-06-17
Add option to random load balance to >1 server nodes

### LINKED ISSUE #630: Load balance traffic to N>1 server nodes
Currently the logic of `tchannel-go` to choose a peer is
```go
for i := 0; i < size; i++ {
	popped := l.peerHeap.popPeer()
	if canChoosePeer(popped.HostPort()) {
		ps = popped
		break
	}
	psPopList = append(psPopList, popped)
}
...
```
Which means, the client node will only talk to one server node, the top chosen one, unless there is a change in scores. This means that even if we use the [HRW scorer](https://github.com/uber/tchannel-go/blob/dev/peers/prefer.go) there will be a traffic imbalance, specially for cases where the number of client and server nodes are low.

I would like to introduce `peerConnectionCount` (`default=1`), so that we can pick the top `peerConnectionCount` nodes and do a random load balancing among those instead of talking to only one node. This should reduce the imbalance of traffic to server nodes.

In future, I think it's better if the whole logic of score calculator, peer heap and the new peer connection count, move to something like `PeerSelectionStrategy`, so other strategies would be easily added.

cc @prashantv @Raynos

kamran-m:
This closes #630 

I have also verified the change for a `client => server` with low number of client/server nodes, where the imbalance was more severe (note that here I used `peerConnectionCount`>= `numPeers=5`, which is the extreme case but it's enough to verify the solution):

screen shot 2017-06-16 at 10 51 25 pm

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -21,7 +21,6 @@
 package tchannel
 
 import (
-	"container/heap"
 	"errors"
 	"strings"
 	"sync"
@@ -63,19 +62,21 @@ type Connectable interface {
 type PeerList struct {
 	sync.RWMutex
 
-	parent          *RootPeerList
-	peersByHostPort map[string]*peerScore
-	peerHeap        *peerHeap
-	scoreCalculator ScoreCalculator
-	lastSelected    uint64
+	parent              *RootPeerList
+	peersByHostPort     map[string]*peerScore
+	peerHeap            *peerHeap
+	scoreCalculator     ScoreCalculator
+	peerConnectionCount uint32
+	lastSelected        uint64
 }
 
 func newPeerList(root *RootPeerList) *PeerList {
 	return &PeerList{
-		parent:          root,
-		peersByHostPort: make(map[string]*peerScore),
-		scoreCalculator: newPreferIncomingCalculator(),
-		peerHeap:        newPeerHeap(),
+		parent:              root,
+		peersByHostPort:     make(map[string]*peerScore),
+		scoreCalculator:     newPreferIncomingCalculator(),
+		peerHeap:            newPeerHeap(),
+		peerConnectionCount: 1,
 	}
 }
 
@@ -89,6 +90,16 @@ func (l *PeerList) SetStrategy(sc ScoreCalculator) {
 		newScore := l.scoreCalculator.GetScore(ps.Peer)
 		l.updatePeer(ps, newScore)
 	}
+}
+
+// SetPeerConnectionCount sets the number of peer connections to be used in
+// combination with the ScoreCalculator to achieve a random load balancing
+// of a single client node to `peerConnectionCount` number of server nodes
+func (l *PeerList) SetPeerConnectionCount(peerConnectionCount uint32) {
+	l.Lock()
+	defer l.Unlock()
+
+	l.peerConnectionCount = peerConnectionCount
 }
 
 // Siblings don't share peer lists (though they take care not to double-connect
@@ -175,8 +186,8 @@ func (l *PeerList) Remove(hostPort string) error {
 	return nil
 }
 func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool) *Peer {
-	var psPopList []*peerScore
-	var ps *peerScore
+	var chosenPSList []*peerScore
+	var poppedList []*peerScore
 
 	canChoosePeer := func(hostPort string) bool {
 		if _, ok := prevSelected[hostPort]; ok {
@@ -191,27 +202,41 @@ func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool) 
 	}
 
 	size := l.peerHeap.Len()
+
+	var connectionCount uint32
 	for i := 0; i < size; i++ {
 		popped := l.peerHeap.popPeer()
+		poppedList = append(poppedList, popped)
 
 		if canChoosePeer(popped.HostPort()) {
-			ps = popped
-			break
+			chosenPSList = append(chosenPSList, popped)
+			connectionCount++
+			if connectionCount >= l.peerConnectionCount {
+				break
+			}
 		}
-		psPopList = append(psPopList, popped)
+
 	}
 
-	for _, p := range psPopList {
-		heap.Push(l.peerHeap, p)
+	for _, p := range poppedList {
+		l.peerHeap.pushPeer(p)
 	}
-
-	if ps == nil {
+	if len(chosenPSList) == 0 {
 		return nil
 	}
 
-	l.peerHeap.pushPeer(ps)
+	ps := randomSampling(chosenPSList)
+	if ps == nil {
+		return nil
+	}
 	ps.chosenCount.Inc()
 	return ps.Peer
+}
+
+func randomSampling(psList []*peerScore) *peerScore {
+	peerRand := trand.NewSeeded()
+	r := peerRand.Intn(len(psList))
+	return psList[r]
 }
 
 // GetOrAdd returns a peer for the given hostPort, creating one if it doesn't yet exist.
--- a/peer_test.go
+++ b/peer_test.go
@@ -697,6 +697,49 @@ func TestPeerSelectionRanking(t *testing.T) {
 	}
 }
 
+func TestPeerRandomSampling(t *testing.T) {
+	const numPeers = 10
+	const numIterations = 1000
+	// Using `numPeers + 1` should just do a random load balancing among `numPeers`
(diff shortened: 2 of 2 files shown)


peer.go
```diff
@@ -175,8 +186,8 @@ func (l *PeerList) Remove(hostPort string) error {
 	return nil
 }
 func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool) *Peer {
-	var psPopList []*peerScore
-	var ps *peerScore
+	var chosenPSList []*peerScore
```
prashantv: presize chosenPSList to avoid a bunch of allocations on each selection, especially when it's > 1

peer.go
```diff
@@ -191,29 +202,43 @@ func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool)
...
 		return nil
 	}
-
-	l.peerHeap.pushPeer(ps)
 	ps.chosenCount.Inc()
 	return ps.Peer
 }
 
+func randomSampling(psList []*peerScore) *peerScore {
+	peerRand := trand.NewSeeded()
```
prashantv: please do not create a new rand for every single peer selection request.

peer.go
```diff
@@ -191,29 +202,43 @@ func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool)
...
-		heap.Push(l.peerHeap, p)
+	for _, p := range poppedList {
+		l.peerHeap.pushPeer(p)
+	}
+	if len(chosenPSList) == 0 {
+		return nil
 	}
 
+	ps := randomSampling(chosenPSList)
 	if ps == nil {
```
prashantv: why would we need this check?

kamran-m:
Thanks for the review @prashantv , I addressed your comments. I also added some benchmark tests with these results:
```
BenchmarkGetPeerWithPeerConnectionCount1  	 5000000	       348 ns/op
BenchmarkGetPeerWithPeerConnectionCount10 	 1000000	      1490 ns/op
```
and also used the same test in the old code  to compare with `peerConnectionCount = 1` case above:
```
BenchmarkGetPeerOld-8   	 5000000	       267 ns/op
```
Please let me know if the `267 => 348` is of any concern.

peer_test.go
```diff
@@ -697,6 +697,118 @@ func TestPeerSelectionRanking(t *testing.T) {
...
+
+	for i := 0; i < numPeers; i++ {
+		hp := fmt.Sprintf("127.0.0.1:60%v", i)
+		ch.Peers().Add(hp)
+	}
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		peer, _ := ch.Peers().Get(nil)
+		if peer == nil {
```
prashantv: should this ever happen? If not, maybe we should do a `t.Fatal` instead of a println

peer_test.go
```diff
@@ -697,6 +697,118 @@ func TestPeerSelectionRanking(t *testing.T) {
...
+			fmt.Println("Just a dummy check to guard against compiler optimization")
+		}
+	}
+}
+
+func BenchmarkGetPeerWithPeerConnectionCount10(b *testing.B) {
+	numPeers := 10
+	peerConnectionCount := uint32(10)
```
prashantv: Instead of copying + pasting the whole benchmark, abstract common logic out into a function, and then call it with different args for the `peerConnectionCount` and maybe even `numPeers`

peer_test.go
```diff
@@ -697,6 +697,118 @@ func TestPeerSelectionRanking(t *testing.T) {
...
+
+	testCases := []struct {
+		numPeers            int
+		peerConnectionCount uint32
+		distMin             float64
+		distMax             float64
+	}{
+		// the higher `peerConnectionCount` is, the smoother the impact of uneven scores
+		// become as we are random sampling among `peerConnectionCount` peers
+		{numPeers: 10, peerConnectionCount: 1, distMin: 1000, distMax: 1000},
```
prashantv: can you add `peerConnectionCount: 2`, i imagine this will be a pretty small value normally?

peer_test.go
```diff
@@ -697,6 +697,118 @@ func TestPeerSelectionRanking(t *testing.T) {
...
+		{numPeers: 10, peerConnectionCount: 5, distMin: 160, distMax: 240},
+		{numPeers: 10, peerConnectionCount: 10, distMin: 50, distMax: 150},
+		{numPeers: 10, peerConnectionCount: 15, distMin: 50, distMax: 150},
+	}
+
+	for _, tc := range testCases {
+		// Selected is a map from rank -> [peer, count]
+		// It tracks how often a peer gets selected at a specific rank.
+		selected := make([]map[string]int, tc.numPeers)
```
prashantv: why do you make a slice here, why can't we just create a singel map in the loop for `numIterations`, and do the `testDistribution` call right there too? Makes 3 loops 1 loop, removes a slice, and simplifies the test a little.

peer_test.go
```diff
@@ -697,6 +697,118 @@ func TestPeerSelectionRanking(t *testing.T) {
...
+	ch.Peers().SetPeerConnectionCount(peerConnectionCount)
+
+	for i := 0; i < numPeers; i++ {
+		hp := fmt.Sprintf("127.0.0.1:60%v", i)
+		ch.Peers().Add(hp)
+	}
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		peer, _ := ch.Peers().Get(nil)
```
prashantv: we should think about testing the impact of passing a non-nil blacklist with `peerConnectionCount` > 1

peer.go
```diff
@@ -48,6 +47,9 @@ var (
 	// ErrNoNewPeers indicates that no previously unselected peer is available.
 	ErrNoNewPeers = errors.New("no new peer available")
 
+	// ErrZeroPeerConnectionCount indicates that the peer connection count is set to zero.
+	ErrZeroPeerConnectionCount = errors.New("peer connection count must be greater than 0")
```
kamran-m: Makes sense!

peer_test.go
kamran-m: Indeed as the impact of it get's smaller with every extra connection. Added the case for `2`!

peer_test.go
kamran-m: I am not sure if I follow. This is a similar test as `TestPeerSelectionRanking` where we are checking the distribution of the ranking after `numIterations` which means we have to do the checking outside of the loop. Or am I missing something here?

peer_test.go
kamran-m: You mean if `prevSelected` actually has some selected peers? I am assuming if a client uses `peerConnectionCount > 1`, it should keep it into account that the prevSelected peers is going to be a bigger list. Or do you have another specific issue in mind?

peer_test.go
kamran-m: done!

peer_test.go
kamran-m: It shouldn't, I just added it to guard against compiler optimization not to artificially lower the runtime of the benchmark. Changed it to Fatal!

peer_test.go
kamran-m: Indeed as the impact of it get's smaller with every extra connection. Added the case for `2`!

peer_test.go
kamran-m: I am not sure if I follow. This is a similar test as `TestPeerSelectionRanking` where we are checking the distribution of the ranking after `numIterations` which means we have to do the checking outside of the loop. Or am I missing something here?

peer_test.go
kamran-m: You mean if `prevSelected` actually has some selected peers? I am assuming if a client uses `peerConnectionCount > 1`, it should keep it into account that the prevSelected peers is going to be a bigger list. Or do you have another specific issue in mind?

peer_test.go
kamran-m: It shouldn't, I just added it to guard against compiler optimization not to artificially lower the runtime of the benchmark. Changed it to Fatal!

kamran-m:
I agree with the idea that the core should not be strategy aware. As I commented on #630 I think all the logic regarding peer connection count, score strategy, peer heap, etc should move to something like PeerSelectionStrategy. 

I am not sure why you think with 2 connected peers we won't need this core change and also get a better load distribution. This change is basically letting the client decide how many connected peers they want. So in cases, like ours, were we have a very low number of server/client nodes, it might actually be helpful to use 5 connected peers as you can verify with the stats I posted above.

Let me know your thoughts.

PR #634, opened 2017-07-27
Pin apache/thrift 0.10.0

kriskowal:
apache/thrift/master breaks TChannel with an API change.
If they are playing by the rules, that change will come in 0.11.0, so pinning
to the ~0.10 range.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -2,7 +2,7 @@ hash: 111bdf9797d23d4b9f8833a9780e382dd53153af70ad286f99815cea4c2fdf4b
 updated: 2016-08-05T18:02:17.171302481-04:00
 imports:
 - name: github.com/apache/thrift
-  version: 0e9fed1e12ed066865e46c6903782b2ef95f4650
+  version: b2a4d4ae21c789b689dd162deb819665567f481c
   subpackages:
   - lib/go/thrift
 - name: github.com/bmizerany/perks
@@ -10,20 +10,22 @@ imports:
   subpackages:
   - quantile
 - name: github.com/cactus/go-statsd-client
-  version: 91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d
+  version: ad551ee7f9f3465fb1ce1695899c612f7808a06a
   subpackages:
   - statsd
+- name: github.com/codahale/hdrhistogram
+  version: 3a0bb77429bd3a61596f5e8a3172445844342120
 - name: github.com/crossdock/crossdock-go
   version: 228792ab861c86f8900b2db0439577feef9ec9d8
   subpackages:
   - assert
   - require
 - name: github.com/davecgh/go-spew
-  version: 5215b55f46b2b919f50a1df0eaa5886afe4e3b3d
+  version: adab96458c51a58dc1783b3335dcce5461522e75
   subpackages:
   - spew
 - name: github.com/jessevdk/go-flags
-  version: f2785f5820ec967043de79c8be97edfc464ca745
+  version: 6cf8f02b4ae8ba723ddc64dcfd403e530c06d927
 - name: github.com/opentracing/opentracing-go
   version: 855519783f479520497c6b3445611b05fc42f009
   subpackages:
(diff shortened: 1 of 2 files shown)


kriskowal:
This will cause a ripple through any dependee that indirectly also depends on apache/thrift as `master`, however, these dependees are already broken since TChannel will not build with `master` when they `glide up`.

breerly:
Watching this ripple through our libs is really sad ;(

prashantv:
Can we make this work with 0.9.3 -> 0.1.0, just to reduce the pain users feel when they have different dependencies that want different versions of Thrift?

kriskowal:
@prashantv agreed. Pushed a relaxed constraint, <0.11.0, 🤞  Apache Thrift doesn’t land the breaking change on master in a 0.10.0 patch release.

kriskowal:
After updating glide again, now traces are broken in crossdock. Hey, we’re doing crossdock in TChannel now? That’s cool.

```
✗ [trace] go→ (s1name=go s2encoding=json s2name=python s3encoding=json s3name=go sampled=false) (1/2 passed, 0/2 skipped)
   ├ ✓ ⇒ Trace ID should not be empty in S1(go)
   └ ✗ ⇒ 
	Error Trace:	validateTrace() /home/travis/gopath/src/github.com/uber/tchannel-go/crossdock/behavior/trace/behavior.go:172
			validateTrace() /home/travis/gopath/src/github.com/uber/tchannel-go/crossdock/behavior/trace/behavior.go:181
			Run() /home/travis/gopath/src/github.com/uber/tchannel-go/crossdock/behavior/trace/behavior.go:113
			Run-fm() /home/travis/gopath/src/github.com/uber/tchannel-go/crossdock/main.go:57
	Error:		Not equal: "33313639306663643834623363626139" (expected)
	        != "31690fcd84b3cba9" (actual)
	Messages:	Trace ID must match in S2(python:8081)
```

Investigating.

kriskowal:
`glide.yaml` does not specify a version for `jaeger-client-go`. Adding `^2.7` and pulling the CI crank.

kriskowal:
I got Crossdock to pass by manually editing out the updates to jaeger-client-go, jaeger-lib, and opentracing-go from glide.yaml. So this is not a clean fix. The Jaeger client has not been upgraded in the lockfile for a year, so isolating the version that broke the build will take some more time.

kriskowal:
@prashantv I’m torn. I can fix the testing problem by pinning jaeger-client-go to ^1. That fixes crossdock, but is backward incompatible. Can we pin it for tests?

prashantv:
Is `jaeger-client-go` have major versions after 1.0? Did we ever support versions before 1.0?

breerly:
#godamongstpeeps

kriskowal:
@prashantv jaeger-client-go has been ^2 for a year.

kriskowal:
The problem isn’t versions *before* ^1, but that it would exclude ^2, which work but produce different traces and break crossdock. We could probably get out of this by updating crossdock to pass for ^2 trace identifiers. @yurishkuro

yurishkuro:
not sure why the trace IDs should be that different, I will have a look.

peter-edge:
Didn't we just pin it to 0.9.3?

peter-edge:
Yo everyone, we just did this https://github.com/uber/tchannel-go/issues/608, we can't do this

kriskowal:
#608 doesn’t touch glide.yaml, and glide.yaml was pinned to apache/thrift at master, not 0.9.3. I have no objection to walking it back to <0.10, or even <=0.9.3.

Perhaps I don’t grok how #608 addresses the issue without updating glide.yaml. Should thrift even be in glide.yaml if we're downloading a version in Makefile to get around glide?

peter-edge:
#608 did touch glide.yaml but we said we'd do it in a later PR https://github.com/uber/tchannel-go/pull/615#discussion_r107702627

peter-edge:
So in summary, we need to do 0.9 in glide.yaml.

PR #635, opened 2017-07-31
bugfix: cancel server context on client tcp close

venkat1109:
**Problem Description:**
When a client crashes, the linux kernel will automatically initiate a tcp_close by sending a FIN request to the server. The server kernel, in response, will the ACK the FIN and move to CLOSE_WAIT, waiting for the app to close the socket. However, tchannel doesn't set the context to done on a connection read error on the server, the conn.Close() won't be called until the context deadline expires. This is bad when tchannel is used for long polling. 

**Solution:**
This patch fixes the problem by marking the server side inbound call context as canceled on a tcp read error. It also marks the msgExchange as expired in the mex, so that it can be subsequently removed. There is already a TODO in the existing code to get rid of the goroutine (on Inbound) per call. This patch does not fix this TODO. 

**Testing:**
Created a unit test that reproduces the mentioned problem. The unit test doesn't use the client channel, instead sends a call directly on the tcp connection.

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -984,10 +984,9 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
 
 		ts.RegisterFunc("test", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
 			defer close(serverDoneC)
-			assert.Equal(t, "client", CurrentCall(ctx).CallerName(), "wrong caller name")
 			close(callForwarded)
 			<-ctx.Done()
-			assert.Equal(t, "context canceled", ctx.Err().Error(), "ctx.Err() returned unexpected error")
+			assert.EqualError(t, ctx.Err(), "context canceled")
 			return &raw.Res{}, nil
 		})
 
@@ -1005,9 +1004,7 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
 		ctx, cancel := NewContext(20 * time.Second)
 		defer cancel()
 
-		clientCh := ts.NewClient(&testutils.ChannelOpts{
-			ServiceName: "client",
-		})
+		clientCh := ts.NewClient(nil)
 		// initiate the call in a background routine and
 		// make it wait for the response
 		clientDoneC := make(chan struct{})
--- a/connection_test.go
+++ b/connection_test.go
@@ -21,6 +21,7 @@
 package tchannel_test
 
 import (
+	goctx "context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


connection_test.go
```diff
@@ -970,3 +970,49 @@ func TestPeerStatusChangeServer(t *testing.T) {
 	})
 	assert.Len(t, changes, 0, "unexpected peer status changes")
 }
+
+func TestContextCanceledOnClientTCPClose(t *testing.T) {
+
```
prashantv: nit: delete blank line

connection_test.go
```diff
@@ -970,3 +970,49 @@ func TestPeerStatusChangeServer(t *testing.T) {
...
+	opts := testutils.NewOpts().DisableLogVerification().NoRelay()
```
prashantv: do you need log verification? Please add a comment about this line for why these are disabled.

connection_test.go
```diff
@@ -970,3 +970,49 @@ func TestPeerStatusChangeServer(t *testing.T) {
...
+
+	testutils.WithTestServer(t, opts, func(server *testutils.TestServer) {
+
```
prashantv: nit: delete blank. In general, we don't do blank lines after a `{` or before a `}`

connection_test.go
```diff
@@ -970,3 +970,49 @@ func TestPeerStatusChangeServer(t *testing.T) {
...
+			close(dispatchC)
+			<-ctx.Done()
+			assert.Equal(t, "test-client", CurrentCall(ctx).CallerName(), "expected caller name to be passed through")
+			return &raw.Res{}, nil
+		})
+
+		conn, err := net.Dial("tcp", server.HostPort())
+		assert.NoError(t, err, "tcp connect to server failed")
+
+		err = DoInitHandshake(conn, 1, conn.LocalAddr().String())
```
prashantv: This test is very low level, but doesn't need to be.

Another way to write this test:
 - set up a [frame relay](http://godoc.org/github.com/uber/tchannel-go/testutils#FrameRelay) which will let you close a connection by calling the returned cancel function
 - set up a server that has a handler that just waits for the context `Done` and verifies that the error is `Canceled` once it's complete
 - make a call in a background goroutine with a long timeout (like 10 seconds) 
 - wait for the call on the frame relay, once you see it, close the relay, which will cause all connections to close, and *should* trigger the server context to end
 - verify that the server handler is done (wait for a channcel to close with a timeout that's significantly shorter than the call timeout)

We don't want tests that rely on writing out frames exactly.

venkat1109: Done

connection_test.go
```diff
@@ -970,3 +970,76 @@ func TestPeerStatusChangeServer(t *testing.T) {
...
+	// that comes with the test server
+	opts := testutils.NewOpts().NoRelay().AddLogFilter("simpleHandler OnError", 1)
+
+	testutils.WithTestServer(t, opts, func(ts *testutils.TestServer) {
+		serverDoneC := make(chan struct{})
+		callForwarded := make(chan struct{})
+
+		ts.RegisterFunc("test", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
+			defer close(serverDoneC)
+			assert.Equal(t, "client", CurrentCall(ctx).CallerName(), "wrong caller name")
```
prashantv: is this validation important for this test? if not, let's remove it

connection_test.go
```diff
@@ -970,3 +970,76 @@ func TestPeerStatusChangeServer(t *testing.T) {
...
+			close(callForwarded)
+			<-ctx.Done()
+			assert.Equal(t, "context canceled", ctx.Err().Error(), "ctx.Err() returned unexpected error")
```
prashantv: use `assert.EqualError` here,
```go
assert.EqualError(t, ctx.Err(), "context canceled")
```

connection_test.go
```diff
@@ -970,3 +970,76 @@ func TestPeerStatusChangeServer(t *testing.T) {
...
+
+		// Make a call with a long timeout. We shutdown the relay
+		// immediately after the server receives the call. Expected
+		// behavior is for both client/server to be done with the call
+		// immediately after relay shutsdown
+		ctx, cancel := NewContext(20 * time.Second)
+		defer cancel()
+
+		clientCh := ts.NewClient(&testutils.ChannelOpts{
+			ServiceName: "client",
```
prashantv: does the client name matter for this test? if not, just do `clientCh := ts.NewClient(nil)`

inbound.go
```diff
@@ -189,6 +189,12 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
 			if c.log.Enabled(LogLevelDebug) {
 				call.log.Debugf("Wait for timeout/cancellation interrupted by error: %v", call.mex.errCh.err)
 			}
+			// conn error, cancel the inbound call
```
prashantv: can you fix the comment to be avoid abbreviations,
```go
// when an exchange errors out, mark the exchange as expired
// and call cancel so the server handler's context is canceled
```

PR #637, opened 2017-08-03
Add tchannel.WithoutHeaders

peter-edge:
This adds a function `tchannel.WithoutHeaders` (name TBD), which hides
TChannel Go-specific information from a Context.

This is needed to alleviate the following issue:

Existing TChannel Go services have been using the standard header
propagation behavior of TChannel Go (incoming headers are copied to
outgoing headers) happily and they start using YARPC+TChannel, or a
library that uses YARPC, to make outgoing calls to a different service.

Their expectation would be to have the headers copied over to that
outgoing call, especially if the underlying transport is TChannel. This
will not be supported in YARPC.

That leaves us with two options: silently ignore TChannel headers, or
reject outgoing requests made with TChannel contexts.

If we silently ignore TChannel headers, service owners who rely on this
behavior will suddenly start seeing hard-to-debug breakages. The number
of support tickets for this will be huge.

A better option is for YARPC to reject TChannel contexts on **all**
transports. YARPC will recommend that users remove the header
information from the context (with the use of `tchannel.WithoutHeaders`)
before passing it to YARPC. In the short-term, this adds a little work
for existing TChannel services but in the long run it saves them a lot
of time in terms of debugging.

### DIFF
--- a/context_header.go
+++ b/context_header.go
@@ -117,16 +117,5 @@ func WrapWithHeaders(ctx context.Context, headers map[string]string) ContextWith
 
 // WithoutHeaders hides any TChannel headers from the given context.
 func WithoutHeaders(ctx context.Context) context.Context {
-	return withoutHeaders{ctx}
-}
-
-type withoutHeaders struct {
-	context.Context
-}
-
-func (ctx withoutHeaders) Value(key interface{}) interface{} {
-	if key == contextKeyHeaders || key == contextKeyTChannel {
-		return nil
-	}
-	return ctx.Context.Value(key)
+	return context.WithValue(context.WithValue(ctx, contextKeyTChannel, nil), contextKeyHeaders, nil)
 }
--- a/context_header.go
+++ b/context_header.go
@@ -115,7 +115,7 @@ func WrapWithHeaders(ctx context.Context, headers map[string]string) ContextWith
 	return headerCtx{Context: newCtx}
 }
 
-// WithoutHeaders removes any TChannel headers from the given context.
+// WithoutHeaders hides any TChannel headers from the given context.
 func WithoutHeaders(ctx context.Context) context.Context {
 	return withoutHeaders{ctx}
 }
--- a/context_header.go
+++ b/context_header.go
@@ -114,3 +114,19 @@ func WrapWithHeaders(ctx context.Context, headers map[string]string) ContextWith
 	newCtx := context.WithValue(ctx, contextKeyHeaders, h)
 	return headerCtx{Context: newCtx}
 }
+
+// WithoutHeaders removes any TChannel headers from the given context.
+func WithoutHeaders(ctx context.Context) context.Context {
+	return withoutHeaders{ctx}
+}
+
+type withoutHeaders struct {
+	context.Context
+}
+
+func (ctx withoutHeaders) Value(key interface{}) interface{} {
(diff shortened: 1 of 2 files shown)


context_header.go
```diff
@@ -114,3 +114,19 @@ func WrapWithHeaders(ctx context.Context, headers map[string]string) ContextWith
 	newCtx := context.WithValue(ctx, contextKeyHeaders, h)
 	return headerCtx{Context: newCtx}
 }
+
+// WithoutHeaders removes any TChannel headers from the given context.
```
akshayjshah: Should we instead say that it "hides" the headers?

context_internal_test.go
```diff
@@ -55,3 +55,12 @@ func TestCurrentSpan(t *testing.T) {
...
 	assert.NotEqual(t, uint64(0), span.TraceID(), "mock tracer is now Zipkin-compatible")
 }
+
+func TestContextWithoutHeaders(t *testing.T) {
+	ctx := WrapWithHeaders(context.Background(), map[string]string{"k1": "v1"})
+	assert.Equal(t, map[string]string{"k1": "v1"}, ctx.Headers())
+	ctx2 := WithoutHeaders(ctx)
+	assert.Nil(t, ctx2.Value(contextKeyHeaders))
+	_, ok := ctx2.(ContextWithHeaders)
+	assert.False(t, ok)
```
akshayjshah: We should test that we're also hiding `contextKeyTChannel`. I think we can do so by using `NewContextBuilder` and `SetShardKey`, then testing that we're hiding the shard key.

context_header.go
peter-edge: done

context_internal_test.go
peter-edge: Done

akshayjshah:
Also, let's pull this branch into YARPC and make sure that all the integration tests still pass (basically, let's make sure that previous-us didn't use this data for something).

peter-edge:
YARPC passes

context_header.go
```diff
@@ -114,3 +114,19 @@ func WrapWithHeaders(ctx context.Context, headers map[string]string) ContextWith
 	newCtx := context.WithValue(ctx, contextKeyHeaders, h)
 	return headerCtx{Context: newCtx}
 }
+
+// WithoutHeaders hides any TChannel headers from the given context.
+func WithoutHeaders(ctx context.Context) context.Context {
+	return withoutHeaders{ctx}
```
prashantv: instead of another type, can we just use `context.WithValue` to override the current key to `nil`?

context_header.go
```diff
@@ -114,3 +114,19 @@ func WrapWithHeaders(ctx context.Context, headers map[string]string) ContextWith
 	newCtx := context.WithValue(ctx, contextKeyHeaders, h)
 	return headerCtx{Context: newCtx}
 }
+
+// WithoutHeaders hides any TChannel headers from the given context.
+func WithoutHeaders(ctx context.Context) context.Context {
```
prashantv: to avoid wrapping without any reason, can we only do the wrapping if the current context has a `non-nil` value for the provided keys?

peter-edge: The real thing we're trying to do here is make sure `ctx` is not a `ContextWithHeaders`, so we want to wrap regardless

context_header.go
peter-edge: Done

PR #638, opened 2017-08-04
Pin Apache Thrift to version range ~0.9

kriskowal:
Finishing #615.

Apache Thrift introduces breaking API changes in version 0.10.0.
This pin will shield dependees from breaking changes impacting TChannel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+# next
+
+* Locks Apache Thrift to version ~0.9 to maintain backward-compatibility, this
+  time for real.
+
 # v1.6.0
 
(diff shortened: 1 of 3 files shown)


kriskowal:
Hey @prashantv please merge and find someone to do a release. Heading out for wedding anniversary. See you Monday.

prashantv:
I'm pretty confident tchannel-go works fine with either 0.9.3 or 0.10.0 version of the Thrift library (and Thrift generated code). (At least `make test` is fine)

I don't think there's breaking changes in the library that affect TChannel, so can we add some context on why we need to do this?

peter-edge:
Yea I buy it, @prashantv and I talked offline

PR #639, opened 2017-08-04
Update version and CHANGELOG for v1.7.0 release

prashantv:


CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 Changelog
 =========
 
+# v1.7.0
```
breerly: Can you circle back and add dates to these?

For example, https://github.com/yarpc/yarpc-go/blob/dev/CHANGELOG.md.

prashantv: Yeah sure, done in https://github.com/uber/tchannel-go/pull/642

PR #649, opened 2017-09-19
glide.yaml improvements: use testImports, only pin versions

### LINKED ISSUE #646: Leave dependencies unpinned instead of pinning to master
Imagine my surprise when, debugging some internal code, I realized that *we're* the ones pinning a bunch of dependencies to `master`!

If I recall, we started doing this a long time ago to cope with version conflicts. I don't think it's necessary anymore - should we instead leave these dependencies unpinned, so that service owners can choose the semver range they'd like to use?

prashantv:
Previously, glide.yaml listed all dependencies (even if they were test
dependencies). Instead, exclude all non-production dependencies to
reduce conflicts for users, and only pin to version ranges. If
dependencies do not have tagged releases, leave them unpinned.

Fixes #646


glide.yaml
```diff
@@ -1,50 +1,56 @@
 package: github.com/uber/tchannel-go
+excludeDirs:
```
breerly: This is nice, thanks.

glide.yaml
```diff
@@ -1,50 +1,56 @@
...
+  - context
+  - ipv4
+  - ipv6
+- package: github.com/uber-go/atomic
+  version: ^1
+  subpackages:
+  - utils
+- package: github.com/opentracing/opentracing-go
+  version: ^1
+testImport:
```
breerly: yish!

glide.yaml
```diff
@@ -1,50 +1,56 @@
 package: github.com/uber/tchannel-go
+excludeDirs:
+ - benchmark
+ - crossdock
+ - examples
+ - relay/relaytest
+ - testutils
```
akshayjshah: Why exclude the test helpers?

glide.yaml
```diff
@@ -1,50 +1,56 @@
...
+  version: ^1
+testImport:
+- package: github.com/jessevdk/go-flags
+  version: ^1
+  subpackages:
+  - ext
+  - mocktracer
 - package: github.com/stretchr/testify
-  version: master
+  version: ^1
```
akshayjshah: Are we using any post-1.0 features? Most of our projects end up with `^1.1.4`.

glide.yaml
```diff
@@ -1,50 +1,56 @@
...
-- package: github.com/uber-go/atomic
+- package: github.com/davecgh/go-spew
+  version: ^1
+  subpackages:
+  - spew
 - package: github.com/prashantv/protectmem
-- package: github.com/opentracing/opentracing-go
+- package: github.com/streadway/quantile
+- package: gopkg.in/yaml.v2
+- package: github.com/crossdock/crossdock-go
```
akshayjshah: 😿  https://github.com/crossdock/crossdock-go/issues/15

glide.yaml
prashantv: Turns out we are fine with ^1.1 (checked out that version and ran tchannel with it)

PR #657, opened 2017-10-27
relay: Create timer after adding item to Items list

### LINKED ISSUE #687: relay: Race condition in relay timer code causes connections to get stuck on close
Issue was discovered in October 207 (See https://github.com/uber/tchannel-go/pull/657#issue-149306386) but was not fixed, as it required some significant changes to the timer logic in the relay code path.

The crux of the issue is that we have:
```go
item.Timer = time.AfterFunc(ttl, func() { r.timeoutRelayItem(...) })
items.Add(id, item)	
```

However, if the running goroutine is scheduled out after the `item.Timer = ` line, and the timer executes, then the `items` map does not yet know about the specified ID. The `timeoutRelayItem` fails to find the id in the `items` map, and logs an error `Can't find relay item to entomb.`. The original goroutine may then be reschedule and then adds the item to the map, but we never clear it out.

This leads to issue if we try to close the connection gracefully, as the state changes to `connectionStartClose`, but the connection can never close as there are pending relay items. Since the TCP connection is still open, the remote side could send calls which will return an error. An error is logged: `Failed to relay frame.` due to `incoming connection is not active: connectionStartClose `.

### LINKED ISSUE #688: relay: Race condition can cause call to timeout and be successful with confusing messages
When a call is complete, we currently [delete the ID from being tracked](https://github.com/uber/tchannel-go/blob/6443d805008bf69bb91325cc5f16a46ae8fe1135/relay.go#L114), then after releasing the lock, try to [stop the timeout timer](https://github.com/uber/tchannel-go/blob/6443d805008bf69bb91325cc5f16a46ae8fe1135/relay.go#L120).

This has a race condition -- if the timer is firing (but hasn't yet [tombed](https://github.com/uber/tchannel-go/blob/6443d805008bf69bb91325cc5f16a46ae8fe1135/relay.go#L146) the relayItem), then we end up with a timer that is running `timeoutRelayItem`, but the element may have been deleted already (or is in the process).

If the item is deleted before we [lookup the item](https://github.com/uber/tchannel-go/blob/6443d805008bf69bb91325cc5f16a46ae8fe1135/relay.go#L134), then we end up throwing a spurious error message, `Can't find relay item to entomb.`.

If the item is deleted after we do the lookup, then we'll likely schedule a clearing of the entombed item which will fail to find the item and log a spurious error message, `Attempted to delete non-existent relay item.`.

This race condition could also cause the same call to be reported as a success and a timeout. It's not clear what other negative impact this race has, but given #687, I'd prefer to remove all races of this kind and ensure logs are clean.

prashantv:
We currently create a timer before adding the relay item to the list,
but this has the potential to race. If we create a timer which expires
before we add the item, the timeout will not be able to entomb the item
since we won't find the ID in the items list. If this call never
completes, we'll end up leaking the relay item.

To support this use-case, we need to create a timer separately from when we start it. To make this easier, add a new relay timer pool that supports separate creation/start.

As a side-benefit, this ends up reducing allocations per call from 16 allocs to 12 allocs, and improves performance slightly. This is useful since we'll likely take a slight performance hit to fix #688.

Fixes #687, #688

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -50,22 +50,22 @@ var (
 )
 
 type relayItem struct {
-	*time.Timer
-
 	remapID     uint32
 	tomb        bool
 	local       bool
 	call        RelayCall
 	destination *Relayer
 	span        Span
+	timeout     *relayTimer
 }
 
 type relayItems struct {
 	sync.RWMutex
 
-	logger Logger
-	tombs  uint64
-	items  map[uint32]relayItem
+	logger   Logger
+	timeouts *relayTimerPool
+	tombs    uint64
+	items    map[uint32]relayItem
 }
 
 func newRelayItems(logger Logger) *relayItems {
@@ -117,7 +117,8 @@ func (r *relayItems) Delete(id uint32) (relayItem, bool) {
 	}
 	r.Unlock()
 
-	item.Stop()
+	item.timeout.Stop()
+	item.timeout.Release()
 	return item, !item.tomb
 }
 
@@ -179,6 +180,9 @@ type Relayer struct {
 	// It stores remappings for all response frames read on this connection.
 	inbound *relayItems
 
+	// TODO
+	timeouts *relayTimerPool
+
 	peers   *RootPeerList
 	conn    *Connection
 	logger  Logger
@@ -187,7 +191,7 @@ type Relayer struct {
 
 // NewRelayer constructs a Relayer.
 func NewRelayer(ch *Channel, conn *Connection) *Relayer {
-	return &Relayer{
+	r := &Relayer{
 		relayHost:    ch.RelayHost(),
 		maxTimeout:   ch.relayMaxTimeout,
 		localHandler: ch.relayLocal,
@@ -197,6 +201,8 @@ func NewRelayer(ch *Channel, conn *Connection) *Relayer {
 		conn:         conn,
 		logger:       conn.log,
 	}
+	r.timeouts = newRelayTimerPool(r.timeoutRelayItem)
+	return r
 }
 
 // Relay is called for each frame that is read on the connection.
@@ -457,12 +463,13 @@ func (r *Relayer) addRelayItem(isOriginator bool, id, remapID uint32, destinatio
 	if isOriginator {
 		items = r.outbound
 	}
+	item.timeout = r.timeouts.Get()
 	items.Add(id, item)
-	item.Timer = time.AfterFunc(ttl, func() { r.timeoutRelayItem(isOriginator, items, id) })
+	item.timeout.Start(ttl, items, id, isOriginator)
 	return item
 }
 
-func (r *Relayer) timeoutRelayItem(isOriginator bool, items *relayItems, id uint32) {
+func (r *Relayer) timeoutRelayItem(items *relayItems, id uint32, isOriginator bool) {
 	item, ok := items.Entomb(id, _relayTombTTL)
 	if !ok {
 		return
--- a/relay.go
+++ b/relay.go
@@ -457,8 +457,8 @@ func (r *Relayer) addRelayItem(isOriginator bool, id, remapID uint32, destinatio
 	if isOriginator {
 		items = r.outbound
 	}
-	item.Timer = time.AfterFunc(ttl, func() { r.timeoutRelayItem(isOriginator, items, id) })
 	items.Add(id, item)
+	item.Timer = time.AfterFunc(ttl, func() { r.timeoutRelayItem(isOriginator, items, id) })
 	return item
 }
 
--- a/relay.go
+++ b/relay.go
@@ -181,7 +181,7 @@ type Relayer struct {
 	inbound *relayItems
 
 	// timeouts is the pool of timers used to track call timeouts.
-	// It allows timer re-use, while allow timers to be created and started separately.
+	// It allows timer re-use, while allowing timers to be created and started separately.
 	timeouts *relayTimerPool
 
 	peers   *RootPeerList
--- a/relay.go
+++ b/relay.go
@@ -117,7 +117,8 @@ func (r *relayItems) Delete(id uint32) (relayItem, bool) {
 	}
 	r.Unlock()
 
-	cancelRelayTimer(item.timeout)
+	item.timeout.Stop()
+	item.timeout.Release()
 	return item, !item.tomb
 }
 
@@ -143,15 +144,9 @@ func (r *relayItems) Entomb(id uint32, deleteAfter time.Duration) (relayItem, bo
 		return item, false
 	}
 	r.tombs++
-
-	prevTimeout := item.timeout
-	item.timeout = nil
 	item.tomb = true
 	r.items[id] = item
 	r.Unlock()
-
-	// We need to check whether the item had a timeout that we need to stop.
-	cancelRelayTimer(prevTimeout)
 
 	// TODO: We should be clearing these out in batches, rather than creating
 	// individual timers for each item.
@@ -245,6 +240,17 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
 		return false, "relay-not-found"
 	}
 
+	finished := finishesCall(f)
+	if item.tomb {
+		// Call timed out, ignore this frame. (We've already handled stats.)
+		// TODO: metrics for late-arriving frames.
+		return true, ""
+	}
+	if finished && !item.timeout.Stop() {
+		// Timeout is firing, so no point proxying this frame
+		return true, ""
+	}
+
 	// call res frames don't include the OK bit, so we can't wait until the last
 	// frame of a relayed RPC to determine if the call succeeded.
 	if fType == responseFrame {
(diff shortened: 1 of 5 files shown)


relay.go
```diff
@@ -179,6 +180,9 @@ type Relayer struct {
 	// It stores remappings for all response frames read on this connection.
 	inbound *relayItems
 
+	// TODO
```
akshayjshah: Don't forget about this TODO.

relay_timer_pool.go
```diff
@@ -0,0 +1,99 @@
...
+type relayTimerPool struct {
+	pool    sync.Pool
+	trigger relayTimerTrigger
+}
+
+type relayTimer struct {
+	pool  *relayTimerPool
+	timer *time.Timer
+
+	// Per-timer paramters passed back when the timer is triggered.
```
akshayjshah: nit: "parameters"

relay_timer_pool.go
```diff
@@ -0,0 +1,99 @@
...
+	if ok {
+		return timer
+	}
+
+	rt := &relayTimer{
+		pool: tp,
+	}
+	rt.timer = time.AfterFunc(time.Duration(math.MaxInt64), rt.OnTimer)
+	if !rt.timer.Stop() {
+		panic("failed to stop timer set for 1 hour")
```
akshayjshah: A comment would be nice here, just to explain that we're creating a timer to avoid nil checks later on. On first seeing this, I spent a while trying to figure out if there was a more complex reason to have this logic here.

Assuming that I'm right, and that this is intended to avoid nil checks elsewhere, is it worth the cost? I don't really understand the implementations of the `startTimer` and `stopTimer` functions in the stdlib's `time/sleep.go`, but I assume that they're inserting and removing the timer from the heap, which I assume requires locking and unlocking it. In the context of this relay, that seems a little expensive. (Yes, that's a lot of assumptions.)

Lastly, is this message accurate? It looks like we've set the timer for *much* longer than an hour.

relay_timer_pool.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return rt
+}
+
+// Stop stops the timer and returns whether the timer was stopped. It returns
+// the same behaviour as https://golang.org/pkg/time/#Timer.Stop.
```
akshayjshah: Should we stick to American English spellings?

relay_timer_pool.go
```diff
@@ -0,0 +1,99 @@
...
+	return rt.timer.Stop()
+}
+
+// Release releases a timer back to the timer pool. The timer MUST be stopped
+// before being released.
+func (rt *relayTimer) Release() {
+	if rt.Stop() {
+		panic("tried to release unstopped timer")
+	}
+	rt.id = 0
```
akshayjshah: Should we reset `items` and `isOriginator` too?

relay.go
```diff
@@ -457,12 +463,13 @@ func (r *Relayer) addRelayItem(isOriginator bool, id, remapID uint32, destinatio
 	if isOriginator {
 		items = r.outbound
 	}
-	item.Timer = time.AfterFunc(ttl, func() { r.timeoutRelayItem(isOriginator, items, id) })
+	item.timeout = r.timeouts.Get()
 	items.Add(id, item)
+	item.timeout.Start(ttl, items, id, isOriginator)
```
akshayjshah: Wow, this is a pretty subtle race - nice catch!

relay_timer_pool.go
```diff
@@ -0,0 +1,99 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package tchannel
```
akshayjshah: Need a blank line between copyright and package declaration - otherwise `golint` thinks that this is the package GoDoc.

prashantv:
Sorry, I should have marked this as `WIP`, the fix turned out to be much larger than expected, and so I'm playing with different approaches to see how we fix this race.

This is probably not going to be the final form, so I've added a WIP prefix.

relay.go
prashantv: yeah, I was looking at why we were seeing a bunch of spurious logs that led me to this. Unfortunately, I didn't fix it early enough, and the race ended up happening in production:
https://github.com/uber/tchannel-go/issues/687

There's a few other reasons for spurious logs,
https://github.com/uber/tchannel-go/issues/688

Lots of race conditions related to timeouts in the relay paths, so I'm going to try and clean them all up.

relay_timer_pool.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Stop stops the timer and returns whether the timer was stopped. It returns
+// the same behaviour as https://golang.org/pkg/time/#Timer.Stop.
+func (rt *relayTimer) Stop() bool {
+	stopped := rt.timer.Stop()
+	if stopped {
+		rt.markTimerInactive()
+	}
+	return stopped
```
akshayjshah: Can we leave a comment here explaining that we *expect* to sometimes call this method concurrently, since we're using it to gate access to other methods? We're very defensive in `Start`, but permissive here - let's help future us and explain why. I don't expect that we'll be in here often, but that'll also remind us that this method needs to be safe for concurrent use, unlike the others.

relay.go
```diff
@@ -233,6 +239,17 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
...
+	finished := finishesCall(f)
+	if item.tomb {
+		// Call timed out, ignore this frame. (We've already handled stats.)
+		// TODO: metrics for late-arriving frames.
+		return true, ""
+	}
+	if finished && !item.timeout.Stop() {
+		// Timeout is firing, so no point proxying this frame
+		return true, ""
+	}
```
akshayjshah: Ahhhh, this makes sense. I think that this resolves an open TODO in the internal metrics code - these blocks should prevent the race we were seeing in use of the per-call metrics object.

relay_internal_test.go
```diff
@@ -46,3 +47,35 @@ func TestFinishesCallResponses(t *testing.T) {
...
+	for _, tt := range tests {
+		trigger := func(*relayItems, uint32, bool) {}
+		rtp := newRelayTimerPool(trigger)
+
+		rt := rtp.Get()
+		assert.Panics(t, func() {
+			tt.f(rt)
+		}, tt.msg)
+	}
+}
```
akshayjshah: Can we add a unit test (vs the integration tests elsewhere) that ensures that `Stop` is safe to use concurrently?

relay_test.go
```diff
@@ -791,3 +791,39 @@ func TestRelayConcurrentNewConnectionAttempts(t *testing.T) {
...
+
+		var wg sync.WaitGroup
+		for i := 0; i < concurrentClients; i++ {
+			wg.Add(1)
+			go func(client *Channel) {
+				defer wg.Done()
+
+				for j := 0; j < callsPerClient; j++ {
+					// Make many concurrent calls, some will timeout
+					ctx, cancel := NewContext(5 * time.Millisecond)
```
akshayjshah: Given the many, many issues we've run into with OS X coalescing timers (last time: https://twitter.com/akshayshah/status/830296255196655616), can we set this to >= 15ms or an obviously "time out immediately" value like `time.Nanosecond`?

relay_test.go
```diff
@@ -791,3 +791,39 @@ func TestRelayConcurrentNewConnectionAttempts(t *testing.T) {
...
+					raw.Call(ctx, client, ts.HostPort(), ts.ServiceName(), "echo", nil, nil)
+					cancel()
+				}
+			}(clients[i])
+		}
+
+		wg.Wait()
+	})
+}
```
akshayjshah: Triggering this bug was depressingly straightforward :/

relay_timer_pool.go
```diff
@@ -0,0 +1,121 @@
...
+func (rt *relayTimer) Start(d time.Duration, items *relayItems, id uint32, isOriginator bool) {
+	if rt.active {
+		panic("Tried to start an already-active timer")
+	}
+
+	rt.active = true
+	rt.items = items
+	rt.id = id
+	rt.isOriginator = isOriginator
+	rt.timer.Reset(d)
```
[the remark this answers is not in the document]
prashantv: Great idea, done

relay_timer_pool.go
```diff
@@ -0,0 +1,121 @@
...
+	"sync"
+	"time"
+)
+
+type relayTimerTrigger func(items *relayItems, id uint32, isOriginator bool)
+
+type relayTimerPool struct {
+	pool    sync.Pool
+	trigger relayTimerTrigger
+}
```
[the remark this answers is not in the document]
prashantv: Done

relay_internal_test.go
prashantv: Yep, added a test which calls `Stop` and tries to race it against itself, but also against the timer firing.

PR #658, opened 2017-10-27
Defer frame allocation in readFrames till we read a header

prashantv:
Currently, we allocate a `*Frame` before we call Read, which may block
for quite a while. With a large number of connections, we end up holding
on to a large amount of memory (64KiB per frame) even if those
connections are idle.

Instead of allocating before the `Read` call, wait till we read a
header, and then allocate a frame.

This will also allow us to allocate a frame based on the size of the
body rather than assuming a worst-case of 64KiB as we do now.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -608,14 +608,27 @@ func (c *Connection) readState() connectionState {
 // incoming frame to a channel; the init handlers are a notable exception,
 // since we cannot process new frames until the initialization is complete.
 func (c *Connection) readFrames(_ uint32) {
+	headerBuf := make([]byte, FrameHeaderSize)
+
+	handleErr := func(err error) {
+		if c.closeNetworkCalled.Load() == 0 {
+			c.connectionError("read frames", err)
+		} else {
+			c.log.Debugf("Ignoring error after connection was closed: %v", err)
+		}
+	}
+
 	for {
+		// Read the header, avoid allocating the frame till we know the size
+		// we need to allocate.
+		if _, err := io.ReadFull(c.conn, headerBuf); err != nil {
+			handleErr(err)
+			return
+		}
+
 		frame := c.opts.FramePool.Get()
-		if err := frame.ReadIn(c.conn); err != nil {
-			if c.closeNetworkCalled.Load() == 0 {
-				c.connectionError("read frames", err)
-			} else {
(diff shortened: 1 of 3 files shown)


frame.go
```diff
@@ -124,28 +124,41 @@ func NewFrame(payloadCapacity int) *Frame {
...
+	default:
+		// No payload to read
+		return nil
 	}
+}
 
-	return nil
+// ReadIn reads the frame from the given io.Reader.
+// This method is deprecated: callers should prefer to read the frame header
+// separately, and using ReadBody.
```
abhinav: Any reason not to use the "Deprecated: .." form which makes linters fail?

connection.go
```diff
@@ -608,14 +608,27 @@ func (c *Connection) readState() connectionState {
 // incoming frame to a channel; the init handlers are a notable exception,
 // since we cannot process new frames until the initialization is complete.
 func (c *Connection) readFrames(_ uint32) {
+	headerBuf := make([]byte, FrameHeaderSize)
+
+	handleErr := func(err error) {
+		if c.closeNetworkCalled.Load() == 0 {
+			c.connectionError("read frames", err)
+		} else {
+			c.log.Debugf("Ignoring error after connection was closed: %v", err)
```
abhinav: does this need to be a closure? it doesn't appear to be accessing anything outside.

frame.go
prashantv: 👍

connection.go
prashantv: doesn't need to be a closure, but wanted to avoid adding yet another method to `*Connection`, this logic is very local to `readFrames`, this was just broken out into a function to DRY up the body.

vkuzmin-uber:
Just for my curiosity. Does this change mean that more likely it will do 2 "read from socket" syscalls instead of 1?

prashantv:
@vkuzmin-uber We already did 2 read syscalls previously:

1. For the header:
```
	if err := f.Header.read(&rbuf); err != nil {		
```

2. For the payload:
```
		if _, err := io.ReadFull(r, f.SizedPayload()); err != nil {		
```

We previously used a single slice to represent the whole payload, while now we read the header into a separate slice, and copy it into the full payload slice. Since it's only 16 bytes, the copy has minimal impact.

PR #663, opened 2017-11-01
thrift: Better logs on errors

prashantv:
Since we log using the channel logger, we have minimal information
(current process name, error, and channel name), but no relevant
information about what specifically failed:
 - the method name
 - the remote process information

We should provide as much useful context on errors to make it easier for
users to understand what failed and the impact.

Error changes from:
```
14:47:06.844295 [D] thrift Server timeout: tchannel error ErrCodeTimeout: timeout tags: [{service testService} {process testService-60389} {chID 1} {service testService} {process testService-60389} {hostPort 127.0.0.1:60389}]
```
to
```
14:46:52.565951 [D] Thrift server timeout. tags: [{service testService} {process testService-60364} {chID 1} {service testService} {process testService-60364} {hostPort 127.0.0.1:60364} {error tchannel error ErrCodeTimeout: timeout} {method SecondService::Echo} {callerName testService-client} {localAddr 127.0.0.1:60364} {remoteHostPort 127.0.0.1:60365} {remoteIsEphemeral true} {remoteProcess testService-client-1}]

```

### DIFF
--- a/thrift/server.go
+++ b/thrift/server.go
@@ -94,13 +94,33 @@ func (s *Server) SetContextFn(f func(ctx context.Context, method string, headers
 	s.ctxFn = f
 }
 
-func (s *Server) onError(err error) {
+func (s *Server) onError(ctx context.Context, method string, err error) {
 	// TODO(prashant): Expose incoming call errors through options for NewServer.
-	// Timeouts should not be reported as errors.
+	logger := s.log.WithFields(
+		tchannel.ErrField(err),
+	)
+
+	// Add additional context about the remote peer if available.
+	if call := tchannel.CurrentCall(ctx); call != nil {
+		remotePeer := call.RemotePeer()
+		logger = logger.WithFields(
+			tchannel.LogField{Key: "method", Value: method},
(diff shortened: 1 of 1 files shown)


thrift/server.go
```diff
@@ -94,13 +94,33 @@ func (s *Server) SetContextFn(f func(ctx context.Context, method string, headers
...
+
+	// Add additional context about the remote peer if available.
+	if call := tchannel.CurrentCall(ctx); call != nil {
+		remotePeer := call.RemotePeer()
+		logger = logger.WithFields(
+			tchannel.LogField{Key: "callerName", Value: call.CallerName()},
+
+			// TODO: These are very similar to the connection fields, but we don't
+			// have access to the connection's logger. Consider exposing the
+			// connection through CurrentCall.
```
abhinav: You have the InboundCall available where you call this method, and InboundCall has access to the connection. Can you not use that?

prashantv: The inbound call has access to the connection in an unexported field, but since we're not in the same package, we can't access it.

However, the `call` has everything that I was getting out of the context, so simplified that logic a little.

PR #668, opened 2017-11-10
Make stats_test less sensitive to the number of timeNow calls

orenkislev:
The tests were written with a specific number of timeNow calls in place, making the implementation difficult to modify.

### DIFF
--- a/stats_test.go
+++ b/stats_test.go
@@ -21,6 +21,7 @@
 package tchannel_test
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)


orenkislev:
Indeed I am adding, in another branch, a time-based mechanism to remove idle connections from a open channels. This results in some additional calls to time.Now() which are making these tests failed. I'm separating these changes into their own branch for ease of review.

prashantv:
Chatted offline with @orenkislev about this, and decided we'll go ahead with this change since it makes it easier to develop, despite the fact that it relies on the test to elapse the clock in the right place.

I've pushed some commits to clean up some other logic:
 * Avoid duplication of the `testHandler` logic
 * Tweak the stats test timeouts so we are confident that we're not mixing up values

@orenkislev Can you review my latest commits, and if it looks good, I can merge.

orenkislev:
LGTM

PR #671, opened 2017-11-27
Automatically close connections after an idle timeout

### LINKED ISSUE #616: Close connections after an idle timeout
Currently, `tchannel-go` holds on to connections indefinitely. As we use this library in more contexts than the original design (compatibility with Hyperbahn), we may not want to hold on to connections forever.

We should add an option for some sort of "idle connection timeout" after which we will close connections automatically. This will require:
 - tracking when the last activity on a connection was
 - a periodic goroutine that will iterate over connections and close them if `now - lastActivity > idleConnectionTimeout`

cc @witriew @billf

orenkislev:
Fixes #616 

If IdleCheckInterval is set (default: disabled), each channel will run a goroutine that periodically checks the last activity time on each connection, and disconnects inbound/outbound connections on which the last activity time exceeds MaxIdleTime (default: 3min). 

Note that pings do not count as activity.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -94,9 +94,10 @@ type ChannelOptions struct {
 	// is set.
 	MaxIdleTime time.Duration
 
-	// IdleCheckInterval controls how often the channel runs the sweep over
-	// all active connections to see if they can be dropped. If this is set to
-	// zero, the idle check is disabled.
+	// IdleCheckInterval controls how often the channel runs a sweep over
+	// all active connections to see if they can be dropped. Connections that
+	// are idle for longer than MaxIdleTime are disconnected. If this is set to
+	// zero (the default), idle checking is disabled.
 	IdleCheckInterval time.Duration
 
 	// Tracer is an OpenTracing Tracer used to manage distributed tracing spans.
--- a/channel.go
+++ b/channel.go
@@ -41,6 +41,7 @@ import (
 var (
 	errAlreadyListening  = errors.New("channel already listening")
 	errInvalidStateForOp = errors.New("channel is in an invalid state for that method")
+	errMaxIdleTimeNotSet = errors.New("IdleCheckInterval is set but MaxIdleTime is zero")
 
 	// ErrNoServiceName is returned when no service name is provided when
 	// creating a new channel.
@@ -229,10 +230,8 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		LogField{"chID", chID},
 	)
 
-	if opts.IdleCheckInterval > 0 && opts.MaxIdleTime <= 0 {
-		logger.Warn("To enable automatically removing idle connections, you must " +
-			"set both IdleCheckInterval and MaxIdleTime.")
-		opts.IdleCheckInterval = 0
+	if err := opts.validateIdleCheck(); err != nil {
+		return nil, err
 	}
 
 	ch := &Channel{
@@ -288,8 +287,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	}
 
 	// Start the idle connection timer.
-	ch.mutable.idleSweep = newIdleSweep(ch, opts)
-	ch.mutable.idleSweep.Start()
+	ch.mutable.idleSweep = startIdleSweep(ch, opts)
 
 	return ch, nil
 }
@@ -812,6 +810,14 @@ func (ch *Channel) Close() {
 // RelayHost returns the channel's RelayHost, if any.
 func (ch *Channel) RelayHost() RelayHost {
 	return ch.relayHost
+}
+
+func (o *ChannelOptions) validateIdleCheck() error {
+	if o.IdleCheckInterval > 0 && o.MaxIdleTime <= 0 {
+		return errMaxIdleTimeNotSet
+	}
+
+	return nil
 }
 
 func toStringSet(ss []string) map[string]struct{} {
--- a/channel.go
+++ b/channel.go
@@ -229,6 +229,12 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		LogField{"chID", chID},
 	)
 
+	if opts.IdleCheckInterval > 0 && opts.MaxIdleTime <= 0 {
+		logger.Warn("To enable automatically removing idle connections, you must " +
+			"set both IdleCheckInterval and MaxIdleTime.")
+		opts.IdleCheckInterval = 0
+	}
+
 	ch := &Channel{
 		channelConnectionCommon: channelConnectionCommon{
 			log:           logger,
@@ -281,8 +287,9 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		opts.RelayHost.SetChannel(ch)
 	}
 
-	// Configure the idle connection timer.
+	// Start the idle connection timer.
 	ch.mutable.idleSweep = newIdleSweep(ch, opts)
+	ch.mutable.idleSweep.Start()
 
 	return ch, nil
 }
@@ -313,9 +320,6 @@ func (ch *Channel) Serve(l net.Listener) error {
 	mutable.peerInfo.HostPort = l.Addr().String()
 	mutable.peerInfo.IsEphemeral = false
 	ch.log = ch.log.WithFields(LogField{"hostPort", mutable.peerInfo.HostPort})
-
-	// Start the idle connection timer.
-	ch.mutable.idleSweep.Start()
 
 	peerInfo := mutable.peerInfo
 	ch.log.WithFields(
@@ -638,8 +642,6 @@ func (ch *Channel) addConnection(c *Connection, direction connectionDirection) b
 		return false
 	}
 
-	// Start the idle connection timer.
-	ch.mutable.idleSweep.Start()
 	ch.mutable.conns[c.connID] = c
 	return true
 }
--- a/channel.go
+++ b/channel.go
@@ -88,6 +88,15 @@ type ChannelOptions struct {
 	// Note: This is not a stable part of the API and may change.
 	TimeTicker func(d time.Duration) *time.Ticker
 
+	// MaxIdleTime controls how long we allow an idle connection to exist
+	// before tearing it down.
+	MaxIdleTime *time.Duration
+
+	// IdleCheckInterval controls how often the channel runs the sweep over
+	// all active connections to see if they can be dropped. If this is set to
+	// zero, the idle check is disabled.
+	IdleCheckInterval *time.Duration
+
 	// Tracer is an OpenTracing Tracer used to manage distributed tracing spans.
 	// If not set, opentracing.GlobalTracer() is used.
 	Tracer opentracing.Tracer
@@ -145,6 +154,7 @@ type Channel struct {
 		state        ChannelState
 		peerInfo     LocalPeerInfo // May be ephemeral if this is a client only channel
 		l            net.Listener  // May be nil if this is a client only channel
+		idleSweep    *IdleSweep
 		conns        map[uint32]*Connection
 	}
 }
@@ -269,6 +279,10 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	if opts.RelayHost != nil {
 		opts.RelayHost.SetChannel(ch)
 	}
+
+	// Configure the idle connection timer.
+	ch.mutable.idleSweep = newIdleSweep(ch, opts)
+
 	return ch, nil
 }
 
@@ -298,6 +312,9 @@ func (ch *Channel) Serve(l net.Listener) error {
 	mutable.peerInfo.HostPort = l.Addr().String()
 	mutable.peerInfo.IsEphemeral = false
 	ch.log = ch.log.WithFields(LogField{"hostPort", mutable.peerInfo.HostPort})
+
+	// Start the idle connection timer.
+	ch.mutable.idleSweep.Start()
 
 	peerInfo := mutable.peerInfo
 	ch.log.WithFields(
@@ -620,6 +637,8 @@ func (ch *Channel) addConnection(c *Connection, direction connectionDirection) b
 		return false
 	}
 
+	// Start the idle connection timer.
+	ch.mutable.idleSweep.Start()
 	ch.mutable.conns[c.connID] = c
 	return true
 }
@@ -758,11 +777,15 @@ func (ch *Channel) Close() {
 	ch.Logger().Info("Channel.Close called.")
 	var connections []*Connection
 	var channelClosed bool
+
 	ch.mutable.Lock()
 
 	if ch.mutable.l != nil {
 		ch.mutable.l.Close()
 	}
+
+	// Stop the idle connections timer.
+	ch.mutable.idleSweep.Stop()
 
 	ch.mutable.state = ChannelStartClose
 	if len(ch.mutable.conns) == 0 {
--- a/channel.go
+++ b/channel.go
@@ -89,13 +89,14 @@ type ChannelOptions struct {
 	TimeTicker func(d time.Duration) *time.Ticker
 
 	// MaxIdleTime controls how long we allow an idle connection to exist
-	// before tearing it down.
-	MaxIdleTime *time.Duration
+	// before tearing it down. Must be set to non-zero if IdleCheckInterval
+	// is set.
+	MaxIdleTime time.Duration
 
 	// IdleCheckInterval controls how often the channel runs the sweep over
 	// all active connections to see if they can be dropped. If this is set to
 	// zero, the idle check is disabled.
-	IdleCheckInterval *time.Duration
+	IdleCheckInterval time.Duration
 
 	// Tracer is an OpenTracing Tracer used to manage distributed tracing spans.
 	// If not set, opentracing.GlobalTracer() is used.
@@ -154,7 +155,7 @@ type Channel struct {
 		state        ChannelState
 		peerInfo     LocalPeerInfo // May be ephemeral if this is a client only channel
 		l            net.Listener  // May be nil if this is a client only channel
-		idleSweep    *IdleSweep
+		idleSweep    *idleSweep
 		conns        map[uint32]*Connection
 	}
 }
--- a/idle_sweep.go
+++ b/idle_sweep.go
@@ -33,6 +33,7 @@ type idleSweep struct {
 	maxIdleTime       time.Duration
 	idleCheckInterval time.Duration
 	stopCh            chan struct{}
+	started           bool
 }
 
 // startIdleSweep starts a poller that checks for idle connections at given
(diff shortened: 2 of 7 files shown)


prashantv:
Thanks for taking this on @orenkislev!

Couple of requests:
 - Can you break out the logic for updating the last activity time from this PR?
 - Run `make test` locally to make sure tests pass locally.

Currently, there's a big race condition in `updateLastActivity`, as it's called from 2 separate goroutines, and may end up trying to update the same `time.Time` at the same time, so I think it's causing tests to fail.

orenkislev:
@prashantv PTAL

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (is *IdleSweep) checkIdleConnections() {
+	// Make a copy of the peer list to reduce the time we keep the lock.
+	rootPeers := is.ch.RootPeers().Copy()
+	now := is.ch.timeNow()
+
+	for hostPort, peer := range rootPeers {
+		// Check idle time on both inbound and outbound connections.
+		for _, conn := range *peer.connectionsFor(inbound) {
+			lastActivity := time.Unix(0, conn.lastActivity.Load())
```
prashantv: add functions to the connection to determine the last activity time as a `time.Time`, rather than inlining that logic here. the `conn` should be encapsulated and we shouldn't reach into the internals.

idle_sweep_test.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+func numConnections(ch *Channel) int {
+	rootPeers := ch.RootPeers().Copy()
+	count := 0
+
+	for _, peer := range rootPeers {
+		in, out := peer.NumConnections()
+
+		ch.Logger().Infof("numConnections IN: %d  OUT: %d", in, out)
```
prashantv: do we need these logs?

idle_sweep_test.go
```diff
@@ -0,0 +1,239 @@
...
+func TestRelayBasedSweep(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	relayTicker := testutils.NewFakeTicker()
+	clock := testutils.NewStubClock(time.Now())
+
+	serverOpts := testutils.NewOpts().
+		SetTimeNow(clock.Now).
+		SetRelayOptionsFn(func(relayOpts *testutils.ChannelOpts) {
```
prashantv: do we need the relay options fn? can we just create a new server which doesn't have tehse options by using `ts.NewServer`, and ignore `ts.Server`?

the `testutils` channel logic is pretty complicated, so I'd like to avoid extra logic there if possible.

idle_sweep_test.go
```diff
@@ -0,0 +1,239 @@
...
+			return &raw.Res{}, nil
+		})
+
+		client := ts.NewClient(clientOpts)
+		raw.Call(ctx, client, ts.HostPort(), ts.ServiceName(), "echo", nil, nil)
+
+		// Both server and client now have an active connection. After 3 minutes they
+		// should be cleared out by the idle sweep.
+		clientTicker.Tick()
+		time.Sleep(testutils.Timeout(10 * time.Millisecond))
```
prashantv: why do we need sleeps here? if anything we should have a `testutils.WaitFor`, but I'm strongly against sleeps in tests, they end up being flaky or cause the tests to slow down

testutils/ticker.go
```diff
@@ -0,0 +1,60 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package testutils
+
+import (
```
prashantv: nit: can simplify to `import "time"`

health_ext_test.go
```diff
@@ -75,7 +46,7 @@ func TestHealthCheckStopBeforeStart(t *testing.T) {
 		})
 		defer cancel()
 
-		ft := newFakeTicker()
+		ft := testutils.NewFakeTicker()
```
prashantv: I've created https://github.com/uber/tchannel-go/pull/674 to move this out separately to reduce noise in this diff (and to add some simple tests to it).

channel.go
```diff
@@ -88,6 +88,15 @@ type ChannelOptions struct {
...
 	TimeTicker func(d time.Duration) *time.Ticker
 
+	// MaxIdleTime controls how long we allow an idle connection to exist
+	// before tearing it down.
+	MaxIdleTime *time.Duration
+
+	// IdleCheckInterval controls how often the channel runs the sweep over
+	// all active connections to see if they can be dropped. If this is set to
+	// zero, the idle check is disabled.
+	IdleCheckInterval *time.Duration
```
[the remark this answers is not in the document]
orenkislev: Done

channel.go
```diff
@@ -299,6 +313,9 @@ func (ch *Channel) Serve(l net.Listener) error {
 	mutable.peerInfo.IsEphemeral = false
 	ch.log = ch.log.WithFields(LogField{"hostPort", mutable.peerInfo.HostPort})
 
+	// Start the idle connection timer.
+	ch.mutable.idleSweep.Start()
```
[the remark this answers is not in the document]
orenkislev: Channels are re-usable after Close(). So we need to have something to start the idle_sweep again if the user decides to re-use the channel. The alternative is to prevent channels from being reused, but that might break some existing code.

channel.go
```diff
@@ -758,12 +777,16 @@ func (ch *Channel) Close() {
...
 	var channelClosed bool
+
 	ch.mutable.Lock()
 
 	if ch.mutable.l != nil {
 		ch.mutable.l.Close()
 	}
 
+	// Stop the idle connections timer.
+	ch.mutable.idleSweep.Stop()
```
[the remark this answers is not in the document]
orenkislev: `IdleSweep.started` will be false, so there won't be any side effects of calling it multiple times.

health_ext_test.go
orenkislev: Thank you! I will merge and rebase

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+package tchannel
+
+import (
+	"time"
+)
+
+const (
+	// defaultMaxIdleTime is the duration a connection is allowed to remain idle
+	// before it is automatically closed.
+	defaultMaxIdleTime = 3 * time.Minute
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// defaultIdleCheckInterval is the frequency in which the channel checks for
+	// idle connections. (Default: check is disabled)
+	defaultIdleCheckInterval = time.Duration(0)
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// IdleSweep controls a periodic task that looks for idle connections and clears
+// them from the peer list.
+// NOTE: This struct is not thread-safe on its own. Calls to Start() and Stop()
+// should be guarded by locking ch.mutable
+type IdleSweep struct {
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+	ch                *Channel
+	maxIdleTime       time.Duration
+	idleCheckInterval time.Duration
+	stopCh            chan struct{}
+	started           bool
+}
+
+// newIdleSweep starts a poller that checks for idle connections at given
+// intervals.
+func newIdleSweep(ch *Channel, opts *ChannelOptions) *IdleSweep {
```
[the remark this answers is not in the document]
orenkislev: I find it better to have an object even if it doesn't do anything, than to have to check for nil everywhere. Makes the code cleaner, and less prone to bugs. I don't feel too strongly about it, if you prefer nil checks.

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+	return is
+}
+
+// Start runs the goroutine responsible for checking idle connections.
+func (is *IdleSweep) Start() {
+	if is.started || is.idleCheckInterval <= time.Duration(0) {
+		return
+	}
+
+	is.ch.log.Info("Starting idle connections poller")
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+	go is.pollerLoop()
+}
+
+// Stop kills the poller checking for idle connections.
+func (is *IdleSweep) Stop() {
+	if !is.started {
+		return
+	}
+
+	is.ch.log.Info("Stopping idle connections poller")
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+
+	is.started = false
+	is.stopCh <- struct{}{}
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,137 @@
...
+		case <-is.stopCh:
+			ticker.Stop()
+			return
+		}
+	}
+}
+
+func (is *IdleSweep) checkIdleConnections() {
+	// Make a copy of the peer list to reduce the time we keep the lock.
+	rootPeers := is.ch.RootPeers().Copy()
```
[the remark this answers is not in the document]
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep_test.go
orenkislev: I agree, this was a tough one. The call to `checkIdleConnections` happens in a goroutine and we have no synchronization channel to know when it completed. There is additionally some small delay between disconnecting the client-side until the server-side receives and processes the disconnect, even though they are both happening in localhost.

I changed to use `testutils.WaitFor` but I'm open to other suggestions.

idle_sweep_test.go
orenkislev: Done

testutils/ticker.go
orenkislev: Done (in your branch)

idle_sweep_test.go
orenkislev: Done

idle_sweep.go
```diff
@@ -0,0 +1,122 @@
...
+	return is
+}
+
+// Start runs the goroutine responsible for checking idle connections.
+func (is *idleSweep) Start() {
+	if is.started || is.idleCheckInterval <= time.Duration(0) {
+		return
+	}
+
+	if is.maxIdleTime <= time.Duration(0) {
```
prashantv: this sort of validation should happen when creating the channel and return an error.

also, no need to type the constant here, you can do `if is.maxIdleTime <= 0`

channel.go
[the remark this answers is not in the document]
prashantv: Hmm I don't think we allow channels to be reused. `Serve` ensures the channel state is `ChannelClient` (which is the initial state), and we don't allow going backwards anywhere.

idle_sweep.go
prashantv: let's also add test coverage for this

idle_sweep.go
```diff
@@ -0,0 +1,122 @@
...
+	go is.pollerLoop()
+}
+
+// Stop kills the poller checking for idle connections.
+func (is *idleSweep) Stop() {
+	if !is.started {
+		return
+	}
+
+	is.ch.log.Info("Stopping idle connections poller.")
```
prashantv: no need for 2 logs here, just "Stopping" is enough, since nothing below blocks.

idle_sweep.go
```diff
@@ -0,0 +1,122 @@
...
+func (is *idleSweep) checkIdleConnections() {
+	now := is.ch.timeNow()
+
+	// Acquire the read lock and examine which connections are idle.
+	idleConnections := make([]*Connection, 0, 10)
+	is.ch.mutable.RLock()
+
+	for _, conn := range is.ch.mutable.conns {
+		idleTime := now.Sub(conn.getLastActivityTime())
+		if idleTime >= is.maxIdleTime {
```
prashantv: optional: can scope the `idleTime` just to the if,
```go
if idleTime := now.Sub(..); idleTime >= is.maxIdleTime {
  [...]
}
```

in general I prefer to scope variables to the minimum scope where they're needed

idle_sweep.go
```diff
@@ -0,0 +1,122 @@
...
+			idleConnections = append(idleConnections, conn)
+		}
+	}
+
+	is.ch.mutable.RUnlock()
```
prashantv: can you move the `RLock` and `RUnlock` right next to the if, we can treat that as a single block.

idle_sweep.go
```diff
@@ -0,0 +1,122 @@
...
+
+	for _, conn := range idleConnections {
+		is.ch.log.WithFields(
+			LogField{"remotePeer", conn.remotePeerInfo}).Info("Closing idle inbound connection.")
```
prashantv: can we log the last activity time and the idle time here. if we're calculating them again, we might want to double check that it really is idle just in case of a race.

idle_sweep.go
```diff
@@ -0,0 +1,122 @@
...
+		conn.close(LogField{"reason", "Idle connection closed"})
```
prashantv: I think when we close a connection, it doesn't immediately get removed from ch.mutable.conns, so we may end up closing multiple times. that should be safe, but just wanted to call that out.

we could instead only close connections that are active here. (suggest here to avoid holding the `RLock` for too long, and the filter of idle but not active is very unlikely, so probably won't get a benefit from doing it earlier)

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

channel.go
```diff
@@ -218,6 +229,12 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		LogField{"chID", chID},
 	)
 
+	if opts.IdleCheckInterval > 0 && opts.MaxIdleTime <= 0 {
+		logger.Warn("To enable automatically removing idle connections, you must " +
```
prashantv: this should return an error, no point allowing users to create a channel with invalid config.

Suggest making an unexported function on `ChannelOptions` that does this,
```go
func (o *ChannelOpts) validateIdleCheck() error {
  ...
}

```

that way we can just return an error from validation:
```go
if err := opts.validateIdleCheck(); err != nil {
  return nil, err
}
```

channel.go
```diff
@@ -269,6 +286,11 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	if opts.RelayHost != nil {
 		opts.RelayHost.SetChannel(ch)
 	}
+
+	// Start the idle connection timer.
+	ch.mutable.idleSweep = newIdleSweep(ch, opts)
+	ch.mutable.idleSweep.Start()
```
prashantv: no point separating the new and start, we can simplify by just having `startIdleSweep` and then we don't have the double-start problem

idle_sweep.go
```diff
@@ -0,0 +1,120 @@
...
+	).Info("Starting idle connections poller.")
+
+	is.stopCh = make(chan struct{})
+	is.started = true
+	go is.pollerLoop()
+}
+
+// Stop kills the poller checking for idle connections.
+func (is *idleSweep) Stop() {
+	if !is.started {
```
prashantv: we can combine `started` with the `stopCh`. on start, set `stopCh`, on `Stop`, if `stopCh == nil`, there's nothing running. once we `close(is.stopCh)`, we can set it to `nil`

idle_sweep.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+}
+
+func (is *idleSweep) checkIdleConnections() {
+	now := is.ch.timeNow()
+
+	// Acquire the read lock and examine which connections are idle.
+	idleConnections := make([]*Connection, 0, 10)
+
```
prashantv: let's remove this blank line as well, so the comment is with the block.

idle_sweep.go
```diff
@@ -0,0 +1,120 @@
...
+	for _, conn := range idleConnections {
+		// It's possible that the connection is already closed when we get here.
+		if !conn.IsActive() {
+			continue
+		}
+
+		is.ch.log.WithFields(
+			LogField{"remotePeer", conn.remotePeerInfo},
+			LogField{"lastActivityTime", conn.getLastActivityTime()}).Info("Closing idle inbound connection.")
```
prashantv: nit: move the `Info(..)` to the next line so fields are clearly separated from the message.

idle_sweep_test.go
```diff
@@ -0,0 +1,273 @@
...
+	for _, ch := range channels {
+
+		status = append(status,
+			fmt.Sprintf("%s: %d open", ch.PeerInfo().ProcessName, numConnections(ch)))
+	}
+
+	return strings.Join(status, ", ")
+}
+
+func waitForZeroConnections(t *testing.T, channels ...*Channel) bool {
```
prashantv: we could actually events here:
```
    // OnPeerStatusChanged is an optional callback that receives a notification
    // whenever the channel establishes a usable connection to a peer, or loses
    // a connection to a peer.
    OnPeerStatusChanged func(*Peer)
```

consider setting up events so we know when there's a connection change rather than waiting in a loop

channel.go
orenkislev: Done

channel.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep.go
orenkislev: Done

idle_sweep_test.go
orenkislev: Done

idle_sweep.go
prashantv: Ahh, fair enough, let's stick to this implementation. thanks

PR #666, opened 2017-11-07
Document and automate parts of the release process

prashantv:
We currently have an undocumented release process. Document the
different steps and try to automate parts.

The version bumping script is just one part of the process, in future
we'll want to automate the branch/tags as well as updating the release
on GitHub.

PR #672, opened 2017-11-27
Address CR comments on vbumper

prashantv:
Address comments from
https://github.com/uber/tchannel-go/pull/666

Keeping all commit logs in the `CHANGELOG`, since we still expect a person to edit the changelog to format them for users. The goal is to automate the mechanical parts.

### DIFF
--- a/scripts/vbumper/main.go
+++ b/scripts/vbumper/main.go
@@ -23,6 +23,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"flag"
 	"html/template"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


scripts/vbumper/main.go
```diff
@@ -156,11 +163,11 @@ func getChanges(prevVersion string) ([]string, error) {
 	newLines := make([]string, 0, len(lines))
 	// Each line is "[commit] [desc]", so ignore everything before the first space.
```
abhinav: outdated comment

prashantv: 👍

PR #673, opened 2017-11-27
Record the last activity timestamp in each connection

orenkislev:
This is related to #616 

We want to record the timestamp whenever a frame is written or read, so that we can measure how long the connection has been idle. Pings do not count as activity.


introspection.go
```diff
@@ -149,6 +149,7 @@ type ConnectionRuntimeState struct {
 	OutboundExchange ExchangeSetRuntimeState `json:"outboundExchange"`
 	Relayer          RelayerRuntimeState     `json:"relayer"`
 	HealthChecks     []bool                  `json:"healthChecks,omitempty"`
+	LastActivity     time.Time               `json:"last_activity"`
```
prashantv: nit: `json:"lastActivity"` (camel case to be consistent with the other fields)

connection.go
```diff
@@ -721,6 +728,18 @@ func (c *Connection) writeFrames(_ uint32) {
 	}
 }
 
+// updateLastActivity marks when the last message was received/sent on the channel.
+// This is used for monitoring idle connections and timing them out.
+func (c *Connection) updateLastActivity(frame *Frame) {
+	// Pings are ignored for last activity.
+	switch frame.Header.messageType {
+	case messageTypeCallReq, messageTypeCallReqContinue, messageTypeCallRes, messageTypeCallResContinue, messageTypeError:
+		c.stateMut.Lock()
```
prashantv: little worried about lock overhead here, we could consider an `atomic.Int64` and store `t.UnixNano()` which gives us enough precision and should be cheaper (a single atomic operation vs the cheapest for a lock which is two atomic ops)

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
 		clientCh.Close()
 	})
 }
+
+// getConnection returns the introspection result for the unique inbound or
+// outbound connection. An assert will be raised if there is more than one
+// connection of the given type.
+func getConnection(t *testing.T, ch *Channel, direction int) *ConnectionRuntimeState {
+	state := ch.IntrospectState(&IntrospectionOptions{
+		IncludeEmptyPeers: false,
```
prashantv: no need to specify this if it's the default, in fact you should be able to do `ch.IntrospectState(nil)` to get the defaults.

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+
+	for _, peer := range state.RootPeers {
+		var connections []ConnectionRuntimeState
+		if direction == inbound {
+			connections = peer.InboundConnections
+		} else {
+			connections = peer.OutboundConnections
+		}
+
+		assert.True(t, len(connections) <= 1, "Too many connections found")
```
prashantv: use `assert.Len` instead of `assert.True` so you get more useful information on a failure

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+		if len(connections) == 1 {
```
prashantv: no need to do the same check twice, you can use the result of `assert` methods, they're true if the assertion succeeded

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+			return &connections[0]
+		}
+	}
+
+	assert.Fail(t, "No connections found")
+	return nil
```
prashantv: maybe make this `panic("not reachable")`

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+				relayInbound := getConnection(t, ts.Relay(), inbound)
+				relayOutbound := getConnection(t, ts.Relay(), outbound)
+				assert.Equal(t, clock.Now(), relayInbound.LastActivity)
+				assert.Equal(t, clock.Now(), relayOutbound.LastActivity)
+			}
+
+			clock.Elapse(1 * time.Second)
+		}
+	})
+}
```
prashantv: suggest verifying the last activity set from read/write separately. e.g.,
 - set up a handler `testutils.RegisterEcho` with a function that will block until a channel is closed
 - send a call, verify that last activity is updated everywhere
 - elapse time
 - unblock the call, which should result in a response
 - verify last activity is updated to new time

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+
+func TestLastActivityTimePings(t *testing.T) {
+	initialTime := time.Date(2017, 11, 27, 21, 0, 0, 0, time.UTC)
+	clock := testutils.NewStubClock(initialTime)
+
+	// Pings are not re-transmitted by the relay.
```
prashantv: is that relevant to the the following lines of code?

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+	sopts := testutils.NewOpts().SetTimeNow(clock.Now)
+	ctx, cancel := NewContext(testutils.Timeout(100 * time.Millisecond))
+	defer cancel()
+
+	testutils.WithTestServer(t, sopts, func(ts *testutils.TestServer) {
+		copts := testutils.NewOpts().SetTimeNow(clock.Now)
+		client := ts.NewClient(copts)
```
prashantv: can reuse the same `sopts`, just call them `opts` since you can reuse them.

connection_test.go
```diff
@@ -1079,3 +1084,104 @@ func TestContextCanceledOnTCPClose(t *testing.T) {
...
+			assert.Equal(t, timeAtStart, clientConn.LastActivity)
+
+			// Relays do not pass pings on to the server.
+			if ts.HasRelay() {
+				relayInbound := getConnection(t, ts.Relay(), inbound)
+				assert.Equal(t, timeAtStart, relayInbound.LastActivity)
+			} else {
+				serverConn := getConnection(t, ts.Server(), inbound)
+				assert.Equal(t, timeAtStart, serverConn.LastActivity)
+			}
```
prashantv: why the `else`, we can validate this regardless right?

connection.go
orenkislev: Done

connection_test.go
orenkislev: Done

connection_test.go
orenkislev: Not the right choice here, because len(connections) == 0 is also valid. The code here also handles the case where a relay has 2 peers, and we're only interested in the inbound direction.

connection_test.go
orenkislev: I'm only interested in the len() == 1 case

connection_test.go
orenkislev: It is reachable, if the test has a problem and no connections were established.

connection_test.go
orenkislev: Done

connection_test.go
orenkislev: Done

connection_test.go
orenkislev: Done

connection_test.go
orenkislev: Done

introspection.go
orenkislev: Done

connection.go
```diff
@@ -196,6 +196,10 @@ type Connection struct {
 	healthCheckQuit    context.CancelFunc
 	healthCheckDone    chan struct{}
 	healthCheckHistory *healthHistory
+
+	// lastActivity is used to track how long the connection has been idle.
+	// (unix time, nano)
+	lastActivity *atomic.Int64
```
prashantv: no reason this has to be a pointer, can you do atomic.Int64 here, and dereference when you create it using atomic.NewInt64

connection_test.go
prashantv: i see, in that case can you add the list of connections found in the error message.
```
assert.True(t, len(connections) <= 1, "Too many connections found: %+v", connections)
```

connection_test.go
prashantv: is it? shouldn't `assert.Fail` stop execution straight away by killing the current goroutine?

connection_test.go
prashantv: makes sense, 👍

connection_test.go
prashantv: Nice, the test looks great!

introspection.go
```diff
@@ -348,6 +349,7 @@ func (c *Connection) IntrospectState(opts *IntrospectionOptions) ConnectionRunti
 		InboundExchange:  c.inbound.IntrospectState(opts),
 		OutboundExchange: c.outbound.IntrospectState(opts),
 		HealthChecks:     c.healthCheckHistory.asBools(),
+		LastActivity:     c.lastActivity.Load(),
```
prashantv: in the introspection, can we make this a `time.Time`, which will probably be easier for humans to read? Or could have 2 separate fields (one as nanos, one as human readable).

That will also allow the tests to compare the time directly.

connection_test.go
orenkislev: Done

connection_test.go
orenkislev: Changed it to FailNow()

introspection.go
orenkislev: That was my first intuition as well, but it makes things weird because the reconstructed time.Time() doesn't have a timezone (Location), so direct comparison with a time.Date created in the test failed, even though they are identical.

connection.go
orenkislev: Done

orenkislev:
PTAL

PR #676, opened 2018-01-30
stats: Add tally reporter to emit tagged metrics

prashantv:
This is an open-sourcing of the tally StatsReporter which will emit
tagged metrics rather than statsd metrics.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,14 +1,16 @@
-hash: d6093eb1a6d4c1eaa9f2ad996ebf07748b0b3a0413f4ab9b9435df1d255567a5
-updated: 2017-09-25T17:19:51.449837394-07:00
+hash: b75ed7f632cfd185b6a8f8bc037b09d32e292f908af6c1e257e444e17a3a39ef
+updated: 2018-01-30T12:06:17.405736-08:00
 imports:
 - name: github.com/apache/thrift
   version: b2a4d4ae21c789b689dd162deb819665567f481c
   subpackages:
   - lib/go/thrift
 - name: github.com/cactus/go-statsd-client
-  version: 91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d
+  version: 138b925ccdf617776955904ba7759fce64406cec
   subpackages:
   - statsd
+- name: github.com/facebookgo/clock
+  version: 600d898af40aa09a7a93ecb9265d87b0504b6f03
 - name: github.com/opentracing/opentracing-go
   version: 1949ddbfd147afd4d964a9f00b24eb291e0e7c38
   subpackages:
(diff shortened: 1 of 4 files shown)


glide.lock
```diff
@@ -1,14 +1,16 @@
-hash: d6093eb1a6d4c1eaa9f2ad996ebf07748b0b3a0413f4ab9b9435df1d255567a5
-updated: 2017-09-25T17:19:51.449837394-07:00
+hash: b75ed7f632cfd185b6a8f8bc037b09d32e292f908af6c1e257e444e17a3a39ef
```
kriskowal: tip: I’ve found that it’s easier to rebase if I keep glide.lock updates in their own commit. Easy to remove the commit when I rebase, then recreate with `glide up`, instead of dealing with glide.lock merge conflicts. More relevant on a more active repository.

stats/tally.go
```diff
@@ -0,0 +1,188 @@
+package stats
```
kriskowal: Might be worth stuttering this package name, e.g., `tchanneltally`.

stats/tally.go
```diff
@@ -0,0 +1,188 @@
...
+
+	counters map[string]tally.Counter
+	gauges   map[string]tally.Gauge
+	timers   map[string]tally.Timer
+}
+
+// NewTallyReporter takes a tally.Scope and wraps it so it ca be used as a
+// StatsReporter. The list of metrics emitted is documented on:
+// https://tchannel.readthedocs.io/en/latest/metrics/
+// The tags emitted are similar to Muttley/YARPC, the tags emitted are:
```
kriskowal: Public mention of Muttley.

stats/tally.go
```diff
@@ -0,0 +1,188 @@
...
+func (w *wrapper) RecordTimer(name string, tags map[string]string, d time.Duration) {
+	ts := w.getTaggedScope(tags)
+	ts.getTimer(name).Record(d)
+}
+
+func (w *wrapper) getTaggedScope(tags map[string]string) *taggedScope {
+	kt := convertTags(tags)
+
+	w.RLock()
+	ts, ok := w.byTags[kt]
```
kriskowal: In YARPC’s observability layer, we often digest the tags into a null-delimited string. You know the performance of struct key map lookup better than me. This is on the hot path?

glide.lock
prashantv: Yep, agreed. Since this repo isn't really updated, this hasn't been an issue.

stats/tally.go
prashantv: This is adding to the existing tchannel stats package:
https://godoc.org/github.com/uber/tchannel-go/stats

stats/tally.go
prashantv: Updated

stats/tally.go
prashantv: Good question -- this is definitely on the hot path (since we emit multiple stats per RPC).

A single string lookup is faster than using a struct with 4 strings. Just the lookup has a huge performance difference (the struct has 4 strings, and the string is the 4 strings concatenated with a null byte separator):
```
string-lookup: 24.6 ns/op
struct-lookup: 84.4 ns/op
```

However, this is not enough to look at, as using a single string requires building that string. Let's add the cost of building a string (without allocating) by reusing a shared buffer, and then do the lookup:
```
string-build-lookup: 63.8
```

This may be called from multiple goroutines, so we can't reuse a single buffer. Once we add buffer pooling, the cost difference is negligible, and the complexity of pooling (and risk of issues) is probably not worth the optimization.

PR #678, opened 2018-01-31
Expose relay span setting in relayer

Yu-Xie:
Expose SetRelaySpan to CallFrame interface. It allows relaying services to report their own spans.

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -24,6 +24,10 @@
 // backwards-compatibility guarantee.
 package relay
 
+import (
+	"github.com/opentracing/opentracing-go"
+)
+
 // CallFrame is an interface that abstracts access to the call req frame.
 type CallFrame interface {
 	// Caller is the name of the originating service.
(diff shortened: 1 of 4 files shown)


prashantv:
@Yu-Xie I was chatting with @kriskowal and @willhug about this change, and we realized this probably won't work since tracing is propagated in headers. For backwards compatibility, it's also added to the call frame, but the clients will use the headers when available.

The headers are in arg2, and there's no efficient way to read/write arg2.

You might have to use TChannel as a normal client (using the arg reader + writer) to get this working.

Yu-Xie:
@prashantv @kriskowal, it looks when tracing information is in call frame, it will be used from there ([extractInboundSpan](https://github.com/uber/tchannel-go/blob/dev/tracing.go#L193)), and the headers will only be used for baggages. Otherwise the headers are used ([ExtractInboundSpan](https://github.com/uber/tchannel-go/blob/dev/tracing.go#L225)).

prashantv:
@Yu-Xie I'm not sure if that's guaranteed across implementations -- E.g., what about Java, Node and Python?

Yu-Xie:
@prashantv Yeah, that's right. Going to close this PR.

PR #681, opened 2018-02-16
Allow blackholing requests

AlexAPB:
This enables us to blackhole requests and release unnecessary long-lived goroutines.

Confirmed with pprof and YARPC that the only goroutines left running are for each connection.

Relates to yarpc/yarpc-go#1436.

### DIFF
--- a/inbound.go
+++ b/inbound.go
@@ -93,7 +93,6 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 	if response.span != nil {
 		mex.ctx = opentracing.ContextWithSpan(mex.ctx, response.span)
 	}
-	response.blackhole = make(chan struct{})
 	response.mex = mex
 	response.conn = c
 	response.cancel = cancel
@@ -195,10 +194,6 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
 			// TODO: move the cancel to the parent context at connnection level
 			call.response.cancel()
 			call.mex.inboundExpired()
-		case <-call.response.blackhole:
-			// close go routine, release resources
-			call.mex.inboundExpired()
-			return
 		}
 	}()
 
@@ -334,8 +329,6 @@ type InboundCallResponse struct {
 	span             opentracing.Span
 	statsReporter    StatsReporter
 	commonStatsTags  map[string]string
-	// blackhole is a channel that gets closed when the repsonse should be blackholed.
-	blackhole chan struct{}
 }
 
 // SendSystemError returns a system error response to the peer.  The call is considered
@@ -371,7 +364,7 @@ func (response *InboundCallResponse) SetApplicationError() error {
 // Blackhole indicates that there should be no response and provides
 // an opportunity to clean up resources.
 func (response *InboundCallResponse) Blackhole() {
-	close(response.blackhole)
+	response.cancel()
 }
 
 // Arg2Writer returns a WriteCloser that can be used to write the second argument.
--- a/inbound.go
+++ b/inbound.go
@@ -93,6 +93,7 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 	if response.span != nil {
 		mex.ctx = opentracing.ContextWithSpan(mex.ctx, response.span)
 	}
+	response.blackhole = make(chan struct{})
 	response.mex = mex
 	response.conn = c
 	response.cancel = cancel
@@ -194,6 +195,10 @@ func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, fram
 			// TODO: move the cancel to the parent context at connnection level
 			call.response.cancel()
 			call.mex.inboundExpired()
+		case <-call.response.blackhole:
+			// close go routine, release resources
+			call.mex.inboundExpired()
+			return
 		}
 	}()
 
@@ -329,6 +334,8 @@ type InboundCallResponse struct {
 	span             opentracing.Span
 	statsReporter    StatsReporter
 	commonStatsTags  map[string]string
+	// blackhole is a channel that gets closed when the repsonse should be blackholed.
+	blackhole chan struct{}
 }
 
 // SendSystemError returns a system error response to the peer.  The call is considered
@@ -359,6 +366,12 @@ func (response *InboundCallResponse) SetApplicationError() error {
 	}
 	response.applicationError = true
 	return nil
+}
+
+// Blackhole indicates that there should be no response and provides
+// an opportunity to clean up resources.
+func (response *InboundCallResponse) Blackhole() {
+	close(response.blackhole)
 }
 
 // Arg2Writer returns a WriteCloser that can be used to write the second argument.
--- a/inbound.go
+++ b/inbound.go
@@ -361,8 +361,9 @@ func (response *InboundCallResponse) SetApplicationError() error {
 	return nil
 }
 
-// Blackhole indicates that there should be no response and provides
-// an opportunity to clean up resources.
+// Blackhole indicates no response will be sent, and cleans up any resources
+// associated with this request. This allows for services to trigger a timeout in
+// clients without holding on to any goroutines on the server.
 func (response *InboundCallResponse) Blackhole() {
 	response.cancel()
 }
--- a/inbound_test.go
+++ b/inbound_test.go
@@ -142,3 +142,27 @@ func TestInboundConnection_CallOptions(t *testing.T) {
 		require.NoError(t, err, "Call through proxy failed")
 	})
 }
+
+func TestBlackhole(t *testing.T) {
+	testutils.WithTestServer(t, nil, func(tServer *testutils.TestServer) {
+		serviceName := tServer.ServiceName()
+
+		server := tServer.NewServer(testutils.NewOpts())
+		defer server.Close()
+
+		subCh := server.GetSubChannel(serviceName)
+		subCh.SetHandler(HandlerFunc(func(ctx context.Context, inbound *InboundCall) {
+			// blackhole all requests
+			inbound.Response().Blackhole()
+		}))
+
+		clientCh := tServer.NewClient(testutils.NewOpts())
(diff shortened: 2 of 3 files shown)


AlexAPB:
I used `response.cancel()` as suggested and confirmed again with pprof that we're not keeping goroutines running after blackholing the request.

I'd rather use this method than create a new error with `SendSystemError`. I think this is a bit more straightforward and would otherwise cause `SendSystemError` to be a misnomer.

I've also added a test, but it relies on the context timing out so its essentially a sleep for 10 milliseconds. I'm not sure what your thoughts on this are @prashantv, but I think that this is the most reasonable way for me to test the change.

mex.go
```diff
@@ -390,7 +390,7 @@ func (mexset *messageExchangeSet) removeExchange(msgID uint32) {
 // will write to the send channel.
 func (mexset *messageExchangeSet) expireExchange(msgID uint32) {
 	mexset.log.Debugf(
-		"Removing %s message exchange %d due to timeout or cancelation",
+		"Removing %s message exchange %d due to timeout, cancelation or blackhole",
```
breerly: cancellation*

inbound.go
```diff
@@ -361,6 +361,12 @@ func (response *InboundCallResponse) SetApplicationError() error {
 	return nil
 }
 
+// Blackhole indicates that there should be no response and provides
+// an opportunity to clean up resources.
+func (response *InboundCallResponse) Blackhole() {
```
breerly: Simpler than I thought it was going to be.

mex.go
kriskowal: Indeed. Corroboration for consistency (since technically both spellings are accepted):

```
tchannel-go ❯ git grep -i cancellation
CHANGELOG.md:* Fix context cancellation not cancelling Dial attempts (#541)
inbound.go:     // TODO(prashant): This is an expensive way to check for cancellation. Use a heap for timeouts.
inbound.go:                             call.log.Debugf("Wait for timeout/cancellation interrupted by error: %v", call.mex.errCh.err)
mex.go:// It returns any existing errors (timeout, cancellation, connection errors).
mex.go: // 1. Timeouts/cancellation (mex.ctx errors)
mex.go: // 1. Timeouts/cancellation (mex.ctx errors)

tchannel-go ❯ git grep -i cancelation
mex.go:         "Removing %s message exchange %d due to timeout or cancelation",
```

inbound_test.go
```diff
@@ -142,3 +142,27 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+		subCh := server.GetSubChannel(serviceName)
+		subCh.SetHandler(HandlerFunc(func(ctx context.Context, inbound *InboundCall) {
+			// blackhole all requests
+			inbound.Response().Blackhole()
+		}))
+
+		clientCh := tServer.NewClient(testutils.NewOpts())
+		defer clientCh.Close()
+
+		ctx, cancel := NewContext(time.Millisecond * 10)
```
prashantv: nit: use `testutils.Timeout` here

inbound_test.go
```diff
@@ -142,3 +142,27 @@ func TestInboundConnection_CallOptions(t *testing.T) {
 		require.NoError(t, err, "Call through proxy failed")
 	})
 }
+
+func TestBlackhole(t *testing.T) {
+	testutils.WithTestServer(t, nil, func(tServer *testutils.TestServer) {
+		serviceName := tServer.ServiceName()
+
+		server := tServer.NewServer(testutils.NewOpts())
```
prashantv: why are you creating a new server here? the `tServer` can have methods registered directly

inbound_test.go
```diff
@@ -142,3 +142,27 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+		defer server.Close()
+
+		subCh := server.GetSubChannel(serviceName)
+		subCh.SetHandler(HandlerFunc(func(ctx context.Context, inbound *InboundCall) {
+			// blackhole all requests
+			inbound.Response().Blackhole()
+		}))
+
+		clientCh := tServer.NewClient(testutils.NewOpts())
```
prashantv: nit: pass in `nil` for default options to be consistent with other tests

inbound_test.go
```diff
@@ -142,3 +142,27 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+	testutils.WithTestServer(t, nil, func(tServer *testutils.TestServer) {
+		serviceName := tServer.ServiceName()
+
+		server := tServer.NewServer(testutils.NewOpts())
+		defer server.Close()
+
+		subCh := server.GetSubChannel(serviceName)
+		subCh.SetHandler(HandlerFunc(func(ctx context.Context, inbound *InboundCall) {
+			// blackhole all requests
+			inbound.Response().Blackhole()
```
prashantv: If you replace the implementation of `Blackhole` with a `time.Sleep`, the test will still pass.
```
diff --git a/inbound.go b/inbound.go
index a7f7739b..605bc52f 100644
--- a/inbound.go
+++ b/inbound.go
@@ -364,7 +364,8 @@ func (response *InboundCallResponse) SetApplicationError() error {
 // Blackhole indicates that there should be no response and provides
 // an opportunity to clean up resources.
 func (response *InboundCallResponse) Blackhole() {
-       response.cancel()
+       time.Sleep(200 * time.Millisecond)
+       //response.cancel()
 }
```

```
> go test -v -run TestBlackhole
=== RUN   TestBlackhole
--- PASS: TestBlackhole (0.42s)
PASS
ok  	github.com/uber/tchannel-go	0.439s
```

That means the test isn't covering that the blackhole implementation is more efficient than a non-blackhole implementation.

Some suggests on how the test can actually validate it:
 * Before the `Blackhole` call, if you [introspect](https://godoc.org/github.com/uber/tchannel-go#Channel.IntrospectState), you should see that there's a pending message exchange
https://godoc.org/github.com/uber/tchannel-go#ExchangeSetRuntimeState

 * After the `Blackhole` call, there should be no exchanges.

Instead of using a short timeout, you could actually:
 - make a call with a huge timeout (e.g., 1 hour), so we know the exchange isn't being cleared due to timeout
 - after the `Blackhole` command, we can cancel the client context, so the client returns

inbound.go
```diff
@@ -361,6 +361,12 @@ func (response *InboundCallResponse) SetApplicationError() error {
 	return nil
 }
 
+// Blackhole indicates that there should be no response and provides
```
prashantv: more context here,
```
// Blackhole indicates no response will be sent, and cleans up any resources associated with this request. This allows for services to trigger a timeout in clients without holding on to any goroutines on the server.
```

inbound_test.go
AlexAPB: yup, my bad. I overlooked the `register` function completely.

AlexAPB:
Updated with your suggestions @prashantv. Since the message exchange cleanup happens in a separate goroutine, I still found the need to add `time.Sleep` to give time for clean up. Tests pass with a 1ms timeout, but I upped it to 10ms to reduce potential flakiness.

inbound_test.go
```diff
@@ -142,3 +142,38 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+			require.NotNil(t, c)
+
+			state := c.IntrospectState(&IntrospectionOptions{})
+			require.Equal(t, 1, state.InboundExchange.Count, "expected exactly one inbound exchange")
+
+			// blackhole request
+			inbound.Response().Blackhole()
+
+			// give time for exchange to cleanup
+			time.Sleep(10 * time.Millisecond)
```
prashantv: instead of sleep, take a look at `testutils.WaitFor` -- it's a lot less flaky for testing since it'll sleep a small amount initially, check whether what we're waiting for is true, if not, sleep a little longer, and repeat.

The issue with a single `time.Sleep` is that it could finish in 1ms, in which case you're wasting 9ms, or it could take 11ms, in which case the test becomes flaky and fails. The only way to avoid flakiness is to use a huge value like a second or more, but then you're likely wasting 900ms for each run just doing nothing.

`WaitFor` is a compromise that lets you sleep (since some things can't be tested without waiting for some background work to be scheduled), without making the test slower or flaky.

inbound_test.go
```diff
@@ -142,3 +142,38 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+
+			state = c.IntrospectState(&IntrospectionOptions{})
+			require.Equal(t, 0, state.InboundExchange.Count, "expected no inbound exchanges")
+		}), handlerName)
+
+		clientCh := server.NewClient(nil)
+		defer clientCh.Close()
+
+		_, _, _, err := raw.Call(ctx, clientCh, server.HostPort(), serviceName, handlerName, nil, nil)
+		require.Error(t, err, "expected to timeout")
```
prashantv: I think you're actually expecting a cancelled error, since you call `cancel`.

Can you also add a `assert.Equal(t, ErrCodeCanceled, GetSystemErrorCode(err))` (or similar) to make sure we're getting an error for the right reason.

inbound_test.go
```diff
@@ -142,3 +142,44 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+		}), handlerName)
+
+		clientCh := server.NewClient(nil)
+		defer clientCh.Close()
+
+		_, _, _, err := raw.Call(ctx, clientCh, server.HostPort(), serviceName, handlerName, nil, nil)
+		require.Error(t, err, "expected call error")
+
+		errCode := GetSystemErrorCode(err)
+		assert.Equal(t, ErrCodeCancelled, errCode, "expected cancelled error code, got: %q", errCode)
```
prashantv: no need for "got %v", `assert` will print the value it expected and what it got. can just make error message something like "got unexpected error code"

AlexAPB: As discussed in person, this would output the `byte` value (eg. `0x02`) since `SystemErrCode` is an alias of `byte`. I'll update with a comment to indicate why I wrote it out this way.

PR #683, opened 2018-02-17
Ensure outbound tracing headers take precedence

### LINKED ISSUE #682: InjectOutboundSpan doesn't Override Tracing Header
Not sure if it is expected behavior or a bug, but [`InjectOutboundSpan`](https://github.com/uber/tchannel-go/blob/dev/tracing.go#L172) would not inject if the input parameter `headers` already contains tracing information using the same tracing header.
[tracing.go#L187](https://github.com/uber/tchannel-go/blob/dev/tracing.go#L187)

Currently to inject tracing header correctly to outbound headers, we need to do the following:
```
delete(headers, "XX-Tracing-Header")
headers := tchannel.InjectOutboundSpan(outboundCall.Response(), headers)
```

yurishkuro:
Resolves #682 

Signed-off-by: Yuri Shkuro <ys@uber.com>

cc @Yu-Xie

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -185,7 +185,11 @@ func InjectOutboundSpan(response *OutboundCallResponse, headers map[string]strin
 		return headers // Tracer did not add any tracing headers, so return the original map
 	}
 	for k, v := range headers {
-		newHeaders[k] = v
+		// Some applications propagate all inbound application headers to outbound calls (issue #682).
+		// If those headers include tracing headers we want to make sure to keep the new tracing headers.
+		if _, ok := newHeaders[k]; !ok {
+			newHeaders[k] = v
+		}
 	}
 	return newHeaders
 }
--- a/tracing_test.go
+++ b/tracing_test.go
@@ -22,10 +22,10 @@ package tchannel_test
 
 import (
 	"sync"
-	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/uber-go/atomic"
 	. "github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/json"
 	"github.com/uber/tchannel-go/testutils"
@@ -69,7 +69,7 @@ func TestTracingSpanAttributes(t *testing.T) {
 			customAppHeaderKey           = "futurama"
 			customAppHeaderExpectedValue = "simpsons"
 		)
-		var customAppHeaderValue atomic.Value
+		var customAppHeaderValue atomic.String
 		// Register JSON handler
 		jsonHandler := &JSONHandler{
 			TraceHandler: testtracing.TraceHandler{Ch: ch},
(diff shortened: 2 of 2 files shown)


yurishkuro:
keep forgetting that `dev` is the trunk branch

tracing.go
```diff
@@ -185,7 +185,11 @@ func InjectOutboundSpan(response *OutboundCallResponse, headers map[string]strin
 		return headers // Tracer did not add any tracing headers, so return the original map
 	}
 	for k, v := range headers {
-		newHeaders[k] = v
+		// Some applications propagate all inbound application headers to outbound calls (issue #682).
+		// If those headers include tracing headers we want to make sure to keep the new tracing headers.
+		if _, ok := newHeaders[k]; !ok {
+			newHeaders[k] = v
```
prashantv: Do you know why we didn't get coverage from the added test here?

yurishkuro: Don't know. When I run this locally it shows the line as covered:
```
$ go test -v -run TestTracingSpanAttributes -coverprofile cover.out
=== RUN   TestTracingSpanAttributes
--- PASS: TestTracingSpanAttributes (0.00s)
PASS
coverage: 48.8% of statements
ok  	github.com/uber/tchannel-go	0.042s

$ go tool cover -html=cover.out
```

image

I added another check to ensure that non-tracing headers are propagated.

tracing_test.go
```diff
@@ -60,23 +65,47 @@ func TestTracingSpanAttributes(t *testing.T) {
 		DisableRelay:   true,
 	}
 	WithVerifiedServer(t, opts, func(ch *Channel, hostPort string) {
+		const (
+			customAppHeaderKey           = "futurama"
+			customAppHeaderExpectedValue = "simpsons"
+		)
+		var customAppHeaderValue atomic.Value
```
prashantv: we use `uber-go/atomic` in tchannel, which has a `atomic.String` that you could use here

tracing_test.go
```diff
@@ -60,23 +65,47 @@ func TestTracingSpanAttributes(t *testing.T) {
...
 		span := ch.Tracer().StartSpan("client")
 		span.SetBaggageItem(testtracing.BaggageKey, testtracing.BaggageValue)
 		ctx := opentracing.ContextWithSpan(context.Background(), span)
 		root := new(testtracing.TracingResponse).ObserveSpan(ctx)
 
+		// Pretend that the client propagated tracing headers from upstream call, and test that the outbound call
+		// will override them (https://github.com/uber/tchannel-go/issues/682).
+		headers := make(map[string]string)
+		assert.NoError(t, tracer.Inject(span.Context(), opentracing.TextMap, opentracing.TextMapCarrier(headers)))
+		tracingHeaders := map[string]string{
```
prashantv: should these be named the opposite? this seems like `headers`, while the thing named `headers` seems more like `tracingHeaders`

tracing_test.go
```diff
@@ -60,23 +65,47 @@ func TestTracingSpanAttributes(t *testing.T) {
...
+			customAppHeaderKey: customAppHeaderExpectedValue,
+		}
+		for k, v := range headers {
+			tracingHeaders["$tracing$"+k] = v
```
prashantv: can you set the value to something incorrect, like `dummy` so we know that it was replaced

Yu-Xie:
Thanks Yuri for fixing this! Do we have a plan to release?

prashantv:
@Yu-Xie This is now released in the latest release:
https://github.com/uber/tchannel-go/releases/tag/v1.10.0

Yu-Xie:
Thanks @prashantv

PR #690, opened 2018-04-02
relay: Verify no use-after-free in timer pool

prashantv:
Ensure there are no use-after-free issues with the relay timer pool by
adding an option to disable pooling for extra verification.

For relay-only tests, run them twice, with and without the verification
to ensure that there are no issues.

Verified than an erroneous Release will cause issues:
```
=== RUN   TestRelayConcurrentNewConnectionAttempts
panic: Released timer cannot be used

goroutine 9332 [running]:
github.com/uber/tchannel-go.(*relayTimer).verifyNotReleased(0xc42046f230)
        /Users/prashant/gocode/src/github.com/uber/tchannel-go/relay_timer_pool.go:147 +0x50
github.com/uber/tchannel-go.(*relayTimer).Stop(0xc42046f230, 0xc420361ef0)
        /Users/prashant/gocode/src/github.com/uber/tchannel-go/relay_timer_pool.go:127 +0x2b
github.com/uber/tchannel-go.(*relayItems).Delete(0xc4200ac440, 0xc400000002, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/prashant/gocode/src/github.com/uber/tchannel-go/relay.go:120 +0x195
github.com/uber/tchannel-go.(*relayItems).Entomb.func1()
        /Users/prashant/gocode/src/github.com/uber/tchannel-go/relay.go:154 +0x31
```

### DIFF
--- a/relay_timer_pool.go
+++ b/relay_timer_pool.go
@@ -38,7 +38,8 @@ type relayTimer struct {
 	pool  *relayTimerPool // const
 	timer *time.Timer     // const
 
-	active bool // mutated on Start/Stop
+	active   bool // mutated on Start/Stop
+	released bool // mutated on Get/Release.
 
 	// Per-timer parameters passed back when the timer is triggered.
 	items        *relayItems
@@ -47,6 +48,7 @@ type relayTimer struct {
 }
 
 func (rt *relayTimer) OnTimer() {
+	rt.verifyNotReleased()
 	items, id, isOriginator := rt.items, rt.id, rt.isOriginator
 	rt.markTimerInactive()
 	rt.pool.trigger(items, id, isOriginator)
@@ -64,6 +66,7 @@ func newRelayTimerPool(trigger relayTimerTrigger, verify bool) *relayTimerPool {
 func (tp *relayTimerPool) Get() *relayTimer {
 	timer, ok := tp.pool.Get().(*relayTimer)
 	if ok {
+		timer.released = false
 		return timer
 	}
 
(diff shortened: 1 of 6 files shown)


relay_timer_pool.go
```diff
@@ -51,9 +52,10 @@ func (rt *relayTimer) OnTimer() {
 	rt.pool.trigger(items, id, isOriginator)
 }
 
-func newRelayTimerPool(trigger relayTimerTrigger) *relayTimerPool {
+func newRelayTimerPool(trigger relayTimerTrigger, verifyPool bool) *relayTimerPool {
 	return &relayTimerPool{
-		trigger: trigger,
+		trigger:    trigger,
+		verifyPool: verifyPool,
```
akshayjshah: nit: since this is the pool struct, name might be nicer as just `verify` or `shouldVerify`.

relay_timer_pool.go
```diff
@@ -81,6 +83,11 @@ func (tp *relayTimerPool) Get() *relayTimer {
 
 // Put returns a relayTimer back to the pool.
 func (tp *relayTimerPool) Put(rt *relayTimer) {
+	if tp.verifyPool {
+		// If we are trying to verify correct pool behavior, then we don't release
+		// the timer, and instead ensure no methods are called after being released.
+		return
+	}
```
akshayjshah: Correct me if I'm wrong: this approach relies on the fact that most of the relay timer's methods assume that they'll only be called once, so not returning timers to the pool and using them concurrently will trigger races. That's not true of the `Stop` method, though - should we also add some safety there?

relay_timer_pool.go
prashantv: 👍 Renamed to `verify`

relay_timer_pool.go
prashantv: Oh wow, I completely missed a commit when making this branch off the previous change!

There's a `verifyNotReleased` method which is called by all the exported methods and `OnTimer` which will verify `!released`. Adding the commit now, sorry about that!

PR #691, opened 2018-04-02
context: If parent has deadline, and timeout is 0, use WithCancel

### LINKED ISSUE #685: context builder should not use create a timeout 0 context when parent context is provided
In the current ContextBuilder implementation, providing a `Parent` context but not `Timeout` options will result in a 0 default value in result `Context`.
```
ctx := context.WithTimeout(context.Background(), time.Second)
ctxBuilder := &tchannel.ContextBuilder{
    ParentContext: ctx,
}
// ctxBuilder.Build will return a context with timeout 0
```

prashantv:
Currently, if no timeout is passed, we create a context with 0 timeout.
However, in cases where the parent has a timeout, we shouldn't force
users to specify another timeout.

Fixes #685.

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -303,7 +303,6 @@ func TestContextInheritParentTimeout(t *testing.T) {
 	deadline, ok := ctx.Deadline()
 	require.True(t, ok, "Missing deadline")
 
-	if deadline.Before(deadlineAfter) {
-		t.Fatalf("Expected deadline to be after %v, got %v", deadlineAfter, deadline)
-	}
+	assert.False(t, deadline.Before(deadlineAfter),
+		"Expected deadline to be after %v, got %v", deadlineAfter, deadline)
 }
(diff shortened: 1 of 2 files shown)


context_test.go
```diff
@@ -286,3 +287,23 @@ func TestContextWrapChild(t *testing.T) {
...
+	}
+	ctx, cancel := ctxBuilder.Build()
+	defer cancel()
+
+	// Ensure deadline is in the future
+	deadline, ok := ctx.Deadline()
+	require.True(t, ok, "Missing deadline")
+
+	if deadline.Before(deadlineAfter) {
+		t.Fatalf("Expected deadline to be after %v, got %v", deadlineAfter, deadline)
```
AllenLuUber: Optionally.
`assert.False(t, deadline.Before(deadlineAfter), "Expected deadline to be after %v, got %v", deadlineAfter, deadline)`?

prashantv: Done

PR #710, opened 2018-07-23
change NewChannel method and channel statsTags map conncurrent

1046102779:


1046102779:
ok, thks. recently i want to tracing systems such as OpenTracing , basictracer-go, appdashing, jaeger etc. jaeger uses tchannel-go rpc framework. thus I want to use etcd and grpc instead of Hyperbahn and thrift.

PR #712, opened 2018-07-27
idle sweeper: Don't close connections with pending calls

### LINKED ISSUE #701: Calls can still be pending for a connection that's idle
If the idle checker has kicked for any particular connection and there're still calls inflight, then something else is going wrong.

* Place some logging to identify that this has happened.
* Don't try to close the connection if there is a pending call.
* Make sure that the idle check timeout is greater than the call timeout.

prashantv:
As documented in #701, we currently attempt closing connections even
if they have pending calls. This can cause issues if the calls is stuck
as the connection enters start-close, so it'll reject new incoming
calls, but it doesn't close, so the remote side keeps sending calls over
the same connection.

Fixes #701.

### DIFF
--- a/idle_sweep_test.go
+++ b/idle_sweep_test.go
@@ -291,6 +291,7 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
 	clock := testutils.NewStubClock(time.Now())
 
 	listener := newPeerStatusListener()
+	// TODO: Log filtering doesn't require the message to be seen.
 	serverOpts := testutils.NewOpts().
 		AddLogFilter("Skip closing idle Connection as it has pending calls.", 1).
 		SetOnPeerStatusChanged(listener.onStatusChange).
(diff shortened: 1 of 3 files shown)


idle_sweep_test.go
```diff
@@ -280,3 +282,64 @@ func TestIdleSweepMisconfiguration(t *testing.T) {
...
+func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+
+	clientTicker := testutils.NewFakeTicker()
+	clock := testutils.NewStubClock(time.Now())
+
+	listener := newPeerStatusListener()
+	serverOpts := testutils.NewOpts().
+		AddLogFilter("Skip closing idle Connection as it has pending calls.", 1).
```
witriew: Was looking over this and I initially thought that it would check and verify for this log message. It'd probably be great if we could do that as well.

prashantv: Added a TODO

PR #713, opened 2018-08-09
Support go1.10, and remove go1.7

prashantv:
go1.10 now has an "internal/testlog" package that's used by production
packages, but it shouldn't count as a test-dependency, so ignore it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ test: clean setup install_test check_no_test_deps $(BIN)/thrift
 	PATH=$(BIN):$$PATH go test -run TestFramesReleased -stressTest $(TEST_ARG)
 
 check_no_test_deps:
-	! go list -json $(PROD_PKGS) | jq -r .Deps[] | grep -e test -e mock
+	! go list -json $(PROD_PKGS) | jq -r '.Deps | select ((. | length) > 0) | .[]' | grep -e test -e mock | grep -v '^internal/testlog'
 
 benchmark: clean setup $(BIN)/thrift
 	echo Running benchmarks:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,20 +6,20 @@ cache:
   - $HOME/.glide/cache
 
 go:
-- 1.7
-- 1.8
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -92,7 +92,7 @@ test: clean setup install_test check_no_test_deps $(BIN)/thrift
 	PATH=$(BIN):$$PATH go test -run TestFramesReleased -stressTest $(TEST_ARG)
 
 check_no_test_deps:
-	! go list -json $(PROD_PKGS) | jq -r .Deps[] | grep -e test -e mock
+	! go list -json $(PROD_PKGS) | jq -r '.Deps | select ((. | length) > 0) | .[]' | grep -e test -e mock | grep -v '^internal/testlog'
```
AllenLuUber: verified:
```
~/gocode/src/github.com/uber/tchannel-go(dev*) » go version                                                                                          
go version go1.10.3 darwin/amd64
~/gocode/src/github.com/uber/tchannel-go(dev) » make check_no_test_deps                                                                              
! go list -json . ./http ./hyperbahn ./json ./peers ./pprof ./raw ./relay ./stats ./thrift ./examples/bench/server ./examples/bench/client ./examples/ping ./examples/thrift ./examples/hyperbahn/echo-server | jq -r .Deps[] | grep -e test -e mock
jq: error (at <stdin>:1308): Cannot iterate over null (null)
internal/testlog
## with change
~/gocode/src/github.com/uber/tchannel-go(dev*) » make check_no_test_deps                                                                             allenlu@allenlu-C02PJ1RYG3QP
! go list -json . ./http ./hyperbahn ./json ./peers ./pprof ./raw ./relay ./stats ./thrift ./examples/bench/server ./examples/bench/client ./examples/ping ./examples/thrift ./examples/hyperbahn/echo-server | jq -r '.Deps | select ((. | length) > 0) | .[]' | grep -e test -e mock | grep -v '^internal/testlog'
```

prashantv: Thanks for checking!

PR #714, opened 2018-08-09
relay: Ensure idle sweeper also considers relay calls

### LINKED ISSUE #701: Calls can still be pending for a connection that's idle

prashantv:
The previous fix for #701 avoids the idle sweeper from closing
connections with pending calls. However, it didn't take into account
relay calls (which do not set up exchanges). Ensure we take any pending
relay calls into account as well.

Fixes #701 for pending calls on relays as well as clients/servers.

### DIFF
--- a/idle_sweep_test.go
+++ b/idle_sweep_test.go
@@ -293,13 +293,13 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
 	listener := newPeerStatusListener()
 	// TODO: Log filtering doesn't require the message to be seen.
 	serverOpts := testutils.NewOpts().
-		AddLogFilter("Skip closing idle Connection as it has pending calls.", 1).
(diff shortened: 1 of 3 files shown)


idle_sweep_test.go
```diff
@@ -326,21 +326,48 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
...
+			preCloseConns int
+			tick          func()
+		}{
+			ch:            ts.Server(),
+			preCloseConns: 2,
+			tick: func() {
+				clock.Elapse(5 * time.Minute)
+				clientTicker.Tick()
+			},
+		}
```
witriew: nit: newline

PR #715, opened 2018-08-28
Use sub-tests when running test multiple times

prashantv:
Most tests are run a couple of times, once where clients connect directly to a server, and once where clients connect to servers through a relay.

Use sub-tests so it's clearer which failed without digging through logs.

### DIFF
--- a/testutils/test_server.go
+++ b/testutils/test_server.go
@@ -95,6 +95,23 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
 	return ts
 }
 
+// runSubTest runs the specified function as a sub-test of a testing.T or
+// testing.B if the types match.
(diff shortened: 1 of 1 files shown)


testutils/test_server.go
```diff
@@ -95,6 +95,23 @@ func NewTestServer(t testing.TB, opts *ChannelOpts) *TestServer {
 	return ts
 }
 
+// runSubTest runs the specified function as a sub-test of a testing.T or
+// testing.B if the types match.
+func runSubTest(t testing.TB, name string, f func(t testing.TB)) {
```
AllenLuUber: nit: `f func(testing.TB)`

witriew: +1

PR #716, opened 2018-09-06
Sending a large error should mention that we're out of space

prashantv:
The returned error now includes the underlying error from the typed
writer, resulting in:
> failed to create outbound error frame: no more room in buffer


twilly:
LGTM

PR #718, opened 2018-10-02
Add a way to signal when a Channel has closed

witriew:
Channel has Close() and Closed() methods that tells a Channel to close
and returns whether the Channel is already closed, respectively.  This
adds a CloseChan() that returns a channel that will close once the
Channel has completed closing.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -157,6 +157,7 @@ type Channel struct {
 	relayTimerVerify    bool
 	handler             Handler
 	onPeerStatusChanged func(*Peer)
+	closed              chan struct{}
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
@@ -257,6 +258,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		relayHost:         opts.RelayHost,
 		relayMaxTimeout:   validateRelayMaxTimeout(opts.RelayMaxTimeout, logger),
 		relayTimerVerify:  opts.RelayTimerVerification,
+		closed:            make(chan struct{}),
 	}
 	ch.peers = newRootPeerList(ch, opts.OnPeerStatusChanged).newChild()
 
@@ -758,12 +760,20 @@ func (ch *Channel) connectionCloseStateChange(c *Connection) {
 
 func (ch *Channel) onClosed() {
 	removeClosedChannel(ch)
+
+	close(ch.closed)
 	ch.log.Infof("Channel closed.")
 }
 
 // Closed returns whether this channel has been closed with .Close()
 func (ch *Channel) Closed() bool {
 	return ch.State() == ChannelClosed
+}
+
+// ClosedChan returns a channel that will close when the Channel has completely
+// closed.
+func (ch *Channel) ClosedChan() <-chan struct{} {
+	return ch.closed
 }
 
 // State returns the current channel state.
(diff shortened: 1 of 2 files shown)


testutils/test_server.go
```diff
@@ -294,6 +294,9 @@ func (ts *TestServer) addChannel(createChannel func(t testing.TB, opts *ChannelO
 func (ts *TestServer) close(ch *tchannel.Channel) {
 	ch.Close()
 	ts.waitForChannelClose(ch)
```
prashantv: can we replace the `waitForChannelClose` with a select statement on the channel instead? The `select` should have a timeout as well.

we should remove all existing loops where we wait for close with the channel instead

channel.go
```diff
@@ -758,6 +760,16 @@ func (ch *Channel) connectionCloseStateChange(c *Connection) {
 
 func (ch *Channel) onClosed() {
 	removeClosedChannel(ch)
+
+	ch.mutable.Lock()
+	select {
+	case <-ch.mutable.closed:
```
prashantv: Do we need this check, I think `onClosed` is guaranteed to only happen once right?

channel.go
```diff
@@ -281,6 +282,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	}
 	ch.mutable.state = ChannelClient
 	ch.mutable.conns = make(map[uint32]*Connection)
+	ch.mutable.closed = make(chan struct{})
```
prashantv: why is `closed` mutable?  I think it's immutable (set on channel creation) and should only ever be closed once

channel.go
```diff
@@ -758,6 +760,16 @@ func (ch *Channel) connectionCloseStateChange(c *Connection) {
 
 func (ch *Channel) onClosed() {
 	removeClosedChannel(ch)
+
+	ch.mutable.Lock()
```
prashantv: see above, don't think this is mutable or needs a lock

testutils/test_server.go
```diff
@@ -293,7 +292,16 @@ func (ts *TestServer) addChannel(createChannel func(t testing.TB, opts *ChannelO
 // it waits for the channels to close.
 func (ts *TestServer) close(ch *tchannel.Channel) {
 	ch.Close()
-	ts.waitForChannelClose(ch)
+
+	select {
+	case <-time.After(500 * time.Millisecond):
```
prashantv: * use `testutils.Timeout` -- the value should be scaled up in slwoer environments.
 * The previous code did 60 iterations, each iteration taking i * 100 microseconds. i think this means we waited 100 microseconds * (1 + 2 + 3 + ... 60). That ends up being roughly 1800 * 100 microseconds or 180ms. This is a slight bump up (probably needed because we weren't using `testutils.Timeout`. I'm OK with keeping this higher, but can we test if 200ms works OK?

testutils/test_server.go
```diff
@@ -293,7 +292,16 @@ func (ts *TestServer) addChannel(createChannel func(t testing.TB, opts *ChannelO
...
+		ts.Errorf("Channel %p did not close after 500 ms, last state: %v", ch, ch.State())
```
prashantv: once we use `Timeout(500ms)`, we might want to use the same value here

testutils/test_server.go
witriew: I tried this with 250ms and it was still failing, https://travis-ci.org/uber/tchannel-go/builds/436414905.

channel.go
witriew: Makes sense.  Moved it out of `mutable`.

channel.go
witriew: I thought so too, but it looks like it's called from both `Close` and `connectionCloseStateChange`.

And the tests appear to fail if I don't guard it.

testutils/test_server.go
witriew: Errrp.. looks like I missed

```
export TEST_TIMEOUT_SCALE=20
```

Sounds like 200ms should work and then the Travis scaling should let it work.

channel.go
```diff
@@ -758,6 +761,14 @@ func (ch *Channel) connectionCloseStateChange(c *Connection) {
 
 func (ch *Channel) onClosed() {
 	removeClosedChannel(ch)
+
+	select {
+	case <-ch.closed:
+		// Do nothing if already closed.
```
prashantv: Why would it already be closed? Only if we have a bug right? We can either:
 - always close (which would panic in prod if we closed an already closed channel). since channels aren't created dynamically anyway, this seems OK
 - log an error if the channel's already closed

channel.go
```diff
@@ -266,6 +267,8 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		ch.handler = channelHandler{ch}
 	}
 
+	ch.closed = make(chan struct{})
```
prashantv: can you add this field to the channel constructor,
```
  closed: make(chan struct{}),
```

PR #719, opened 2018-10-02
relay: Export TCP connection's RemoteAddr with calls

prashantv:
This will be used internally to track connections with what they are
calling.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -184,10 +184,12 @@ type Relayer struct {
 	// It allows timer re-use, while allowing timers to be created and started separately.
 	timeouts *relayTimerPool
 
-	peers   *RootPeerList
-	conn    *Connection
-	logger  Logger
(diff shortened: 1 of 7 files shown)


relay.go
```diff
@@ -184,10 +184,12 @@ type Relayer struct {
...
 	timeouts *relayTimerPool
 
-	peers   *RootPeerList
-	conn    *Connection
-	logger  Logger
-	pending atomic.Uint32
+	peers      *RootPeerList
+	conn       *Connection
+	relayConn  *relay.Conn
+	remoteAddr string
```
jaricftw: Do you need a separate field `remoteAddr` since it's also in the `relayConn`?

prashantv: Unused field from a different variation of the change, removed. Thanks!

PR #727, opened 2019-01-29
Simplify shutdown by removing sendCh ref tracking

prashantv:
Previously, the sendCh was closed, we had to carefully track all
references to it, as any send to a closed channel would panic.

However, this was finally removed in #388:
https://github.com/uber/tchannel-go/pull/388/commits/93ef5c112c8b321367ae52d2bd79396e2e874f31

As mentioned in the change, we can simplify the shutdown path
as we no longer need to track references to sendCh. Caller can
attempt writes to sendCh, but those frames will not be sent
as they race with the close.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -24,7 +24,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"net"
 	"sync"
 	"time"
@@ -190,8 +189,6 @@ type Connection struct {
 	closeNetworkCalled atomic.Bool
 	// stoppedExchanges is atomically set when exchanges are stopped due to error.
 	stoppedExchanges atomic.Bool
-	// pendingMethods is the number of methods running that may block closing of sendCh.
-	pendingMethods atomic.Int64
 	// remotePeerAddress is used as a cache for remote peer address parsed into individual
 	// components that can be used to set peer tags on OpenTracing Span.
 	remotePeerAddress peerAddressComponents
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -740,30 +725,10 @@ func (c *Connection) updateLastActivity(frame *Frame) {
...
-}
-
-// pendingExchangeMethodDone should be deferred by a method called
-// pendingExchangeMessageAdd.
-func (c *Connection) pendingExchangeMethodDone() {
-	c.pendingMethods.Dec()
-}
-
-// closeSendCh waits till there are no other goroutines that may try to write
+// closeStopCh waits till there are no other goroutines that may try to write
```
witriew: This comment isn't entirely accurate anymore.

Can we also add that closing the `sendCh` is not needed and ref 93ef5c1 as well?

prashantv: Removed this method completely as it's pretty trivial now, and added comments to the only place this is used.

prashantv:
`stopExchanges` is weird right now, since it's a round-about way of closing the mex to avoid races with closing `sendCh` (that no longer happens).

I initially tried to clean that up here, but it has other impacts (some tests get different errors), so planning to clean that up as a separate followup.

connection.go
```diff
@@ -740,33 +725,6 @@ func (c *Connection) updateLastActivity(frame *Frame) {
...
-	// will ensure that if any other methods start that may add exchanges
-	// they will fail due to closed connection.
-	for !c.pendingMethods.CAS(0, math.MinInt32) {
-		time.Sleep(time.Millisecond)
-	}
-
-	close(c.stopCh)
-}
-
 // hasExchanges returns whether there's any pending inbound or outbound calls on this connection.
```
witriew: nit: `hasPendingCalls`

PR #728, opened 2019-02-20
Check max string length when using WriteLenXString

prashantv:
Previously, if a long string was passed in, we would overflow the length
and write the full string which resulted in a corrupted packet. Instead,
return an error if the passed in string is too long for the specified
max length (defined by size of the prefix).

Plumbing context for the error message would require significant plumbing
and may affect performance, so left it as a generic error.

### DIFF
--- a/typed/buffer.go
+++ b/typed/buffer.go
@@ -32,6 +32,10 @@ var (
 
 	// ErrBufferFull is returned when trying to write past end of buffer
 	ErrBufferFull = errors.New("no more room in buffer")
+
+	// errStringTooLong is returned when writing a string with length larger
+	// than the allows length limit.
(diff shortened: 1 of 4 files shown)


typed/buffer.go
```diff
@@ -32,6 +32,10 @@ var (
 
 	// ErrBufferFull is returned when trying to write past end of buffer
 	ErrBufferFull = errors.New("no more room in buffer")
+
+	// errStringTooLong is returned when writing a string with length larger
+	// than the allows length limit.
+	errStringTooLong = errors.New("string is too long")
```
witriew: Not exported?

witriew: Intentionally not exported?

prashantv: added note

PR #730, opened 2019-04-01
Update CHANGELOG and version in prep for 1.13.0 release

jcorbin:


CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+## [1.13.0] - 2019-04-01
+### Added
+ * Add MaxCloseTime as a graceful close timeout (#724)
+
+### Changed
+ * Eliminate []byte(string) allocation under thrift transport (#729)
+ * Use atomic.Bool from go.uber.org/atomic v1.3 (#726)
```
prashantv: I think we can remove this -- we tend to just have user-facing changes

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+## [1.13.0] - 2019-04-01
+### Added
+ * Add MaxCloseTime as a graceful close timeout (#724)
```
prashantv: can we update to something like:
 * Add `MaxCloseTime` which sets a timeout for graceful connection close.

 - prefer backticks for go type/field names
 - more user-friendly explanation

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
...
+
+### Changed
+ * Eliminate []byte(string) allocation under thrift transport (#729)
```
prashantv: More user-friendly: `Optimize Thrift string field serialization by eliminating `[]byte(string)` allocation.

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
...
+ * Use atomic.Bool from go.uber.org/atomic v1.3 (#726)
+
+### Fixed
+ * Check max string length when using WriteLenXString (#728)
```
prashantv: More user-friendly: Return an error if transport header keys/values exceed the maximum allowed string length.

(that's the only case affected by this change iirc)

PR #731, opened 2019-05-01
relay: Fix stuck calls when client TCP buffer is full

### LINKED ISSUE #700: relay: Calls can get "stuck" if clients TCP buffer is blocked
We saw a case where a connection close got stuck (similar to #687) in a specific edge case:

 * The client makes a call, but by the time the response is received, they are not processing response frames, and cause the TCP writes to block. This causes the `sendCh` for the connection to fill up, and eventaully causes calls to [be dropped](https://github.com/uber/tchannel-go/blob/v1.11.0/relay.go#L271)
 * On the last frame of the call response, we attempt to stop the timeout timer and succeed (e.g., the call has not yet timed out)
 * Due to the full buffer, we attempt to call `failRelayItem`

Since this is the last frame of the response, we stop the timeout, and in doing so, essentially take ownership of cleaning up the relay item. We call `failRelayItem` to clean up the item, but `failRelayItem` only cleans up the relay if it's able to stop the timeout. Since we've already stopped the timeout, we end up leaving an orphan call, causing connection closes to get stuck.

We need to always clear out the relay item once we receive the last call of a frame.

prashantv:
See #700 for the cause of the bug. The summary:
 * When a frame finishes the relay item, we explicitly stop the timeout
 * If the frame fails to send, failRelayItem is called
 * failRelayItem returns early (without cleaning up) if it can't stop
 the timeout

The double-stop of the timeout caused the call to not be cleaned up. The
fix is simple: avoid stopping the timeout initially. The code already
handles the timeout firing right as we do the check, so there's no
advantage in checking. It's stopped a few lines later (after we write
the frame to the sendCh).

Added a test to reproduce client buffer filling up, which fails in a few
ways without the fix:
 * Not all calls are marked as "Ended"
 * Leftover pending calls
 * Channels fail to close.

Fixes #700.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -21,6 +21,7 @@
 package tchannel_test
 
 import (
+	"bytes"
 	"errors"
 	"io"
 	"runtime"
@@ -60,6 +61,23 @@ func withRelayedEcho(t testing.TB, f func(relay, server, client *Channel, ts *te
 		client.Peers().Add(ts.HostPort())
 		f(ts.Relay(), ts.Server(), client, ts)
 	})
+}
+
+func getRelayPending(state *RuntimeState) int {
+	var totalPending int
+
+	aggregateConns := func(conns []ConnectionRuntimeState) {
+		for _, c := range conns {
+			totalPending += c.Relayer.Count
+		}
+	}
+
+	for _, peer := range state.RootPeers {
+		aggregateConns(peer.InboundConnections)
+		aggregateConns(peer.OutboundConnections)
+	}
+
+	return totalPending
 }
 
 func TestRelay(t *testing.T) {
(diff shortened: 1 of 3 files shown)


relay_test.go
```diff
@@ -775,6 +793,80 @@ func TestRelayStalledConnection(t *testing.T) {
...
+// See https://github.com/uber/tchannel-go/issues/700 for more info.
+func TestRelayStalledClientConnection(t *testing.T) {
+	// This needs to be large enough to fill up the client TCP buffer.
+	const _calls = 100
+
+	opts := testutils.NewOpts().
+		// Expect errors from dropped frames.
+		AddLogFilter("Dropping call due to slow connection to destination.", _calls).
+		AddLogFilter("Couldn't send outbound frame.", _calls).
+		// DisableLogVerification(). // we expect warnings due to removed relay item.
```
witriew: Intentionally commented out?

relay_test.go
```diff
@@ -775,6 +793,80 @@ func TestRelayStalledConnection(t *testing.T) {
...
+			// Wait for server to receive the call
+			<-gotCall
+
+			calls = append(calls, call)
+		}
+
+		// Wait for all calls to end on the relay, and ensure we got failures from the slow client.
+		stats := ts.RelayHost().Stats()
+		stats.WaitForEnd()
+		assert.Contains(t, stats.Map(), "testService-client->s1::echo.failed-relay-dest-conn-slow", "Expect at least 1 failed call due to slow client")
```
witriew: We're also expecting that none of the calls succeeded, right? Also that the remaining calls should've timed-out?

prashantv: from a relay's perspective, we wrote the response frame, so they "succeeded". we expect them to succeed, but the client won't actually see them due to the block.

unfortunately the write syscall succeeding doesn't actually mean the other side will ever see that data.

PR #741, opened 2019-05-16
Expose `CallOptions` caller name for transparent proxying

### LINKED ISSUE #725: Expose callerName on the CallOptions so it is easier to implement proxies
Currently, if we implement a proxy in tchannel-go and want to retain the original caller information, we need to do `(*InboundCall).CallOptions()` and set that on the outbound. Else we would see proxy as the caller. 
The [suggested workaround](https://github.com/uber/tchannel-go/blob/dev/calloptions.go#L61) does not work when the code does not have a reference to the *InboundCall. Can we expose this `callerName` so we can directly set it?

AlexAPB:
This makes the existing caller name field on `CallOptions` public. This
enables frameworks like `yarpc-go` to support transparent proxying, 
when the inbound caller is not known until run time.

Fixes #725 and T2469647.

### DIFF
--- a/calloptions.go
+++ b/calloptions.go
@@ -56,9 +56,10 @@ type CallOptions struct {
 	// to an instance of the intended service.
 	RoutingDelegate string
 
-	// callerName can only be used when forwarding a request. It can only be set internally,
-	// e.g. by calling (*InboundCall).CallOptions() when forwarding a request
-	callerName string
+	// CallerName is the name of the service issuing the call. It can also be used
+	// to support transparent proxying when the inbound caller name varies across
+	// calls.
+	CallerName string
 }
 
 var defaultCallOptions = &CallOptions{}
@@ -82,8 +83,8 @@ func (c *CallOptions) overrideHeaders(headers transportHeaders) {
 	if c.RoutingDelegate != "" {
 		headers[RoutingDelegate] = c.RoutingDelegate
 	}
-	if c.callerName != "" {
-		headers[CallerName] = c.callerName
+	if c.CallerName != "" {
+		headers[CallerName] = c.CallerName
 	}
 }
 
--- a/calloptions.go
+++ b/calloptions.go
@@ -56,9 +56,9 @@ type CallOptions struct {
 	// to an instance of the intended service.
 	RoutingDelegate string
 
-	// CallerName is the name of the service issuing the call. It can also be used
-	// to support transparent proxying when the inbound caller name varies across
-	// calls.
+	// CallerName defaults to the channel's service name for an outbound call.
+	// Optionally override this field to support transparent proxying when inbound
+	// caller names vary across calls.
 	CallerName string
 }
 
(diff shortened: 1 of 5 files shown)


inbound_test.go
```diff
@@ -143,6 +143,41 @@ func TestInboundConnection_CallOptions(t *testing.T) {
 	})
 }
 
+func TestCallOptionsPropogated(t *testing.T) {
```
AlexAPB: Let me know if this isn't the right spot for this integration test, wasn't sure of a better spot to put this.

prashantv: lgtm

calloptions.go
```diff
@@ -56,9 +56,10 @@ type CallOptions struct {
 	// to an instance of the intended service.
 	RoutingDelegate string
 
-	// callerName can only be used when forwarding a request. It can only be set internally,
-	// e.g. by calling (*InboundCall).CallOptions() when forwarding a request
-	callerName string
+	// CallerName is the name of the service issuing the call. It can also be used
```
prashantv: Should we mention that this for outbound calls, it's automatically set to the channel's service name, and should only be specified when that needs to be overridden (e.g., transparent proxying bit you have).

inbound_test.go
```diff
@@ -143,6 +143,41 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+
+		ctx, cancel := NewContext(testutils.Timeout(time.Second))
+		defer cancel()
+
+		call, err := ts.Server().BeginCall(ctx, ts.HostPort(), ts.ServiceName(), handler, &giveCallOpts)
+		require.NoError(t, err, "could not call test server")
+
+		_, _, _, err = raw.WriteArgs(call, nil, nil)
+		require.NoError(t, err, "could not write args")
+
```
prashantv: we should probably read the response just to make sure the call is "completed" successfully. otherwise the test cleanup logic will probably wait for a timeout and will run slower

inbound_test.go
```diff
@@ -143,6 +143,41 @@ func TestInboundConnection_CallOptions(t *testing.T) {
...
+		assert.Equal(t, giveCallOpts, *gotCallOptions)
```
prashantv: prefer to do `assert.Equal(t, &giveCallOpts, gotCallOptions)` so that even if `gotCallOptions` is `nil`, it won't panic

we should remove the `require.NotNil` in the handler since `require` is intended for the main test goroutine

if it's `nil`, this `assert` can catch it

prashantv: super nit: opts or options as suffix for both the names for consistency

inbound_test.go
AlexAPB: `raw.WriteArgs` already reads both arg2 and arg3 is there something else I need to read?

https://github.com/uber/tchannel-go/blob/b1db3fe904fde57d507ecb4c4caf068dbce1a0f1/raw/call.go#L59

prashantv: Oh cool, I did not name this function well, thanks :)

AlexAPB:
Restarted builds, saw same flaky test as described in #689

PR #742, opened 2019-05-20
Update CHANGELOG and version in prep for 1.14.0 release

AlexAPB:


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+## [1.14.0] - 2019-05-20
+### Added
+ * Expose `CallOptions` caller name for transparent proxying (#741)
+
+### Fixed
```
prashantv: can remove these, the relay path isn't part of the "public" API (muttley consumes `dev` directly) so no need to add these

AlexAPB: Sounds good, didn't realize this.

PR #744, opened 2019-07-03
thrift: deal with TStruct serialization failures

### LINKED ISSUE #743: Handle errors when writing the thrift struct value returned by a TChanServer.Handle.
`TStruct.Write` may return an error yet [resp.Write](https://github.com/uber/tchannel-go/blob/dev/thrift/server.go#L206)'s error is not checked. This may result in an empty or broken final argument for a call response.

twilly:
A Thrift server handler can return success with a TStruct that may
return an error when writing a serialized form. This error is ignored,
which cause a invalid third argument to be written (or not written at
all).

We can better handle these kinds of errors earlier by buffering a
serialized TStruct and sending along a system error if the
serialization fails.

Fixes #743

### DIFF
--- a/thrift/server.go
+++ b/thrift/server.go
@@ -21,7 +21,6 @@
 package thrift
 
 import (
-	"bytes"
 	"log"
 	"strings"
 	"sync"
@@ -186,17 +185,6 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
 		return err
 	}
 
-	// Buffer the response write so we can send a system error to the client.
-	// It's too late to send a system error by the time we write Arg3.
-	var respStruct bytes.Buffer // XXX should we make a buffer pool?
-	wp = getProtocolWriter(&respStruct)
-	if err := resp.Write(wp.protocol); err != nil {
-		call.Response().SendSystemError(err)
-		thriftProtocolPool.Put(wp)
-		return err
-	}
-	thriftProtocolPool.Put(wp)
-
 	if !success {
 		call.Response().SetApplicationError()
 	}
@@ -213,10 +201,13 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
 		return err
 	}
 
-	// Write buffered response.
 	writer, err = call.Response().Arg3Writer()
-	if _, err := writer.Write(respStruct.Bytes()); err != nil {
-		writer.Close() // XXX wrap this error?
+
+	wp = getProtocolWriter(writer)
+	defer thriftProtocolPool.Put(wp)
+
+	if err := resp.Write(wp.protocol); err != nil {
+		call.Response().SendSystemError(err)
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


thrift/server.go
```diff
@@ -185,6 +186,17 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
 		return err
 	}
 
+	// Buffer the response write so we can send a system error to the client.
+	// It's too late to send a system error by the time we write Arg3.
+	var respStruct bytes.Buffer // XXX should we make a buffer pool?
```
kriskowal: That’s a decision best made with data, but I expect pooling to yield a 20% improvement in latency.

thrift/server.go
```diff
@@ -185,6 +186,17 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
...
+	wp = getProtocolWriter(&respStruct)
+	if err := resp.Write(wp.protocol); err != nil {
+		call.Response().SendSystemError(err)
```
abhinav: This can return an err at which point it should be a tableflip

thrift/server.go
```diff
@@ -185,6 +186,17 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
 		return err
 	}
 
+	// Buffer the response write so we can send a system error to the client.
+	// It's too late to send a system error by the time we write Arg3.
```
prashantv: Are we sure about this? I'm quite sure Muttley can send an error halfway through a response, and the clients see the error.

twilly:
> writing directly to arg3, but then sending an error halfway through?

I'm still looking into this. I did a quick pass and wasn't sure we can terminate a half-written arg3 and then start another trailer arg.

> writing to arg3, but if it fails, logging an error, and blackholing that response (no more bytes, but the response is never completed. client will see a timeout)

I'm trying to avoid this case. Timeouts don't tell the client what happened and the server knows what happened.

thrift/server.go
twilly: Same goes with writing out the buffer, eg the socket died. Plenty of tableflip failure modes that will at least get a server side error.

thrift/server.go
```diff
@@ -202,12 +202,16 @@ func (s *Server) handle(origCtx context.Context, handler handler, method string,
...
 
 	writer, err = call.Response().Arg3Writer()
+
 	wp = getProtocolWriter(writer)
-	resp.Write(wp.protocol)
-	thriftProtocolPool.Put(wp)
-	err = writer.Close()
+	defer thriftProtocolPool.Put(wp)
+
+	if err := resp.Write(wp.protocol); err != nil {
```
prashantv: (unrelated to this diff but just thought of this) @twilly just as a safety check, let's verify that the passed in protocol here is a `TBinaryProtocol` so that users don't end up calling `Write` with a custom protocol and expect it to work?

abhinav: Agreed. Created an internal ticket to track this.

PR #746, opened 2019-07-27
Add method to RelayFrame to reveal Arg2 offset

devastating:
Add a method "Arg2EndOffset" to RelayFrame interface so that we could
know how much space Arg2 is used. This is the first step to know if it's
possible to add more methods in Relay code to allow Relayer to inject
more headers or read the headers to decide which peer to forward to.

NOTE: this is a breaking change to RelayFrame interface.

Related to #745. Will use this change to know Arg2 usage.

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -88,6 +88,8 @@ type lazyCallReq struct {
 	*Frame
 
 	caller, method, delegate, key []byte
+
+	arg2EndOffset int
 }
 
 // TODO: Consider pooling lazyCallReq and using pointers to the struct.
@@ -132,6 +134,10 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 	arg1Len := int(binary.BigEndian.Uint16(f.Payload[cur : cur+2]))
 	cur += 2
 	cr.method = f.Payload[cur : cur+arg1Len]
+
+	// arg2~2
+	cur += arg1Len
+	cr.arg2EndOffset = cur + 2 /*arg2 len*/ + int(binary.BigEndian.Uint16(f.Payload[cur:cur+2]))
 	return cr
 }
 
@@ -181,6 +187,12 @@ func (f lazyCallReq) Span() Span {
 // HasMoreFragments returns whether the callReq has more fragments.
 func (f lazyCallReq) HasMoreFragments() bool {
 	return f.Payload[_flagsIndex]&hasMoreFragmentsFlag != 0
+}
+
+// Arg2EndOffset returns the offset from start of frame to the end of Arg2
+// in bytes.
+func (f lazyCallReq) Arg2EndOffset() int {
+	return f.arg2EndOffset
 }
 
 // finishesCall checks whether this frame is the last one we should expect for
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -89,7 +89,8 @@ type lazyCallReq struct {
 
 	caller, method, delegate, key []byte
 
-	arg2EndOffset int
+	arg2StartOffset, arg2EndOffset int
+	isArg2Fragmented               bool
 }
 
 // TODO: Consider pooling lazyCallReq and using pointers to the struct.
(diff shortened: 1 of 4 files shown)


relay_messages.go
```diff
@@ -132,6 +134,10 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 	arg1Len := int(binary.BigEndian.Uint16(f.Payload[cur : cur+2]))
 	cur += 2
 	cr.method = f.Payload[cur : cur+arg1Len]
+
+	// arg2~2
+	cur += arg1Len
+	cr.arg2EndOffset = cur + 2 /*arg2 len*/ + int(binary.BigEndian.Uint16(f.Payload[cur:cur+2]))
```
prashantv: if the frame has a continue bit, then arg2 offset is past this. should we somehow flag this to the caller as well?

devastating: Good point - added isFragmented in the Arg2EndOffset method to indicate if Arg2 is fragmented.

devastating:
> Add test cases for:
> 
>     * 0 length arg2
> 
>     * positive length arg2 (with and without overflow to another frame)
> 
>     * max length arg2 (with and without overflow to another frame)
> 
> 
> We might also want to add arg2 starting offset so we can use [end-start] as an estimate of the header size for metrics (related to T2938739)

Not quite sure I understand the difference between positive length and max length, but I've added tests for 0 length arg2 and arg2 which spans across frames.

relay_messages.go
```diff
@@ -132,6 +135,13 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 	arg1Len := int(binary.BigEndian.Uint16(f.Payload[cur : cur+2]))
 	cur += 2
 	cr.method = f.Payload[cur : cur+arg1Len]
+
+	// arg2~2
+	cur += arg1Len
+	cr.arg2StartOffset = cur + 2
+	cr.arg2EndOffset = cur + 2 /*arg2 len*/ + int(binary.BigEndian.Uint16(f.Payload[cur:cur+2]))
```
prashantv: instead of repeating `cur + 2` to calculate the end offset, do `arg2EndOffset = cr.arg2StartOffset + [...]`

relay_messages.go
```diff
@@ -132,6 +135,13 @@ func newLazyCallReq(f *Frame) lazyCallReq {
...
+	// arg2 is fragmented if we don't see arg3 in this frame.
+	cr.isArg2Fragmented = int(cr.Header.FrameSize()) <= (FrameHeaderSize+cr.arg2EndOffset) && f.Payload[_flagsIndex]&hasMoreFragmentsFlag != 0
```
prashantv: instead of using frame size here, if you use `PayloadSize()`, then you won't need to add `FrameHeaderSize` and it'll be a litlte simpler.

similarly, don't access the flags directly, use the method `HasMoreFragments()`

```go
cr.isArg2Fragmented = int(cr.Header.PayloadSize()) <= cr.arg2EndOffset && cr.HasMoreFragments()
```

can we add a test for case where a call doesn't set the has more fragments flag, and doesn't have arg3

testutils/call.go
```diff
@@ -122,3 +122,15 @@ func (f FakeCallFrame) RoutingKey() []byte {
 func (f FakeCallFrame) RoutingDelegate() []byte {
 	return []byte(f.RoutingDelegateF)
 }
+
+// Arg2StartOffset returns the offset from start of frame to
+// the beginning of Arg2.
+func (f FakeCallFrame) Arg2StartOffset() int {
+	return 0
```
prashantv: we should probably add fields here to control these to match the others

relay_messages.go
devastating: Good point - I think I made the mistake that the offset is from start of the payload to Arg2, not from the start of frame (which includes header).

relay_messages.go
```diff
@@ -183,6 +193,18 @@ func (f lazyCallReq) HasMoreFragments() bool {
 	return f.Payload[_flagsIndex]&hasMoreFragmentsFlag != 0
 }
 
+// Arg2EndOffset returns the offset from start of payload to the end of Arg2
+// in bytes, and whether there are more data from other frames.
```
prashantv: "more data from other frames" --> let's make this more clear. we only return true here if there's more frames and we haven't yet started arg3, right?

prashantv:
As discussed offline, let's follow this up with integration tests that use the `Arg2Writer` to write different values (huge values, forcefully flush, etc) to ensure the returned values match what we expected.

Separately, can you file an issue for ensuring we don't ever "overflow" reads? If we send a corrupt packet, it may cause out-of-bounds access.

devastating:
> As discussed offline, let's follow this up with integration tests that use the `Arg2Writer` to write different values (huge values, forcefully flush, etc) to ensure the returned values match what we expected.
> 
> Separately, can you file an issue for ensuring we don't ever "overflow" reads? If we send a corrupt packet, it may cause out-of-bounds access.

Created #747

PR #748, opened 2019-07-31
Add integration test for Arg2StartOffset/Arg2EndOFfset methods

devastating:
Add integration test to ensure the arg2 start/end offsets are properly
set. Several test cases are included:

1. Everything in one frame
2. Arg2 is flushed before reaching frame size limit
3. Arg2 has not data
4. Arg2 is larger than frame size limit and thus fragmented

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -1061,7 +1061,7 @@ func TestRelayArg2OffsetIntegration(t *testing.T) {
 		{
 			msg:           "XL arg2 which is fragmented",
 			arg2Data:      string(make([]byte, MaxFrameSize+100)),
-			wantEndOffset: 65519,
+			wantEndOffset: wantArg2Start + payloadLeft,
 			wantHasMore:   true,
 		},
 		{
--- a/relay_test.go
+++ b/relay_test.go
@@ -1022,12 +1022,11 @@ func TestRelayArg2OffsetIntegration(t *testing.T) {
 	exLargeArg2 := make([]byte, MaxFrameSize+100)
 
 	tests := []struct {
-		msg                string
-		arg2Data           []byte
-		arg2Flush          bool
-		arg3Data           []byte
-		wantArg2Fragmented bool
-		wantHasMore        bool
+		msg         string
+		arg2Data    []byte
+		arg2Flush   bool
+		arg3Data    []byte
+		wantHasMore bool
 	}{
 		{
 			msg:      "all within a frame",
@@ -1046,10 +1045,9 @@ func TestRelayArg2OffsetIntegration(t *testing.T) {
 			arg3Data: arg3Data,
 		},
 		{
-			msg:                "huge arg2",
-			arg2Data:           exLargeArg2,
-			wantHasMore:        true,
-			wantArg2Fragmented: true,
+			msg:         "huge arg2",
+			arg2Data:    exLargeArg2,
+			wantHasMore: true,
 		},
 	}
 
@@ -1067,7 +1065,9 @@ func TestRelayArg2OffsetIntegration(t *testing.T) {
 			f := <-inspector.received
 			start := f.Arg2StartOffset()
 			end, hasMore := f.Arg2EndOffset()
-			if tt.wantArg2Fragmented {
+			if len(tt.arg2Data) > MaxFrameSize {
+				// Arg2 is larger than max allowed frame size and
+				// should cause fragmentation.
 				assert.True(t, end > start)
 				assert.Equal(t, MaxFrameSize-FrameHeaderSize, end)
 			} else {
--- a/relay_test.go
+++ b/relay_test.go
@@ -1074,7 +1074,7 @@ func TestRelayArg2OffsetIntegration(t *testing.T) {
 				wantHasMore:       true,
 			},
 			{
-				msg:           "no arg2",
+				msg:           "no arg2 but flushed",
 				wantEndOffset: wantArg2Start,
 				wantHasMore:   false,
 			},
@@ -1138,14 +1138,14 @@ func TestRelayArg2OffsetIntegration(t *testing.T) {
 				f := <-inspector.received
 				start := f.Arg2StartOffset()
 				end, hasMore := f.Arg2EndOffset()
-				assert.Equal(t, wantArg2Start, start)
-				assert.Equal(t, tt.wantEndOffset, end)
-				assert.Equal(t, tt.wantHasMore, hasMore)
+				assert.Equal(t, wantArg2Start, start, "arg2 start offset does not match expectation")
+				assert.Equal(t, tt.wantEndOffset, end, "arg2 end offset does not match expectation")
+				assert.Equal(t, tt.wantHasMore, hasMore, "arg2 hasMore bit does not match expectation")
 
 				gotArg2, gotArg3, err := raw.ReadArgsV2(call.Response())
 				assert.NoError(t, err)
-				assert.Equal(t, tt.arg2Data+tt.arg2PostFlushData, string(gotArg2))
-				assert.Equal(t, arg3DataToWrite, string(gotArg3))
+				assert.Equal(t, tt.arg2Data+tt.arg2PostFlushData, string(gotArg2), "arg2 in response does not meet expectation")
+				assert.Equal(t, arg3DataToWrite, string(gotArg3), "arg3 in response does not meet expectation")
 			})
 		}
 	})
--- a/relay_test.go
+++ b/relay_test.go
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
 		wg.Wait()
 	})
 }
+
+func TestRelayArg2OffsetIntegration(t *testing.T) {
+	ctx, cancel := NewContext(testutils.Timeout(time.Second))
+	defer cancel()
+
+	inspector := newFailedRelayHost()
+	opts := testutils.NewOpts().
+		SetRelayOnly().
+		SetRelayHost(inspector)
+
+	ch := testutils.NewServer(t, opts)
+	defer ch.Close()
+
+	client := testutils.NewClient(t, nil /*opts*/)
+	defer client.Close()
+
+	arg2Data := []byte("moe")
+	arg3Data := []byte("auto-tconfig")
+	exLargeArg2 := make([]byte, MaxFrameSize+100)
+
+	tests := []struct {
+		msg                string
+		arg2Data           []byte
+		arg2Flush          bool
+		arg3Data           []byte
+		wantArg2Fragmented bool
+		wantHasMore        bool
+	}{
+		{
+			msg:      "all within a frame",
+			arg2Data: arg2Data,
+			arg3Data: arg3Data,
+		},
+		{
+			msg:         "arg2 flushed",
+			arg2Data:    arg2Data,
+			arg2Flush:   true,
+			arg3Data:    arg3Data,
+			wantHasMore: true,
+		},
+		{
+			msg:      "no arg2",
+			arg3Data: arg3Data,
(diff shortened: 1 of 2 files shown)


relay_test.go
```diff
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+	arg2Data := []byte("moe")
+	arg3Data := []byte("auto-tconfig")
+	exLargeArg2 := make([]byte, MaxFrameSize+100)
+
+	tests := []struct {
+		msg         string
+		arg2Data    []byte
+		arg2Flush   bool
+		arg3Data    []byte
+		wantHasMore bool
```
prashantv: can we also validate the start/end offsets, since those should be constant

relay_test.go
```diff
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+		SetRelayOnly().
+		SetRelayHost(inspector)
+
+	ch := testutils.NewServer(t, opts)
+	defer ch.Close()
+
+	client := testutils.NewClient(t, nil /*opts*/)
+	defer client.Close()
+
+	arg2Data := []byte("moe")
```
prashantv: optional: can use `string` everywhere here and just convert to `[]byte` before writing

relay_test.go
```diff
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+	}{
+		{
+			msg:      "all within a frame",
+			arg2Data: arg2Data,
+			arg3Data: arg3Data,
+		},
+		{
+			msg:         "arg2 flushed",
+			arg2Data:    arg2Data,
+			arg2Flush:   true,
```
prashantv: can we optionally support a post-flush write to arg2, just in case (shouldn't matter but still)

relay_test.go
```diff
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+			f := <-inspector.received
+			start := f.Arg2StartOffset()
+			end, hasMore := f.Arg2EndOffset()
+			if len(tt.arg2Data) > MaxFrameSize {
+				// Arg2 is larger than max allowed frame size and
+				// should cause fragmentation.
+				assert.True(t, end > start)
+				assert.Equal(t, MaxFrameSize-FrameHeaderSize, end)
+			} else {
+				assert.Equal(t, len(tt.arg2Data), end-start)
```
prashantv: should we compare the frame's payload directly to arg2?

relay_test.go
```diff
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+func newFailedRelayHost() *failedRelayHost {
+	return &failedRelayHost{
+		received: make(chan relay.CallFrame, 1),
+	}
+}
+
+func (r *failedRelayHost) SetChannel(*Channel) {}
+
+func (r *failedRelayHost) Start(f relay.CallFrame, conn *relay.Conn) (RelayCall, error) {
+	r.received <- f
```
prashantv: I think the API doesn't allow you to hold a reference to anything from the `relay.CallFrame` after the method returns, since the frame could be freed and reused

can you copy the relevant fields (maybe into a fake call frame) and pass that through?

relay_test.go
```diff
@@ -1001,3 +1001,98 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+		},
+		{
+			msg:      "no arg2",
+			arg3Data: arg3Data,
+		},
+		{
+			msg:         "huge arg2",
+			arg2Data:    exLargeArg2,
+			wantHasMore: true,
+		},
```
prashantv: can we add more edge conditions here:
 - arg2 ends around the frame border (e.g., 3 bytes left is enough for small arg3, 2 bytes left is enough for arg3 length but nothing else, 1 byte left isn't enough for anything for arg3, etc)

relay_test.go
devastating: I was thinking to do it, but hesitated due to the hard coded number. Will add the values.

relay_test.go
devastating: Don't think we have access to the Payload.

relay_test.go
devastating: Good point. Added a method in testutils to do the copy.

relay_test.go
```diff
@@ -1001,3 +1001,132 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+			msg:           "large arg2 with 3 bytes left for arg3",
+			arg2Data:      string(make([]byte, payloadLeft-3)),
+			wantEndOffset: wantArg2Start + payloadLeft - 3,
+			wantHasMore:   false,
+		},
+		{
+			msg:           "large arg2, 2 bytes left",
+			arg2Data:      string(make([]byte, payloadLeft-2)),
+			wantEndOffset: wantArg2Start + payloadLeft - 2,
+			wantHasMore:   true, // arg3 is fragmented
```
prashantv: when there's 2 bytes left in the payload, do we not send arg3 length of 0? i think regardless of the length of arg3, we'd always set the "hasMore" flag but only have arg3 on the next frame

devastating: Good point - I think there is a 2 byte mistake, as I forgot to account for the 2 bytes Arg2 length field; writing payloadLeft-2 for arg2 actually means we will write

2 (arg2Length) + (payloadLeft - 2) = payloadLeft

i.e. we used up all payload space for arg2. Hence the frame is fragmented.

devastating: Actually, we don't use the 2 bytes in current implementation. Validated it by using tcpdump and wireshark.

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				assert.NoError(t, NewArgReader(call.Response().Arg3Reader()).Read(&resArg))
+				assert.Equal(t, arg3DataToWrite, string(resArg))
+			})
+		}
+	})
+}
+
+// failedRelayHost is a RelayHost which always fails RelayCall,
+// but provides way to introspect the received CallFrame.
+type failedRelayHost struct {
```
prashantv: rename type + constructor, update comment since we no longer fail. it's really just a decorator that stores the received frame now

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+}
+
+func newFailedRelayHost(r RelayHost) *failedRelayHost {
+	return &failedRelayHost{
+		RelayHost: r,
+		received:  make(chan relay.CallFrame, 1),
+	}
+}
+
+//func (r *failedRelayHost) SetChannel(*Channel) {}
```
prashantv: can delete

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				msg:           "large arg2, 1 bytes left",
+				arg2Data:      string(make([]byte, payloadLeft-1)),
+				wantEndOffset: wantArg2Start + payloadLeft - 1,
+				wantHasMore:   true, // no arg3
+			},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.msg, func(t *testing.T) {
+
```
prashantv: nit: delete blank

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				call, err := client.BeginCall(ctx, ts.HostPort(), ts.ServiceName(), testMethod, nil)
+				require.NoError(t, err, "BeginCall failed")
+				writer, err := call.Arg2Writer()
+				arg2WriteHelper := NewArgWriter(writer, err)
+				require.NoError(t, arg2WriteHelper.Write([]byte(tt.arg2Data)), "arg3 write failed")
```
prashantv: don't use the arg writer here, since it also calls `Close`. we should use the returned `Arg2Writer` directly.

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				if tt.arg2Flush {
+					writer.Flush()
+					// tries to write after flush
+					if tt.arg2PostFlushData != "" {
+						arg2WriteHelper.Write([]byte(tt.arg2PostFlushData))
+						writer.Flush()
```
prashantv: is this flush necessary?

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				call, err := client.BeginCall(ctx, ts.HostPort(), ts.ServiceName(), testMethod, nil)
+				require.NoError(t, err, "BeginCall failed")
+				writer, err := call.Arg2Writer()
+				arg2WriteHelper := NewArgWriter(writer, err)
+				require.NoError(t, arg2WriteHelper.Write([]byte(tt.arg2Data)), "arg3 write failed")
+				if tt.arg2Flush {
+					writer.Flush()
+					// tries to write after flush
+					if tt.arg2PostFlushData != "" {
+						arg2WriteHelper.Write([]byte(tt.arg2PostFlushData))
```
prashantv: ensure no errors on write

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+
+				f := <-inspector.received
+				start := f.Arg2StartOffset()
+				end, hasMore := f.Arg2EndOffset()
+				assert.Equal(t, wantArg2Start, start)
+				assert.Equal(t, tt.wantEndOffset, end)
+				assert.Equal(t, tt.wantHasMore, hasMore)
+
+				var resArg []byte
+				assert.NoError(t, NewArgReader(call.Response().Arg2Reader()).Read(&resArg))
```
prashantv: optionally: can use `raw.ReadArgsV2` to read all of these in one call.

relay_test.go
```diff
@@ -1014,3 +1014,161 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				assert.Equal(t, tt.arg2Data, string(resArg))
```
prashantv: shouldn't this be `arg2Data + arg2PostFlushData`?

relay_test.go
```diff
@@ -1014,3 +1014,159 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+			{
+				msg:               "arg2 flushed called then write again",
+				arg2Data:          arg2Data,
+				arg2Flush:         true,
+				arg2PostFlushData: "more data",
+				wantEndOffset:     wantArg2Start + len(arg2Data),
+				wantHasMore:       true,
+			},
+			{
+				msg:           "no arg2",
```
prashantv: let's add `no arg2 but flushed`

relay_test.go
```diff
@@ -1014,3 +1014,159 @@ func TestRelayRaceCompletionAndTimeout(t *testing.T) {
...
+				arg3DataToWrite := arg3Data
+				if tt.noArg3 {
+					arg3DataToWrite = ""
+				}
+				require.NoError(t, NewArgWriter(call.Arg3Writer()).Write([]byte(arg3DataToWrite)), "arg3 write failed")
+
+				f := <-inspector.received
+				start := f.Arg2StartOffset()
+				end, hasMore := f.Arg2EndOffset()
+				assert.Equal(t, wantArg2Start, start)
```
prashantv: here and above, add context to the assert so if it fails, we have useful messages

PR #749, opened 2019-08-01
Close hyperbahn client when benchmark.internalServer is closed

cinchurge:
Currently in package benchmark, when `internalServer.Advertise()` is called, a `hyperbahn.Client` is created but never closed even when the `internalServer` is closed. this can result in a goroutine leak since `hyperbahn.Client.advertiseLoop()` can never return.

This change fixes two things:
1. it keeps track of the allocated `*hyperbahn.Client` in `internalServer`  and  closes the client when `internalServer` is closed.
2. the default function for `hyperbahn.Client.sleep` is changed to a `select` which will not block when the client has been closed.

### DIFF
--- a/benchmark/internal_server.go
+++ b/benchmark/internal_server.go
@@ -37,6 +37,7 @@ import (
 // internalServer represents a benchmark server.
 type internalServer struct {
 	ch          *tchannel.Channel
+	hc          *hyperbahn.Client
 	opts        *options
 	rawCalls    atomic.Int64
 	thriftCalls atomic.Int64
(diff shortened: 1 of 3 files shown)


benchmark/internal_server.go
```diff
@@ -84,16 +85,22 @@ func (s *internalServer) HostPort() string {
...
 	if err != nil {
 		panic("failed to setup Hyperbahn client: " + err.Error())
 	}
-	return hc.Advertise()
+	return s.hc.Advertise()
 }
 
 func (s *internalServer) Close() {
-	s.ch.Close()
+	if s.ch != nil {
```
prashantv: this is never `nil` right?

cinchurge: yeah.. was trying to be defensive, but agreed that it's not necessary

cinchurge: > Maybe we should add a simple test that uses this and the `WithTestServer` and verifies that there's no goroutine leaks after Close?

yea, but it seems like goleak isn't used in this code base. perhaps a follow up to add goleak would be useful.

prashantv: If we use `WithTestServer` and add a test in the tchannel package, it does goroutine leak checks already.

PR #751, opened 2019-08-08
Add Tchannel-Thrift Arg2 Key/Value pair iterator

devastating:
The goal of the iterator is to allow RelayHost to have a way to inspect
the Arg2 header and make the relay decision based on some application
header.

To optimize for performance, we try to avoid allocation as possible.

TODO: we should add more checks to avoid accessing out of bound.

### DIFF
--- a/thrift/arg2/kv_iterator.go
+++ b/thrift/arg2/kv_iterator.go
@@ -4,7 +4,11 @@
 // backwards-compatibility guarantee.
 package arg2
 
-import "encoding/binary"
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+)
 
 // KeyValIterator is a iterator for reading tchannel-thrift Arg2 Scheme,
 // which has key/value pairs (k~2 v~2).
@@ -18,16 +22,16 @@ type KeyValIterator struct {
 }
 
 // NewKeyValIterator inits a KeyValIterator with the buffer pointing at
-// start of Arg2.
+// start of Arg2. Return io.EOF if no iterator is available.
 // NOTE: tchannel-thrift Arg Scheme starts with number of key/value pair.
-func NewKeyValIterator(arg2Payload []byte) (KeyValIterator, bool) {
+func NewKeyValIterator(arg2Payload []byte) (KeyValIterator, error) {
 	if len(arg2Payload) < 2 {
-		return KeyValIterator{}, false
+		return KeyValIterator{}, io.EOF
 	}
 
 	nh := int(binary.BigEndian.Uint16(arg2Payload[0:2]))
 	if nh <= 0 {
-		return KeyValIterator{}, false
+		return KeyValIterator{}, io.EOF
 	}
 
 	return KeyValIterator{
@@ -36,7 +40,7 @@ func NewKeyValIterator(arg2Payload []byte) (KeyValIterator, bool) {
 	}.next(2 /*nh*/)
 }
 
-// Key Returns the key
+// Key Returns the key.
 func (i KeyValIterator) Key() []byte {
 	return i.arg2Payload[i.keyOffset : i.valueOffset-2 /*2B length*/]
 }
@@ -46,10 +50,11 @@ func (i KeyValIterator) Value() []byte {
 	return i.arg2Payload[i.valueOffset : i.valueOffset+i.valueLen]
 }
 
-// Next returns next iterator. Return nil if no more key/value pair available.
-func (i KeyValIterator) Next() (KeyValIterator, bool) {
+// Next returns next iterator. Return io.EOF if no more key/value pair is
+// available.
+func (i KeyValIterator) Next() (KeyValIterator, error) {
 	if i.leftPairCount <= 0 {
-		return KeyValIterator{}, false
+		return KeyValIterator{}, io.EOF
 	}
 
 	return i.next(i.valueOffset + i.valueLen)
@@ -57,10 +62,10 @@ func (i KeyValIterator) Next() (KeyValIterator, bool) {
 
 // cur is the offset from the start of Arg2 payload to next key/value pair
 // we want to iterate to.
-func (i KeyValIterator) next(cur int) (KeyValIterator, bool) {
+func (i KeyValIterator) next(cur int) (KeyValIterator, error) {
 	arg2Len := len(i.arg2Payload)
 	if cur+2 > arg2Len {
-		return KeyValIterator{}, false
+		return KeyValIterator{}, fmt.Errorf("invalid key offset %v (arg2 len %v)", cur, arg2Len)
 	}
 	keyLen := int(binary.BigEndian.Uint16(i.arg2Payload[cur : cur+2]))
 	cur += 2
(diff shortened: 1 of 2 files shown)


thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,87 @@
+package arg2
+
+import "encoding/binary"
+
+// KeyValIterator is a iterator for reading tchannel-thrift Arg2 Scheme,
+// which has key/value pairs (k~2 v~2).
+// NOTE: to be optimized for performance, we try to limit the allocation
+// done in the process of iteration.
+type KeyValIterator struct {
+	arg2Len               int
```
prashantv: do we need this and `arg2Payload`? i think we should use `len(arg2Payload)`

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,87 @@
...
+	leftPairCount         int
+	keyOffset, keyLen     int
+	valueOffset, valueLen int
```
prashantv: do we need offset + length for both of these? i'd assume `keyOffset` and `valueOffset` are enough to figure out `keyLen`?

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,87 @@
...
+	if i.leftPairCount <= 0 {
+		return KeyValIterator{}, false
+	}
+
+	return i.next(i.valueOffset + i.valueLen)
+}
+
+// cur is the offset from the start of Arg2 payload to next key/value pair
+// we want to iterate to.
+func (i KeyValIterator) next(cur int) (KeyValIterator, bool) {
```
prashantv: hmm this is OK, but we're reimplementing what `typed.ReadBuffer` does, can we reuse some of that logic?

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,87 @@
+package arg2
```
prashantv: should we add package docs that this is not a supported stable API, but is internal for relays?

thrift/arg2/kv_iterator.go
devastating: Good point.

thrift/arg2/kv_iterator.go
devastating: That's a good optimization.

thrift/arg2/kv_iterator.go
devastating: The main reason I did not use it is typed.ReadBuffer is using allocation now.

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,88 @@
...
+	arg2Payload           []byte
+	leftPairCount         int
+	keyOffset             int
+	valueOffset, valueLen int
+}
+
+// InitKeyValIterator inits a KeyValIterator with the buffer pointing at
+// start of Arg2.
+// NOTE: tchannel-thrift Arg Scheme starts with number of key/value pair.
+func InitKeyValIterator(arg2Payload []byte) (KeyValIterator, bool) {
```
prashantv: why `Init` instead of `NewKeyValueIterator`?

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,88 @@
...
+	if len(arg2Payload) < 2 {
+		return KeyValIterator{}, false
+	}
+
+	nh := int(binary.BigEndian.Uint16(arg2Payload[0:2]))
+	if nh <= 0 {
```
prashantv: is this branch worth having? it seems like the resulting slice without this will be basically the same

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,88 @@
...
+	if i.leftPairCount <= 0 {
+		return KeyValIterator{}, false
+	}
+
+	return i.next(i.valueOffset + i.valueLen)
+}
+
+// cur is the offset from the start of Arg2 payload to next key/value pair
+// we want to iterate to.
+func (i KeyValIterator) next(cur int) (KeyValIterator, bool) {
```
prashantv: should we return an error in the "unexpected" cases? it seems like truncated key/value should be an error that we expose.

thrift/arg2/kv_iterator.go
prashantv: Ok, since the `relay_messages.go` doesn't use it either, let's leave that for later.

thrift/arg2/kv_iterator.go
devastating: Initially it's operated on a struct, but since we are changing it to return a struct, I will update the name to NewKeyValIterator.

thrift/arg2/kv_iterator.go
devastating: I think it's worthy having this branch as it won't cause much for the if check on an cached value, and it provides the opportunity to exit quit.

Also based on existing code, if the arg2 length is set to be long enough, but nh == 0 (invalid case), ignoring the branch could cause the iterator iterating on an invalid kv pair.

thrift/arg2/kv_iterator.go
devastating: I did think about this - do you think it makes sense to change the interface to

```
// error is set if encountering any unexpected error during iteration. KeyValIterator
// will be nil if no more KV to be iterated.
(i KeyValIterator) Next() (KeyValIterator, error)
```

devastating: Or we could have

```
var ErrNoMoreArg2KV = errors.New("end of thrift Arg2")

// return ErrNoMoreArg2KV if no more iterator available.
(i KeyValIterator) Next() (KeyValIterator, error)
```

thrift/arg2/kv_iterator.go
prashantv: Since we're using a `Uint16`, the value can never be negative -- it can only be 0 or positive right? The 0 value is positive that we handle already.

If we replace this branch, and do `.Next()`, then we're essentially doing the same check just in `Next` instead of here (we're creating a `KeyValueIterator`, but not sure that the optimization for that is worth it).

If we do that, we probably don't even have a need for `next(..)` since the logic can be inlined into `Next(.)`

thrift/arg2/kv_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+	wb := typed.NewWriteBuffer(buf)
+	wb.WriteUint16(nh)
+	for i := 0; i < nh; i++ {
+		wb.WriteLen16String(fmt.Sprintf("key%v", i))
+		wb.WriteLen16String(fmt.Sprintf("value%v", i))
+	}
+
+	iter, err := NewKeyValIterator(buf[:wb.BytesWritten()])
+	for i := 0; i < nh; i++ {
+		assert.NoError(t, err)
```
prashantv: nit: add context messages to assert

thrift/arg2/kv_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+		assert.Equal(t, fmt.Sprintf("key%v", i), string(iter.Key()))
+		assert.Equal(t, fmt.Sprintf("value%v", i), string(iter.Value()))
+		iter, err = iter.Next()
+	}
+	assert.Equal(t, io.EOF, err)
+
+	t.Run("init iterator w/o Arg2", func(t *testing.T) {
+		var buf []byte
+		_, err := NewKeyValIterator(buf)
```
prashantv: can just pass in `nil` here, no need to create `var buf []byte`

thrift/arg2/kv_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+		}{
+			{
+				msg:     "ok",
+				arg2Len: wb.BytesWritten(),
+			},
+			{
+				msg:     "not enough to read key len",
+				arg2Len: 3, // nh (2) + 1
+				wantErr: "invalid key offset 2 (arg2 len 3)",
+			},
```
prashantv: can we have `not enough to read key` where we pass in the length fully, but only part of the key

thrift/arg2/kv_iterator.go
```diff
@@ -0,0 +1,84 @@
...
+// NOTE: tchannel-thrift Arg Scheme starts with number of key/value pair.
+func NewKeyValIterator(arg2Payload []byte) (KeyValIterator, error) {
+	if len(arg2Payload) < 2 {
+		return KeyValIterator{}, io.EOF
+	}
+
+	return KeyValIterator{
+		valueOffset:   2, // nh has 2B offset
+		leftPairCount: int(binary.BigEndian.Uint16(arg2Payload[0:2])),
+		arg2Payload:   arg2Payload,
```
prashantv: optionally: instead of setting `valueOffset` to `2`, you can leave it at default 0, and pass in `arg2Payload[2:]`

PR #752, opened 2019-08-14
Add Arg2 KV helper

devastating:
This change adds a new test helper to build Arg2 buffer based on given key/value pair. Will be used in next change to hook up Arg2Iterator with RelayFrame.


thrift/arg2/testutils/arg2_kv_builder.go
```diff
@@ -0,0 +1,23 @@
+package testutils
```
prashantv: let's move this to `testutils/thriftarg2test` or something, `testutils` is too generic of a package and clashes with the other `testutils` we already have

PR #753, opened 2019-08-14
Relay msg kv iterator

devastating:
Add Arg2Iterator method into RelayFrame interface to start exposing Arg2 to RelayHost.

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -23,6 +23,7 @@ package tchannel
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"time"
@@ -36,6 +37,8 @@ var (
 	_routingKeyKeyBytes      = []byte(RoutingKey)
 	_argSchemeKeyBytes       = []byte(ArgScheme)
 	_tchanThriftValueBytes   = []byte(Thrift)
+
+	errBadArg2Len = errors.New("bad Arg2 length")
 )
 
 const (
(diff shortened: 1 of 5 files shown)


relay_messages.go
```diff
@@ -206,6 +217,20 @@ func (f lazyCallReq) Arg2StartOffset() int {
 	return f.arg2StartOffset
 }
 
+// Arg2Iterator returns the iterator for reading Arg2 key value pair
+// of TChannel-Thrift Arg Scheme.
+func (f lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
+	if !f.hasTChanThrift {
+		return arg2.KeyValIterator{}, io.EOF
```
prashantv: if we have calls that are not tchannel, can we add more context on what we got? maybe a different error, and should we try to capture the actual arg scheme we saw?

devastating: Updated - was thinking to save some info, but I guess bool won't be too far from []byte anyway.

relay_messages_test.go
```diff
@@ -43,8 +46,10 @@ const (
 )
 
 type testCallReqParams struct {
-	flags   byte
-	arg2Buf []byte
+	flags           byte
+	hasTChanThrift  bool
```
prashantv: can we make this `argScheme Format` and just pass in what we want (whether it's `Thrift` or `Raw`) and if nothing is set, then just not set the header. that way we can test for both cases (not set, and not "thrift") in the test cases

relay_messages_test.go
```diff
@@ -352,6 +364,113 @@ func TestLazyCallArg2Offset(t *testing.T) {
...
+					arg2Buf:        arg2Buf,
+				})
+				gotIter := make(map[string]string)
+				iter, err := cr.Arg2Iterator()
+				for err == nil {
+					gotIter[string(iter.Key())] = string(iter.Value())
+					iter, err = iter.Next()
+				}
+				if tt.wantBadErr != "" {
+					require.NotEqual(t, io.EOF, err)
```
prashantv: here and below, add context messages to assert

relay_messages_test.go
```diff
@@ -352,6 +364,113 @@ func TestLazyCallArg2Offset(t *testing.T) {
...
+					assert.Contains(t, err.Error(), tt.wantBadErr)
+				} else {
+					assert.Equal(t, io.EOF, err)
+					wantKV := tt.wantKV
+					if wantKV == nil {
+						wantKV = tt.bufKV
+					}
+					assert.Equal(t, wantKV, gotIter, "%v", crt)
```
prashantv: do you want to do `%+v` to make the frame more readable, and maybe add context. E.g., "unexpected arg2 key values for call request: %+v"

that said, the crt isn't really easy to display (since it'll be lots of bytes), so might not be worth having it at all

relay_messages_test.go
devastating: done

PR #755, opened 2019-08-22
Ensure Introspection endpoints are always available

prashantv:
Currently, introspection endpoints only work if the service doesn't use
SetHandler to completely take over the procedure->handler mapping, and
if the caller uses either the service name of the channel or "tchannel"
to hit the introspection endpoints.

Instead, let's always handle introspection first regardless of the
service name.

The test added panics without this change,
```
=== RUN   TestIntrospectionNotBlocked
=== RUN   TestIntrospectionNotBlocked/no_relay
panic: should not be called

goroutine 20 [running]:
github.com/uber/tchannel-go_test.TestIntrospectionNotBlocked.func1.1(0xa498c0,
0xc000152070, 0xc0001da000)
        /home/prashant/go/src/github.com/uber/tchannel-go/introspection_test.go:125
```


subchannel_test.go
```diff
@@ -195,8 +195,7 @@ func TestGetHandlers(t *testing.T) {
 	}{
 		{
 			serviceName: ch.ServiceName(),
-			// Default service name comes with extra introspection methods.
-			wantMethods: []string{"_gometa_introspect", "_gometa_runtime", "method1", "method2"},
+			wantMethods: []string{"method1", "method2"},
```
AllenLuUber: To confirm this removal: before we registered both introspection and actual procedure together. Now we don't and those introspection are not visible anymore during actual handling.

prashantv: Yep exactly. They're no longer just "user-level" methods, but internal methods that are always registered at a higher layer, so they don't show up here anymore.

PR #757, opened 2019-08-23
Export FakeFrame.Arg2KVIterator to faciliate tests

cinchurge:
With arg2KVIterator unexported, it is extremely difficult for consumers to test with call frames when arg2 is involved. This change simply exports that attr so that is is possible to supply a custom iterator, such as via `arg2.NewKeyValIterator`.

### DIFF
--- a/testutils/call.go
+++ b/testutils/call.go
@@ -100,7 +100,7 @@ type FakeCallFrame struct {
 	Arg2StartOffsetVal, Arg2EndOffsetVal int
 	IsArg2Fragmented                     bool
 
-	arg2KVIterator    arg2.KeyValIterator
+	Arg2KVIterator    arg2.KeyValIterator
 	hasArg2KVIterator error
 }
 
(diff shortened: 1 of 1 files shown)


testutils/call.go
```diff
@@ -100,7 +100,7 @@ type FakeCallFrame struct {
 	Arg2StartOffsetVal, Arg2EndOffsetVal int
 	IsArg2Fragmented                     bool
 
-	arg2KVIterator    arg2.KeyValIterator
+	Arg2KVIterator    arg2.KeyValIterator
 	hasArg2KVIterator error
```
devastating: I think you need to export hasArg2KVIterator as well.

prashantv:
For someone creating a fake call frame, does it make more sense to provide a setter that takes a `map[string]string` that it uses to create the iterator?

Similarly, should we merge `Arg2StartOffset` and `Arg2EndOffset` into a single `Arg2Size`? I think that's the only use-case we have for the values right?

PR #759, opened 2019-09-17
Support for factory method for outbound connections

samarabbas:
New optional factory method on ChannelOptions to allow passing in custom dialer
for outbound connections.  This could be used for things like TLS handshake
considering tchannel already has support for custom TLS listener.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -114,8 +114,7 @@ type ChannelOptions struct {
 	// default handler that delegates to a subchannel.
 	Handler Handler
 
-
-	Dialer *net.Dialer
+	Dialer func(ctx context.Context, hostPort string) (net.Conn, error)
 }
 
 // ChannelState is the state of a channel.
@@ -161,7 +160,7 @@ type Channel struct {
 	internalHandlers    *handlerMap
 	handler             Handler
 	onPeerStatusChanged func(*Peer)
-	dialer              *net.Dialer
+	dialer              func(ctx context.Context, hostPort string) (net.Conn, error)
 	closed              chan struct{}
 
 	// mutable contains all the members of Channel which are mutable.
@@ -571,7 +570,7 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string) (*Connection, e
 	var tcpConn net.Conn
 	var err error
 	if ch.dialer != nil {
-		tcpConn, err = ch.dialer.DialContext(ctx, "tcp", hostPort)
+		tcpConn, err = ch.dialer(ctx, hostPort)
 	} else {
 		tcpConn, err = dialContext(ctx, hostPort)
 	}
--- a/channel.go
+++ b/channel.go
@@ -116,7 +116,7 @@ type ChannelOptions struct {
 
 	// Dialer is optional factory method which can be used for overriding
 	// outbound connections for things like TLS handshake
-	Dialer func(ctx context.Context, hostPort string) (net.Conn, error)
+	Dialer func(ctx context.Context, network, hostPort string) (net.Conn, error)
 }
 
 // ChannelState is the state of a channel.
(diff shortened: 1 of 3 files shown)


channel.go
```diff
@@ -113,6 +113,10 @@ type ChannelOptions struct {
 	// Handler is an alternate handler for all inbound requests, overriding the
 	// default handler that delegates to a subchannel.
 	Handler Handler
+
+	// Dialer is optional factory method which can be used for overriding
+	// outbound connections for things like TLS handshake
+	Dialer func(ctx context.Context, hostPort string) (net.Conn, error)
```
prashantv: Should this function match the signature of `net.DialContext`?

channel.go
```diff
@@ -563,7 +569,13 @@ func (ch *Channel) Connect(ctx context.Context, hostPort string) (*Connection, e
 	}
 
 	timeout := getTimeout(ctx)
-	tcpConn, err := dialContext(ctx, hostPort)
+	var tcpConn net.Conn
+	var err error
+	if ch.dialer != nil {
```
prashantv: instead of an `if` here, can we set the dialer function to `dialContext` in the constructor instead. That way we avoid some of the noise here and simplify this logic.

connection_test.go
```diff
@@ -1322,3 +1322,28 @@ func TestInvalidTransportHeaders(t *testing.T) {
 		})
 	}
 }
+
+func TestCustomDialer(t *testing.T) {
+	sopts := testutils.NewOpts().NoRelay()
```
prashantv: add a comment for why are doing `NoRelay`

it would be great if we didn't have that option and made the test work for relays as well

connection_test.go
```diff
@@ -1322,3 +1322,28 @@ func TestInvalidTransportHeaders(t *testing.T) {
...
+
+		copts := testutils.NewOpts().SetDialer(func(ctx context.Context, hostPort string) (net.Conn, error) {
+			customDialerCalled = true
+			d := net.Dialer{}
+			return d.DialContext(ctx, "tcp", hostPort)
+		})
+
+		// Induce the creation of a connection from client to server.
+		client := ts.NewClient(copts)
+		require.NoError(t, testutils.CallEcho(client, ts.HostPort(), ts.ServiceName(), nil))
```
prashantv: use `testutils.AssertEcho`

connection_test.go
```diff
@@ -1322,3 +1322,28 @@ func TestInvalidTransportHeaders(t *testing.T) {
...
+		assert.True(t, customDialerCalled, "custom dialer used for establishing connection")
+
+		// Re-use
```
prashantv: what are we trying to validate here? That the connection isn't created again? If so, we should make `customDialerCalled` a counter and verify it's still `1`.

connection_test.go
```diff
@@ -1322,3 +1322,28 @@ func TestInvalidTransportHeaders(t *testing.T) {
...
+		testutils.AssertEcho(t, client, ts.HostPort(), ts.ServiceName())
+
+		client.Close()
```
prashantv: nit: no need, the `WithTestServer` will automatically close clients

connection_test.go
samarabbas: I copied most of the test from "TestPeerStatusChangeClient".  So inherited setting NoRelay from there.  Able to get the test to pass with NoRelay to removing it.

connection_test.go
samarabbas: Done.

connection_test.go
samarabbas: Yes.  Added another assert on `customDialerCalledCount` as suggested.

connection_test.go
samarabbas: removed.

channel.go
samarabbas: Didn't see much value in it originally as network is hardcoded to "tcp" all over the place.  Here is how it looks like if I make signature look the same.  Let me know what you think.

connection_test.go
```diff
@@ -1322,3 +1322,27 @@ func TestInvalidTransportHeaders(t *testing.T) {
 		})
 	}
 }
+
+func TestCustomDialer(t *testing.T) {
+	sopts := testutils.NewOpts()
```
prashantv: you can just pass `nil` to `WithTestServer` and it'll use "default" opts which is what you're passing right now

PR #760, opened 2019-09-18
Add --thriftNoRecurse flag to control passing -r to thrift

robbertvanginkel:
To allow generating thrift without generating for all dependencies. Flag
is a negative to preserve current behaviour.


thrift/thrift-gen/generate.go
```diff
@@ -78,7 +79,13 @@ func runThrift(inFile string, outDir string) error {
 
 	// Generate the Apache Thrift generated code.
 	goArgs := fmt.Sprintf("go:thrift_import=%s,package_prefix=%s", *apacheThriftImport, *packagePrefix)
-	if err := execThrift("-r", "--gen", goArgs, "-out", outDir, inFile); err != nil {
+	args := []string{
+		"--gen", goArgs, "-out", outDir, inFile,
+	}
+	if *thriftNoRecurse == false {
```
kriskowal: Style for bool comparisons is generally to obviate the constants like `!*thriftNoRecurse`.

PR #763, opened 2020-02-19
Update CHANGELOG and version in prep for 1.17.0 release

witriew:


witriew:
> We haven't been documenting relay-related changes in the CHANGELOG, since it doesn't affect end-users. Should we just mention that some internal API changes for relaying?

Sure thing, done.

PR #766, opened 2020-03-16
Add option to random load balance to >1 server nodes

oarap:
Add option to random load balance to >1 server nodes

Adding possibility to set `peerConnectionCount` so that a single client
node can do a random load balancing among `peerConnectionCount` top
nodes using the score calculator and peer heap.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -21,7 +21,6 @@
 package tchannel
 
 import (
-	"container/heap"
 	"errors"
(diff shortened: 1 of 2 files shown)


prashantv:
This looks very similar to #631 -- can we consolidate into a single PR? Ideally into the already existing PR, as it's got some history + comments.

PR #767, opened 2020-03-17
Remove Go versions 1.11, 1.12 and add versions 1.13, 1.14

witriew:


witriew:
I was wondering whether 1.12 should be skipped entirely and move straight to 1.13 and 1.14...

abhinav:
> I was wondering whether 1.12 should be skipped entirely and move straight to 1.13 and 1.14...

Yeah, that sounds reasonable. 1.13 and 1.14 is the way to go now.

PR #768, opened 2020-03-17
relay: Log send buffer usage and size when dropping calls

prashantv:
Ref T5413559

We drop calls when the sendCh is full, which is typically because
writeFrames is not reading from sendCh quickly enough. This could
be because of unfortunate scheduling, or because the send buffer is
full and write syscalls were blocked. Logging the send buffer usage
will help determine which is the cause of dropped calls.

We can get the socket send buffer usage with ioctl, and buffer size
using getsockopt.

ioctl with SIOCOUTQ returns the number of bytes in the send buffer:
See https://linux.die.net/man/7/tcp for details
> Returns the amount of unsent data in the socket send queue.

Extend the introspection output to contain the same buffer sizes as
well as the sendCh length.

Example log message fields added to the log:
```
   "destConnSendBufferCurrent":3994463,
   "destConnSendBufferLimit":4194304
```

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -908,11 +908,7 @@ func (c *Connection) sendBufSize() (sendBufUsage int, sendBufSize int, _ error) 
 }
 
 func getSysConn(conn net.Conn, log Logger) syscall.RawConn {
-	type syscallConner interface {
-		SyscallConn() (syscall.RawConn, error)
-	}
-
-	connSyscall, ok := conn.(syscallConner)
+	connSyscall, ok := conn.(syscall.Conn)
 	if !ok {
 		log.WithFields(LogField{"connectionType", fmt.Sprintf("%T", conn)}).
 			Error("Connection does not implement SyscallConn.")
--- a/connection.go
+++ b/connection.go
@@ -26,14 +26,17 @@ import (
 	"io"
 	"net"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/uber/tchannel-go/tos"
 
 	"go.uber.org/atomic"
+	"go.uber.org/multierr"
 	"golang.org/x/net/context"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -165,6 +168,7 @@ type Connection struct {
 	connDirection    connectionDirection
 	opts             ConnectionOptions
 	conn             net.Conn
+	sysConn          syscall.RawConn // may be nil if conn cannot be converted
 	localPeerInfo    LocalPeerInfo
 	remotePeerInfo   PeerInfo
 	sendCh           chan *Frame
@@ -301,6 +305,7 @@ func (ch *Channel) newConnection(conn net.Conn, initialID uint32, outboundHP str
 
 		connID:             connID,
 		conn:               conn,
+		sysConn:            getSysConn(conn, log),
 		connDirection:      connDirection,
 		opts:               opts,
 		state:              connectionActive,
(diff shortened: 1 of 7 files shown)


connection.go
```diff
@@ -873,3 +878,49 @@ func (c *Connection) closeNetwork() {
 func (c *Connection) getLastActivityTime() time.Time {
 	return time.Unix(0, c.lastActivity.Load())
 }
+
+func (c *Connection) sendBufSize() (int, int, error) {
```
alxn: we need named returns or a comment for `int, int`

connection.go
```diff
@@ -873,3 +878,49 @@ func (c *Connection) closeNetwork() {
...
+	if c.sysConn == nil {
+		return 0, 0, fmt.Errorf("no sys call conn")
```
jaricftw: nit: `errors.New` when there is no arg?

connection_internal_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+type errSyscallConn struct {
+	net.Conn
+}
+
+func (c errSyscallConn) SyscallConn() (syscall.RawConn, error) {
+	return nil, assert.AnError
+}
+
+func TestGetSysConn(t *testing.T) {
```
jaricftw: no tests for success case of `getSysConn`?

connection.go
```diff
@@ -873,3 +878,49 @@ func (c *Connection) closeNetwork() {
...
+		sendBufLimit, sockoptErr = unix.GetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_SNDBUF)
+	})
+
+	errs = multierr.Append(errs, sockoptErr)
+	errs = multierr.Append(errs, ioctlErr)
+	return sendBuf, sendBufLimit, errs
+}
+
+func getSysConn(conn net.Conn, log Logger) syscall.RawConn {
+	type syscallConner interface {
```
jaricftw: why define an internal interface instead of type asserting to `tcpConn`?

connection.go
prashantv: Named the returns, :+1:

connection.go
prashantv: Yep, leftover from my PoC. Created a top-level error using `errors.New` and returning that instead.

connection.go
prashantv: Good question -- I wanted to keep the internal logic in TChannel agnostic to TCP, in case we want to test using unix domain sockets etc in future. Technically, it's possible to pass in a custom `net.Listener` and dialer such that we deal with a `net.Conn` backed by some other transport. Since this logic is no more complex than casting to a `net.TCPConn` while being more generic, thought it was better to keep it generic

connection_internal_test.go
prashantv: The success case is tested via the end-to-end test `TestRelayStalledConnection`, but it doesn't hurt to have a positive test here too, so added.

connection.go
witriew: https://golang.org/pkg/syscall/#Conn seems to implement the same interface?  Should we use this instead?

relay.go
```diff
@@ -281,9 +281,20 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
...
+		logFields := []LogField{
+			{"id", id},
+			{"destConnSendBufferCurrent", sendBuf},
+			{"destConnSendBufferLimit", sendBufLimit},
+		}
+		if sendBufErr != nil {
+			logFields = append(logFields, LogField{"destConnSendBufferError", sendBufErr.Error()})
+		}
+
+		r.logger.WithFields(logFields...).Warn("Dropping call due to slow connection.")
```
fiibbb: I think we should we also log `len(sendCh)`, since that's gonna be configurable and may differ depending on the connection.

connection.go
prashantv: Missed that, updated to use `syscall.Conn`, thanks!

relay.go
prashantv: Good idea, I've added both `len(sendCh)` and `cap(sendCh)` to both the log output as well as the introspection output.

connection.go
```diff
@@ -873,3 +880,46 @@ func (c *Connection) closeNetwork() {
...
+
+func getSysConn(conn net.Conn, log Logger) syscall.RawConn {
+	connSyscall, ok := conn.(syscall.Conn)
+	if !ok {
+		log.WithFields(LogField{"connectionType", fmt.Sprintf("%T", conn)}).
+			Error("Connection does not implement SyscallConn.")
+		return nil
+	}
+
+	sysConn, err := connSyscall.SyscallConn()
```
masiulaniec: `*net.TCPConn` and others already have the SyscallConn method so the indirection via syscall.Conn could be avoided.

relay.go
```diff
@@ -281,9 +281,22 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
...
-		).Warn("Dropping call due to slow connection.")
+
+		// Since this is typically due to the send buffer being full, get send buffer
+		// usage + limit and add that to the log.
+		sendBuf, sendBufLimit, sendBufErr := r.conn.sendBufSize()
+		logFields := []LogField{
+			{"id", id},
+			{"destConnSendBufferCurrent", sendBuf},
+			{"destConnSendBufferLimit", sendBufLimit},
+			{"sendChQueued", len(r.conn.sendCh)},
```
masiulaniec: Perhaps also `sendChQueuedBytes`? That could hint if it's many frames in a short time window ([microburst](https://en.wikipedia.org/wiki/Micro-bursting_(networking))) vs. fragmentation of a large response.

I realize there's coordination overhead associated with this kind of accounting... Might be not worth it.

connection.go
prashantv: We're not dealing with a `*net.TCPConn`, but a `net.Conn`, so we need to cast. TChannel allows both customization of the dialer which returns any `net.Conn`, so using `syscall.Conn` is more general than casting to a `*net.TCPConn`.

relay.go
prashantv: Yeah this would require a larger accounting change, given that we get the numbre of bytes in the TCP send buffer, I think we should start with that.

If that's not enough, we can then add tracking for `sendChQueuedBytes` as a separate change, but I'd prefer to avoid that complexity for now.

connection.go
masiulaniec: I was actually thinking more along the lines of testing abstractly for presence of the `SyscallConn` method rather than concretely for `*net.TCPConn`. Having actually looked at the definition of `syscall.Conn` I realized the code as written already does that effectively (I expected the `syscall.Conn` interface to be larger).

PR #769, opened 2020-03-17
WIP: Add timeout option for writing frames to sendCh

prashantv:
Ref T2984301

Instead of dropping calls immediately if the sendCh is full, add
a timeout option which will wait for a specified duration of time
for the sendCh to have space.

This helps in the case that the client is actively reading from the
socket, or in the case that the writeFrames goroutine is actively
writing to the socket, and applies some backpressure to the backend
generating large responses.

This does add latency to process all responses from that backend, so
if there's other clients who have responses in the backend socket's
buffer, they will wait for the timeout as well.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -146,6 +146,11 @@ type ConnectionOptions struct {
 	// MaxCloseTime controls how long we allow a connection to complete pending
 	// calls before shutting down. Only used if it is non-zero.
 	MaxCloseTime time.Duration
+
+	// RelaySendTimeout is only for relays, and is the timeout to wait if the
+	// send buffer is full.
+	// This is an unstable API - breaking changes are likely.
+	RelaySendTimeout time.Duration
 }
 
 // connectionEvents are the events that can be triggered by a connection.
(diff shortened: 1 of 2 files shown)


connection.go
```diff
@@ -146,6 +146,11 @@ type ConnectionOptions struct {
 	// MaxCloseTime controls how long we allow a connection to complete pending
 	// calls before shutting down. Only used if it is non-zero.
 	MaxCloseTime time.Duration
+
+	// RelaySendTimeout is only for relays, and is the timeout to wait if the
+	// send buffer is full.
+	// This is an unstable API - breaking changes are likely.
+	RelaySendTimeout time.Duration
```
fiibbb: Nit: Add a comment on suggesting this timeout to be very small? As this can potentially put back pressure all the way to the thread reading the frames.

prashantv:
Internal testing shows that this does reduce the number of dropped packets, but it impacts other caller connections waiting on responses from the same backend.

Since we get similar results from increasing the send channel size, it seems like a better mitigation than this approach. Closing.

PR #770, opened 2020-03-19
connection: Separate out lastActivity into recv and send activity

witriew:
`lastActivity` recorded when the last receive or send activity occurred over
its connection.  Separate this out into `lastActivityRecv` and
`lastActivitySend` to observe the last receives or sends allowing monitoring of
the elapsed time between reading from the recv buffer and writing to the send
buffer.   This will provide information on how stalled either buffer is.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -218,9 +218,10 @@ type Connection struct {
 	healthCheckDone    chan struct{}
 	healthCheckHistory *healthHistory
 
-	// lastActivity is used to track how long the connection has been idle.
-	// (unix time, nano)
-	lastActivity atomic.Int64
+	// lastActivity{Read,Write} is used to track how long the connection has been
+	// idle for the recieve and send connections respectively. (unix time, nano)
+	lastActivityRead  atomic.Int64
+	lastActivityWrite atomic.Int64
 }
 
 type peerAddressComponents struct {
@@ -322,6 +323,7 @@ func (ch *Channel) newConnection(conn net.Conn, initialID uint32, outboundHP str
 
 	log = log.WithFields(LogField{"connectionDirection", connDirection})
 	peerInfo := ch.PeerInfo()
+	timeNow := ch.timeNow().UnixNano()
 
 	c := &Connection{
 		channelConnectionCommon: ch.channelConnectionCommon,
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -324,7 +325,8 @@ func (ch *Channel) newConnection(conn net.Conn, initialID uint32, outboundHP str
 		events:             events,
 		commonStatsTags:    ch.commonStatsTags,
 		healthCheckHistory: newHealthHistory(),
-		lastActivity:       *atomic.NewInt64(ch.timeNow().UnixNano()),
+		lastActivityRecv:   *atomic.NewInt64(ch.timeNow().UnixNano()),
```
prashantv: nit: can we call `ch.timeNow().UnixNano()` once and use the same value in both places -- that way they'll be the same value when a connection starts

connection.go
```diff
@@ -206,9 +206,10 @@ type Connection struct {
 	healthCheckDone    chan struct{}
 	healthCheckHistory *healthHistory
 
-	// lastActivity is used to track how long the connection has been idle.
-	// (unix time, nano)
-	lastActivity atomic.Int64
+	// lastActivity{Recv,Send} is used to track how long the connection has been
+	// idle for the recieve and send connections respectively. (unix time, nano)
+	lastActivityRecv atomic.Int64
```
prashantv: nit: since Go uses `Read` and `Write`, and our functions mirror that name (`readFrames` and `writeFrames`) it may be better to use the same names here, `lastActivityRead` and `lastActivityWrite`

connection.go
```diff
@@ -734,13 +736,23 @@ func (c *Connection) writeFrames(_ uint32) {
...
+		c.lastActivityRecv.Store(c.timeNow().UnixNano())
+	}
+}
+
+// updateLastActivitySend marks when the last message was sent on the channel.
+// This is used for monitoring idle connections and timing them out.
+func (c *Connection) updateLastActivitySend(frame *Frame) {
+	// Pings are ignored for last activity.
+	switch frame.Header.messageType {
+	case messageTypeCallReq, messageTypeCallReqContinue, messageTypeCallRes, messageTypeCallResContinue, messageTypeError:
```
prashantv: since we consider the same frames in both cases, can we move that into a separate private function

connection.go
witriew: Done.

connection.go
witriew: Sure thing, done.

alxn:
```
        18:04:32.676015 [D] reader failed: tchannel error ErrCodeTimeout: timeout existing err: tchannel error ErrCodeTimeout: timeout [{serviceName testService-client} {process testService-client-1347} {chID 1832} {connID 1028} {localAddr 127.0.0.1:59382} {remoteAddr 127.0.0.1:40123} {remoteHostPort 127.0.0.1:40123} {remoteIsEphemeral false} {remoteProcess relay-40123} {outboundHP 127.0.0.1:40123} {connectionDirection outbound} {Out-Response 39}]
        18:04:32.676165 [D] Creatin
The job exceeded the maximum log length, and has been terminated.
```

err restarted

PR #771, opened 2020-03-19
Add separate Linux and Darwin build files for socket information

witriew:
#768 introduced fetching the amount of unsent data in the send queue for Linux.
Introduce fetching the same information, but for Darwin as well.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -889,21 +889,14 @@ func (c *Connection) sendBufSize() (sendBufUsage int, sendBufSize int, _ error) 
 		return sendBufUsage, sendBufSize, errNoSyscallConn
 	}
 
-	var (
-		// current send buffer usage
-		ioctlErr error
-
-		// current send buffer limit
-		sockoptErr error
-	)
-
+	var sendBufLenErr, sendBufLimErr error
 	errs := c.sysConn.Control(func(fd uintptr) {
-		sendBufUsage, ioctlErr = unix.IoctlGetInt(int(fd), unix.SIOCOUTQ)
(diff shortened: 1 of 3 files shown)


alxn:
weird:
```
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,5 +1,5 @@
            	            	 (*relay.Conn)({
            	            	- RemoteAddr: (string) (len=15) "127.0.0.1:40885",
            	            	+ RemoteAddr: (string) (len=15) "127.0.0.1:45380",
            	            	  RemoteProcessName: (string) (len=17) "testService-40885",
            	            	- IsOutbound: (bool) true
            	            	+ IsOutbound: (bool) false
            	            	 })
            	Test:       	TestRelayConnection/with_relay
            	Messages:   	Unexpected remote addr
```

connection.go
```diff
@@ -891,19 +891,19 @@ func (c *Connection) sendBufSize() (sendBufUsage int, sendBufSize int, _ error)
 
 	var (
 		// current send buffer usage
```
prashantv: can you delete this comment and below (they are stale from when I was declaring the variables here), can probably make this block just:
```
var sendBufLenErr, sendBufLimitErr error
```

sockio_darwin.go
```diff
@@ -0,0 +1,10 @@
+// +build darwin amd64
```
prashantv: Can you add copyright headers to this file (same for the linux file as well)

connection.go
witriew: Sure thing, done.

sockio_darwin.go
witriew: Done.

PR #772, opened 2020-03-19
Add options to allow overriding SendBufferSize pre process name prefix

fiibbb:
This adds the ability to override SendBufferSize per process name prefix. I'm using prefix instead of regex because prefix seem to be sufficient for most usecases and is significantly faster:

```
➜  regexp_test go test -v -bench=.
goos: darwin
goarch: amd64
pkg: github.com/fiibbb/regexp_test
BenchmarkRegex
BenchmarkRegex-12                5526828               208 ns/op               0 B/op          0 allocs/op
BenchmarkPrefix
BenchmarkPrefix-12              344802882                3.46 ns/op            0 B/op          0 allocs/op
BenchmarkSuffix
BenchmarkSuffix-12              369626205                3.21 ns/op            0 B/op          0 allocs/op
BenchmarkContains
BenchmarkContains-12            142756165                8.41 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/fiibbb/regexp_test   7.896s
```

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -48,9 +49,9 @@ const (
 	// is specified in the context.
 	DefaultConnectTimeout = 5 * time.Second
 
-	// defaultConnectionBufferSize is the default size for the connection's
-	// read and write channels.
-	defaultConnectionBufferSize = 512
+	// DefaultConnectionBufferSize is the default size for the connection's read
+	//and write channels.
+	DefaultConnectionBufferSize = 512
 )
 
 // PeerVersion contains version related information for a specific peer.
@@ -124,6 +125,13 @@ func (e errConnectionUnknownState) Error() string {
 	return fmt.Sprintf("connection is in unknown state: %v at %v", e.state, e.site)
 }
 
+// SendBufferSizeOverride is used for overriding per-process send buffer channel size for a
+// connection, using process name prefix matching.
+type SendBufferSizeOverride struct {
+	ProcessNamePrefix string
+	SendBufferSize    int
+}
+
 // ConnectionOptions are options that control the behavior of a Connection
 type ConnectionOptions struct {
 	// The frame pool, allowing better management of frame buffers. Defaults to using raw heap.
@@ -134,6 +142,10 @@ type ConnectionOptions struct {
 
 	// The size of send channel buffers. Defaults to 512.
 	SendBufferSize int
+
+	// Per-process name prefix override for SendBufferSize
+	// Note that order matters, if there are multiple matches, the first one is used.
+	SendBufferSizeOverrides []SendBufferSizeOverride
 
 	// The type of checksum to use when sending messages.
 	ChecksumType ChecksumType
@@ -258,10 +270,19 @@ func (co ConnectionOptions) withDefaults() ConnectionOptions {
 		co.FramePool = DefaultFramePool
 	}
 	if co.SendBufferSize <= 0 {
-		co.SendBufferSize = defaultConnectionBufferSize
+		co.SendBufferSize = DefaultConnectionBufferSize
 	}
 	co.HealthChecks = co.HealthChecks.withDefaults()
 	return co
+}
+
+func (co ConnectionOptions) getSendBufferSize(processName string) int {
+	for _, override := range co.SendBufferSizeOverrides {
+		if strings.HasPrefix(processName, override.ProcessNamePrefix) {
+			return override.SendBufferSize
+		}
+	}
+	return co.SendBufferSize
 }
 
 func (ch *Channel) setConnectionTosPriority(tosPriority tos.ToS, c net.Conn) error {
@@ -311,7 +332,7 @@ func (ch *Channel) newConnection(conn net.Conn, initialID uint32, outboundHP str
 		connDirection:      connDirection,
 		opts:               opts,
 		state:              connectionActive,
-		sendCh:             make(chan *Frame, opts.SendBufferSize),
+		sendCh:             make(chan *Frame, opts.getSendBufferSize(remotePeer.ProcessName)),
 		stopCh:             make(chan struct{}),
 		localPeerInfo:      peerInfo,
 		remotePeerInfo:     remotePeer,
--- a/connection_test.go
+++ b/connection_test.go
@@ -1260,6 +1260,66 @@ func TestLastActivityTimePings(t *testing.T) {
 	})
 }
 
+func TestSendBufferSize(t *testing.T) {
+	opts := testutils.NewOpts().SetSendBufferSize(512).SetSendBufferSizeOverrides([]SendBufferSizeOverride{
+		{"abc", 1024},
+		{"abcd", 2048}, // This should never match, since we match the list in order.
+		{"xyz", 3072},
+	})
+	tests := []struct {
+		processName          string
+		expectSendChCapacity int
+	}{
+		{
+			processName:          "abc",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "abcd",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "bcd",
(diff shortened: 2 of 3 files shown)


connection.go
```diff
@@ -302,6 +306,13 @@ func (ch *Channel) newConnection(conn net.Conn, initialID uint32, outboundHP str
 	log = log.WithFields(LogField{"connectionDirection", connDirection})
 	peerInfo := ch.PeerInfo()
 
+	sendBufferSize := opts.SendBufferSize
+	for processNamePrefix, override := range opts.SendBufferSizeOverrides {
```
prashantv: move this logic into a separate function (can be on the options type) that takes in the `remotePeer`, ideally this should just be `sendBufferSize := opts.GetSendBufferSize(remotePeer)` (and we could inline that into the `sendCh` constructor too)

connection_test.go
```diff
@@ -1260,6 +1260,44 @@ func TestLastActivityTimePings(t *testing.T) {
...
+		{
+			processName:          "abcd",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "bcd",
+			expectSendChCapacity: 512,
+		},
+	}
+	for _, testCase := range tests {
```
prashantv: nit: for consistency, use `tt`

connection_test.go
```diff
@@ -1260,6 +1260,44 @@ func TestLastActivityTimePings(t *testing.T) {
...
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "abcd",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "bcd",
+			expectSendChCapacity: 512,
+		},
```
prashantv: add test case where abc is the suffix, or is just a substring

connection.go
prashantv: Once you make it a function on the options, you can also write a unit test that is more functional (but still keep the end-to-end test with a couple of cases)

fiibbb: Not sure why `connection_test.go` belongs to a different package -- this means `GetSendBufferSize` has to be exported for unit tests. Feels cleaner unit tests is a small gain compared to adding the downside of unnecessary public API? (But yea pulled the logic into its own function.

connection.go
```diff
@@ -135,6 +136,9 @@ type ConnectionOptions struct {
 	// The size of send channel buffers. Defaults to 512.
 	SendBufferSize int
 
+	// Per-process name prefix override for SendBufferSize
+	SendBufferSizeOverrides map[string]int
```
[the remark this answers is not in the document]
fiibbb: Yea this had to be done not just for readability but also functionality. Order matters in this case so map is a bad idea.

connection.go
prashantv: Separate package is used for the end-to-end test helpers (`testutils` package has helpers for creating channels and test verification, which we are used by all subpackages, but we don't want to export as part of this package's API)

If you want unit tests, you can use `connection_internal_test.go` which is in the same package for "internal" test that need access to the unexported API.

connection_test.go
```diff
@@ -1260,6 +1260,64 @@ func TestLastActivityTimePings(t *testing.T) {
 	})
 }
 
+func TestSendBufferSize(t *testing.T) {
+	opts := testutils.NewOpts().SetSendBufferSize(512).SetSendBufferSizeOverrides([]SendBufferSizeOverride{
+		{"abc", 1024},
+		{"abcd", 2048},
```
prashantv: add a comment that this will never match, since we look for matches in order

connection_test.go
```diff
@@ -1260,6 +1260,64 @@ func TestLastActivityTimePings(t *testing.T) {
...
+		},
+		{
+			processName:          "abcde",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "xyzabc",
+			expectSendChCapacity: 3072,
+		},
+	}
```
prashantv: nit: suggest a blank line after `}` to separate the test data from the run logic

connection_test.go
```diff
@@ -1260,6 +1260,64 @@ func TestLastActivityTimePings(t *testing.T) {
...
+	for _, tt := range tests {
```
prashantv: use `t.Run` to name each test case

connection_test.go
```diff
@@ -1260,6 +1260,64 @@ func TestLastActivityTimePings(t *testing.T) {
...
+		testutils.WithTestServer(t, opts.NoRelay(), func(tb testing.TB, ts *testutils.TestServer) {
```
prashantv: you have `NoRelay` here. Can you move it up to the top where opts is declared, and add a comment about why we use `NoRelay`, e.g., `// Relay requires test logic to fork when checking incoming connections` or something

connection_test.go
```diff
@@ -1260,6 +1260,64 @@ func TestLastActivityTimePings(t *testing.T) {
...
+
+			// Send an 'echo' to establish the connection.
+			testutils.RegisterEcho(ts.Server(), nil)
+			require.NoError(t, testutils.CallEcho(client, ts.HostPort(), ts.ServiceName(), nil))
+
+			// WithTestSever will test with and without relay.
+			if opts.DisableRelay {
+				assert.Equal(t, tt.expectSendChCapacity, getConnection(t, ts.Server(), inbound).SendChCapacity)
+			} else {
+				assert.Equal(t, tt.expectSendChCapacity, getConnection(t, ts.Relay(), inbound).SendChCapacity)
```
prashantv: this `else` case should not be needed if we're using `NoRelay`

connection.go
```diff
@@ -124,6 +125,12 @@ func (e errConnectionUnknownState) Error() string {
 	return fmt.Sprintf("connection is in unknown state: %v at %v", e.state, e.site)
 }
 
+// SendBufferSizeOverride is used for overriding per process name prefix send buffer channel size.
```
jaricftw: nit: comments can be improved, and call out this is per-connection, maybe: 
`SendBufferSizeOverride is used for overriding per-process send buffer channel size for a connection, using process name prefix matching.`

connection_test.go
fiibbb: Actually this was a mistake. We do want to test with and without relay. Changed the if condition to to `ts.HasRelay()`.

connection.go
fiibbb: Looks like moving to `connection_internal_test.go` will introduce circular import. I'll just leave it as is unless there's a strong opinion on unit tests vs e2e tests.

connection_test.go
```diff
@@ -1260,6 +1260,66 @@ func TestLastActivityTimePings(t *testing.T) {
...
+			processName:          "abc",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "abcd",
+			expectSendChCapacity: 1024,
+		},
+		{
+			processName:          "bcd",
+			expectSendChCapacity: 512,
```
witriew: nit: Use `defaultConnectionBufferSize`.

testutils/channel_opts.go
```diff
@@ -133,6 +133,12 @@ func (o *ChannelOpts) SetSendBufferSize(bufSize int) *ChannelOpts {
 	return o
 }
 
+// SetSendBufferSizeOverrides sets the SendBufferOverrides in DefaultConnectionOptions.
+func (o *ChannelOpts) SetSendBufferSizeOverrides(overrides []tchannel.SendBufferSizeOverride) *ChannelOpts {
+	o.DefaultConnectionOptions.SendBufferSizeOverrides = overrides
```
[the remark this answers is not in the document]
fiibbb: Need a way for user to specify order though -- and you can't do that in a map without complicating the API too much?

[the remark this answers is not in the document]
fiibbb: At least for our use case, we will also have validation in our dynamic config repo.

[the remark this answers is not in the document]
alxn: That's true. Is that the goal of the prefixes to be executed in order? If so lgtm.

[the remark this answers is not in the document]
fiibbb: Yea the prefixes will be executed in order.

PR #776, opened 2020-05-14
relay: Limit connection timeouts to avoid stalls

prashantv:
Ref T5890079

If we need to connect in the relay path, we're basically stalling the
connection (all calls on the connection will be backed up behind the
connection). If a call has a large timeout (e.g., 5 seconds), and the
backend is unavailable, leading to a timeout, then all calls behind will
likely be timed out by the time we've started processing them.

Limit the connection timeout to a shorter value than the call timeout
since connection time is known at the infrastructural layer.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -419,7 +419,7 @@ func (p *Peer) GetConnection(ctx context.Context) (*Connection, error) {
 
 // getConnectionRelay gets a connection, and uses the given timeout to lazily
 // create a context if a new connection is required.
-func (p *Peer) getConnectionRelay(timeout time.Duration) (*Connection, error) {
+func (p *Peer) getConnectionRelay(callTimeout, relayMaxConnTimeout time.Duration) (*Connection, error) {
 	if conn, ok := p.getActiveConn(); ok {
 		return conn, nil
 	}
@@ -431,6 +431,12 @@ func (p *Peer) getConnectionRelay(timeout time.Duration) (*Connection, error) {
 	// Check active connections again in case someone else got ahead of us.
 	if activeConn, ok := p.getActiveConn(); ok {
 		return activeConn, nil
+	}
+
+	// Use the lower timeout value of the call timeout and the relay connection timeout.
+	timeout := callTimeout
+	if timeout > relayMaxConnTimeout && relayMaxConnTimeout > 0 {
+		timeout = relayMaxConnTimeout
 	}
 
 	// When the relay creates outbound connections, we don't want those services
--- a/relay.go
+++ b/relay.go
@@ -362,7 +362,6 @@ func (r *Relayer) getDestination(f lazyCallReq, call RelayCall) (*Connection, bo
 		return nil, false, errBadRelayHost
 	}
 
-	// TODO: Should connections use the call timeout? Or a separate timeout?
 	remoteConn, err := peer.getConnectionRelay(f.TTL(), r.maxConnTimeout)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


relay.go
```diff
@@ -361,7 +363,7 @@ func (r *Relayer) getDestination(f lazyCallReq, call RelayCall) (*Connection, bo
 	}
 
 	// TODO: Should connections use the call timeout? Or a separate timeout?
```
cinchurge, line 365: should this comment be removed?

relay.go
```diff
@@ -169,8 +169,9 @@ const (
 
 // A Relayer forwards frames.
 type Relayer struct {
-	relayHost  RelayHost
-	maxTimeout time.Duration
+	relayHost      RelayHost
+	maxTimeout     time.Duration
+	maxConnTimeout time.Duration
```
witriew, line 174: Should this be exposed for introspection as well?

peer.go
```diff
@@ -433,6 +433,12 @@ func (p *Peer) getConnectionRelay(timeout time.Duration) (*Connection, error) {
 		return activeConn, nil
 	}
 
+	// Use the lower timeout value of the call timeout and the relay connection timeout.
+	timeout := callTimeout
+	if timeout > relayMaxConnTimeout && relayMaxConnTimeout > 0 {
+		timeout = relayMaxConnTimeout
+	}
```
witriew, line 440: This looks like it gets plumbed through by `NewContextBuilder` as the context's timeout, but `NewContextBuilder` seems to support a `ConnectTimeout`, which is used by `Connect`.

How do those differ for this situation? Don't we want to use `ConnectTimeout` for this case?

relay.go
prashantv: Good point, added introspection.

peer.go
prashantv: Good digging, the `ConnectTimeout` is intended for calls where you want to limit how long is spent connecting. E.g., I may have a call that has a 5 second timeout, but I only want 1 second spent connecting (so there's enough time to try on a different peer).

The `Connect` logic prefers to use the `ConnectTimeout` is specified, but otherwise, falls back to the context timeout:
https://github.com/uber/tchannel-go/blob/dev/channel.go#L563

In our case, since we're creating the context purely for the connection timeout, there's no need to specify the `ConnectTimeout` (since it'll fallback to the context timeout).

witriew: I think this makes sense - this makes it so that the relay connection fails faster with `timeout` than it normally would?

witriew:
> Ref T5890079

Do we care if we include internal references? It's internal, so it doesn't really matter, just wondering.

peer.go
prashantv: The `ConnectTimeout` is useful for standard non-relay clients (e.g., `client.CallMethod(ctx, ...)` where the method may need more time separate to connects). If you don't specify it, the `ctx` timeout is used. In this case, setting `ConnectTimeout` wouldn't have any impact, since the only purpose of the `ctx` created below is for the connect timeout.

The `ConnectTimeout` is an advanced option that makes sense for normal clients, but in the relay code path, when we create a context purely for connections, it doesn't make sense to use.

prashantv:
> > Ref T5890079
> 
> Do we care if we include internal references? It's internal, so it doesn't really matter, just wondering.

I think with other OSS repos (E.g., YARPC and Fx) we started using task IDs even if they were internal,. Since there's no harm, thought it'd be useful for us to start doing too.

PR #783, opened 2020-07-23
echung/add arg2 append

cinchurge:


relay/relay.go
```diff
@@ -56,6 +56,8 @@ type CallFrame interface {
 	// of TChannel-Thrift Arg Scheme. If no iterator is available, return
 	// io.EOF.
 	Arg2Iterator() (arg2.KeyValIterator, error)
+	// Arg2Append appends a key/val pair to arg2
+	Arg2Append(key, val []byte) error
```
prashantv, line 60: As we discussed offline, this API may be limiting if we ever want to add more than one header(without the performance impact of moving arg3 around for each key/value we had), or for other types of modifications we want.

it also may not work well with the spill-over of arg3 into a new frame

relay_messages.go
```diff
@@ -228,6 +238,40 @@ func (f lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 	return arg2.NewKeyValIterator(f.Payload[f.arg2StartOffset:f.arg2EndOffset])
 }
 
+func (f *lazyCallReq) Arg2Append(key, val []byte) error {
+	arg3 := bufPool.Get().(*bytes.Buffer)
+	arg3.Write(f.Payload[f.arg2EndOffset:])
```
prashantv, line 243: the `Payload` field is confusing as it ends up being the whole 64kb - header, even in the case that the payload is actually smaller than 64kb. You probably want `SizedPayload()` to get just the parts of the frame that are "set"

(unfortunate API)

relay_messages.go
```diff
@@ -228,6 +238,40 @@ func (f lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
...
+	}
+
+	delta := bufPool.Get().(*bytes.Buffer)
+	writeDataToBuffer(key, delta)
+	writeDataToBuffer(val, delta)
+	f.Payload = append(f.Payload[:f.arg2EndOffset], delta.Bytes()...)
+	f.arg2EndOffset += deltaLen
+
+	arg2PairCount := int(binary.BigEndian.Uint16(arg2Payload[0:2]))
+	binary.BigEndian.PutUint16(arg2Payload[0:2], uint16(arg2PairCount+1))
```
prashantv, line 260: So arg2 modification requires the following
 * Update the actual "arg2" which is what this is doing, which is relevant to the Thrift-specific encoding. (Note, we should only ever modify arg2 in this way when format or "as" header == thrift)
 * Update the length of arg2 (which is 2 bytes before arg2 starts)
 * Move arg3
 * Update the size of the frame (as it has likely gotten larger)
 * Update the checksum

cinchurge:
abandoning since the root cause is fixed in #788

PR #784, opened 2020-07-27
Add tests for validating arg2 appending

prashantv:
Related to #783


prashantv:
Split out changes into multiple PRs, #797, #798, and #799 which have all landed.

PR #786, opened 2020-07-30
relay: Report bytes sent/received via Call

prashantv:
Ref T6369485

Track the number of bytes sent/received on a per-call basis by
reporting the frame sizes in the relay path.

For a call, the flow is:

```
   S1 (caller)            S2 (callee)
  ------------------------------------
   Relay(callReq)
                     ->   Receive(callReq)

                          // Request sent to backend, receive response.

                          Relay(callRes)
   Receive(callRes)  <-
```

We could track outbound frames either at caller.Relay or callee.Receive
Inbound frames could similary by tracked in either callee.Relay or
caller.Receive.

If we track metrics only at the caller (Relay and Receive), we can
end up missing frames that are received, but dropped because the call
has timed out.

Instead, we choose to always track metrics in Relay, but we need to take
into account whether it's a request or response frame to determine
whether these are received or sent bytes.

We also need the call to be tracked as part of the relayItem on both
the caller and callee side, as we emit metrics from both sides.

This is not going to be 100% accurate, as there are other paths were
frames may be dropped (E.g., sendCh queue full), but this should work
well for the majority of use-cases.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -82,6 +82,14 @@ func newRelayItems(logger Logger) *relayItems {
 	}
 }
 
+func (ri *relayItem) reportRelayBytes(fType frameType, frameSize uint16) {
+	if fType == requestFrame {
+		ri.call.SentBytes(frameSize)
+	} else {
+		ri.call.ReceivedBytes(frameSize)
+	}
+}
+
 // Count returns the number of non-tombstone items in the relay.
 func (r *relayItems) Count() int {
 	r.RLock()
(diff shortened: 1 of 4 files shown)


prashantv:
> This seems like it'd be complicated, but should there be additional testing/validation at the call level to make sure that the number of bytes sent/received are what's expected?

Discussed offline, we're comparing the metrics to the payload size (+- 1kb) so that should cover the required validation.

PR #787, opened 2020-07-31
Fix goroutine leak in TestLargeTimeoutsAreClamped

cinchurge:
`t.Fatal()` from failures in `TestLargeTimeoutsAreClamped` caused the test to exit without cleanup. Do `assert.Fail()` and `return` instead so we don't trigger the goroutine detector (which is super noisy).

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -385,7 +385,8 @@ func TestLargeTimeoutsAreClamped(t *testing.T) {
 
 		select {
 		case <-time.After(testutils.Timeout(10 * clampTTL)):
-			t.Fatal("Failed to clamp timeout.")
+			assert.Fail(t, "Failed to clamp timeout.")
+			return
 		case <-done:
 		}
 	})


prashantv:
If I understand correctly, the goal of this change is to avoid a goroutine leak on failures, so the goroutine leak detector doesn't print a lot of noise.

The leak detector is only supposed to run on success though, and a quick test showed that it isn't printing a lot of noise in the failure case:
```
# make a change to force the test to fail
$ git diff                                                                                                                                                                                                                                                                 
diff --git a/relay_test.go b/relay_test.go
index de52b84..d847030 100644
--- a/relay_test.go
+++ b/relay_test.go
@@ -384,6 +384,8 @@ func TestLargeTimeoutsAreClamped(t *testing.T) {
                        close(done)
                }()

+               t.Fatal("test")
+
                select {
                case <-time.After(testutils.Timeout(10 * clampTTL)):
                        t.Fatal("Failed to clamp timeout.")
$ go test -v -run TestLargeTimeoutsAreClamped
=== RUN   TestLargeTimeoutsAreClamped
=== RUN   TestLargeTimeoutsAreClamped/with_relay
    TestLargeTimeoutsAreClamped/with_relay: relay_test.go:387: test
=== RUN   TestLargeTimeoutsAreClamped/with_relay_and_timer_verification
    TestLargeTimeoutsAreClamped/with_relay_and_timer_verification: relay_test.go:387: test
--- FAIL: TestLargeTimeoutsAreClamped (0.00s)
    --- FAIL: TestLargeTimeoutsAreClamped/with_relay (0.00s)
    --- FAIL: TestLargeTimeoutsAreClamped/with_relay_and_timer_verification (0.00s)
FAIL
exit status 1
FAIL    github.com/uber/tchannel-go     0.004s
```

cinchurge:
> The leak detector is only supposed to run on success though, and a quick test showed that it isn't printing a lot of noise in the failure case:

it happens with `make test`, but not with `go test`:
```
$ make test
...
=== RUN   TestRelayConcurrentCalls/with_relay
    TestRelayConcurrentCalls/with_relay: test_server.go:420: found unexpected goroutines:
        [Goroutine 4985 in state IO wait, with internal/poll.runtime_pollWait on top of the stack:
        goroutine 4985 [IO wait]:
        internal/poll.runtime_pollWait(0x5119d28, 0x72, 0x0)
        	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/netpoll.go:203 +0x55
        internal/poll.(*pollDesc).wait(0xc000cbe098, 0x72, 0x0, 0x0, 0x1849370)
        	/usr/local/Cellar/go/1.14.3/libexec/src/internal/poll/fd_poll_runtime.go:87 +0xe4
        internal/poll.(*pollDesc).waitRead(...)
        	/usr/local/Cellar/go/1.14.3/libexec/src/internal/poll/fd_poll_runtime.go:92
        internal/poll.(*FD).Accept(0xc000cbe080, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        	/usr/local/Cellar/go/1.14.3/libexec/src/internal/poll/fd_unix.go:384 +0x2c9
        net.(*netFD).accept(0xc000cbe080, 0x1087ca7, 0x1087ca7, 0x100a280)
        	/usr/local/Cellar/go/1.14.3/libexec/src/net/fd_unix.go:238 +0x56
        net.(*TCPListener).accept(0xc000d88d20, 0x10ac45c, 0x10ddb5f, 0xc000a709a0)
        	/usr/local/Cellar/go/1.14.3/libexec/src/net/tcpsock_posix.go:139 +0x50
        net.(*TCPListener).Accept(0xc000d88d20, 0x1945e80, 0xc0004d4028, 0x0, 0x0)
        	/usr/local/Cellar/go/1.14.3/libexec/src/net/tcpsock.go:261 +0x50
...
```

cinchurge:
abandoning since root cause if fixed in #788

PR #791, opened 2020-08-04
Enable arg2 appends in the relayer

cinchurge:


cinchurge:
abandoned in favor of #794

PR #792, opened 2020-08-05
Replace manual "cur" pointer with typed.ReadBuffer

cinchurge:
The "cur" pointer approach has made the code very difficult to follow and has likely resulted in a missed error check in `KeyValIterator.Next()`.

This change migrates `newLazyCallReq()` and `KeyValIterator.Next()` to `typed.ReadBuffer` instead for better readability.

Benchmarks indicate that there are no additional allocs resulting from this change.

`BenchmarkRelay*` benchmarks:
```
   17408	     71408 ns/op	  14.34 MB/s	     446 B/op	      12 allocs/op
   21661	     55907 ns/op	  18.32 MB/s	     506 B/op	      12 allocs/op
   13016	     86108 ns/op	  47.57 MB/s	     497 B/op	      12 allocs/op
   15798	     75712 ns/op	  13.52 MB/s	     453 B/op	      12 allocs/op
   24147	     48498 ns/op	  21.11 MB/s	     496 B/op	      12 allocs/op
   14484	     81832 ns/op	  50.05 MB/s	     499 B/op	      12 allocs/op
```

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -38,7 +38,10 @@ var (
 	_argSchemeKeyBytes       = []byte(ArgScheme)
 	_tchanThriftValueBytes   = []byte(Thrift)
 
-	errBadArg2Len = errors.New("bad Arg2 length")
+	errBadHeaderLen   = errors.New("bad header length")
+	errBadChecksumLen = errors.New("bad checksum length")
+	errBadArg1Len     = errors.New("bad Arg1 length")
+	errBadArg2Len     = errors.New("bad Arg2 length")
 )
 
 const (
@@ -103,7 +106,7 @@ type lazyCallReq struct {
 
 // TODO: Consider pooling lazyCallReq and using pointers to the struct.
 
-func newLazyCallReq(f *Frame) lazyCallReq {
+func newLazyCallReq(f *Frame) (lazyCallReq, error) {
 	if msgType := f.Header.messageType; msgType != messageTypeCallReq {
 		panic(fmt.Errorf("newLazyCallReq called for wrong messageType: %v", msgType))
 	}
@@ -134,13 +137,25 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 		}
 	}
 
+	if rbuf.Err() != nil {
+		return lazyCallReq{}, errBadHeaderLen
+	}
+
 	// csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2
 	checkSumType := ChecksumType(rbuf.ReadSingleByte())
 	rbuf.ReadBytes(checkSumType.ChecksumSize())
 
+	if rbuf.Err() != nil {
+		return lazyCallReq{}, errBadChecksumLen
+	}
+
 	// arg1~2
 	arg1Len := int(rbuf.ReadUint16())
 	cr.method = rbuf.ReadBytes(arg1Len)
+
+	if rbuf.Err() != nil {
+		return lazyCallReq{}, errBadArg1Len
+	}
 
 	// arg2~2
 	arg2Len := int(rbuf.ReadUint16())
@@ -150,7 +165,11 @@ func newLazyCallReq(f *Frame) lazyCallReq {
 	// arg2 is fragmented if we don't see arg3 in this frame.
 	cr.isArg2Fragmented = rbuf.BytesRemaining() == 0 && cr.HasMoreFragments()
 
-	return cr
+	if rbuf.Err() != nil {
+		return lazyCallReq{}, errBadArg2Len
+	}
+
+	return cr, nil
 }
 
 // Caller returns the name of the originator of this callReq.
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/uber/tchannel-go/thrift/arg2"
+	"github.com/uber/tchannel-go/typed"
 )
 
 var (
@@ -96,6 +97,7 @@ type lazyCallReq struct {
 	caller, method, delegate, key, as []byte
 
 	arg2StartOffset, arg2EndOffset int
+	arg2                           []byte
 	isArg2Fragmented               bool
 }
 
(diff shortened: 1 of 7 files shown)


relay_messages.go
```diff
@@ -96,6 +97,7 @@ type lazyCallReq struct {
 	caller, method, delegate, key, as []byte
 
 	arg2StartOffset, arg2EndOffset int
+	arg2                           []byte
```
prashantv, line 100: instead of a field, can this be a method that uses the `StartOffset` and `EndOffset` to slice into the payload? Likely not a big performance impact, but want to try and keep the fields here minimal ideally (and right now having the start/end offset as well as the byte buf seems a little redundant)

relay_messages.go
```diff
@@ -110,19 +112,16 @@ func newLazyCallReq(f *Frame) lazyCallReq {
...
-	numHeaders := int(f.Payload[headerStart])
-	cur := int(headerStart) + 1
-	for i := 0; i < numHeaders; i++ {
-		keyLen := int(f.Payload[cur])
-		cur++
-		key := f.Payload[cur : cur+keyLen]
-		cur += keyLen
+	headerStartIndex := _serviceLenIndex + 1 /* length byte */ + serviceLen
+
+	rbuf := typed.NewReadBuffer(f.SizedPayload()[headerStartIndex:])
```
prashantv, line 117: heads up that when using this, we should be careful to check `rbuf.Err()`,
https://godoc.org/github.com/uber/tchannel-go/typed#ReadBuffer.Err

the other methods will set the error if there's not enough bytes to read etc, so we might not notice if we don't check that

thrift/arg2/kv_iterator.go
```diff
@@ -53,32 +58,34 @@ func (i KeyValIterator) Next() (KeyValIterator, error) {
 		return KeyValIterator{}, io.EOF
 	}
 
-	arg2Len := len(i.arg2Payload)
-	cur := i.valueOffset + i.valueLen
-	if cur+2 > arg2Len {
-		return KeyValIterator{}, fmt.Errorf("invalid key offset %v (arg2 len %v)", cur, arg2Len)
+	if i.rbuf.BytesRemaining() < 2 {
```
prashantv, line 61: previously, we had a bunch of error checks to avoid panics, although now with readbuf, since it stores errors internally, we can consider only checking errors once at the end if the `rbuf.Err() != nil`, what do you think?

cinchurge: i considered that, but the errors returned by `rbuf.Err()` lacks helpful details for surfacing the issue, so decided it was better to maintain the current set of error checks. what we could do though, is to do a `rbuf.Err()` check at the end, *in addition* to the current set of checks to avoid missing any edge cases.

relay_messages.go
cinchurge: actually i would've preferred to deprecate the offsets completely since it really isn't needed, and we will be adding slices for arg1 and arg2 later for the fragmented relay change. the problem is that the offsets are already exposed through the `CallFrame` interface, and removing them would be a breaking change :( would such breakage be acceptable in this case?

relay_messages.go
cinchurge: yea will need to think about this a little more, not sure if we should do granular checks or just a single check at the end

relay_messages.go
prashantv: Ah, the `CallFrame` interface (and all relay-specific interfaces) are considered internal, and not part of any API contracts. Since the only consumer is Muttley, we can make changes here (we'll just need to update the Muttley code at the same time as the dependency upgrade).

Just check how we're using it in Muttley to ensure we don't need the offset values directly (I think we just need the length which is also fine to get from the `[]byte`).

A micro-optimization note: a slice is 3 values: a data pointer, length, and a capacity. This often means that 2 ints will be more efficient than a slice or `[]byte` in this case. It's likely not going to cause any noticeable performance impact given how many `[]byte` we already store in this case.

relay_messages.go
prashantv: I think a single check at the end is probably fine, since we can't really react to intermediate errors differently (in either case, it's an invalid frame and we'll reject it)

thrift/arg2/kv_iterator.go
prashantv: I'm OK with doing both checks, or just checking `rbuf.Err()`. Agree that it doesn't provide many details, but given that we don't expect malformed frames, and providing more details on how the frame is malformed likely won't help clients anyway, it doesn't seem too important to provide that the more granular errors.

cinchurge: actually looking at the code a bit more, the final check might not be necessary, as the existing checks already cover possible failures. i vote for leaving the way it currently is.

relay_messages.go
cinchurge: yea if performance isn't impacted materially, i think going with the more readable/less complex implementation would be more beneficial overall.

thrift/arg2/kv_iterator.go
cinchurge: oops didn't see your comments before posting this. in that case let's just consistently do a single check at the end.

relay_messages.go
cinchurge: turns out it's super involved to deprecate the offsets, even just within tchannel-go. will drop the arg2 slice and leave a TODO for later.

relay_messages.go
```diff
@@ -136,21 +134,28 @@ func newLazyCallReq(f *Frame) lazyCallReq {
...
+	cr.method = rbuf.ReadBytes(arg1Len)
 
 	// arg2~2
-	cur += arg1Len
-	cr.arg2StartOffset = cur + 2
-	cr.arg2EndOffset = cr.arg2StartOffset + int(binary.BigEndian.Uint16(f.Payload[cur:cur+2]))
+	arg2Len := int(rbuf.ReadUint16())
+	cr.arg2StartOffset = int(headerStartIndex) + rbuf.BytesRead()
+	// TODO(echung): deprecate the arg2 offsets and store just the arg2 slice instead
+	// Read pass arg2
```
prashantv, line 148: nit: did you mean `read past arg2`? or

thrift/arg2/kv_iterator.go
```diff
@@ -5,20 +5,21 @@
...
 // which has key/value pairs (k~2 v~2).
 // NOTE: to be optimized for performance, we try to limit the allocation
 // done in the process of iteration.
 type KeyValIterator struct {
-	arg2Payload           []byte
-	leftPairCount         int
-	keyOffset             int
-	valueOffset, valueLen int
+	arg2Len       int
+	rbuf          *typed.ReadBuffer
```
prashantv, line 19: i think the kv iterator isn't included in the current relay benchmark, can you do a quick benchmark of just the KeyValueIterator with the test frames to see if there's any performance impact to using the `*typed.ReadBuffer` here?

cinchurge: turns out we do get 1 alloc with this:
`dev`:
```
BenchmarkKeyValIterator-8   	16034449	        73.0 ns/op	       0 B/op	       0 allocs/op
```
this branch:
```
BenchmarkKeyValIterator-8   	10106662	       116 ns/op	      64 B/op	       1 allocs/op
```
which makes sense, since we need to hold on to a `*typed.ReadBuffer` in the iterator. Will revert the change for the iterator and keep the change in `lazyCallReq`

PR #794, opened 2020-08-06
Enable arg2 appends in the relayer

cinchurge:
This change updates the `relay.CallFrame` interface to include a `.Arg2Append(key, val []byte)` method, which can be called from the user-provided `RelayHost.Start()` to e.g. inject auth tokens during relays.


cinchurge:
gonna fill in tests

relay.go
```diff
@@ -654,6 +664,80 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+
+	return nil
+}
+
+func writeArg2WithAppends(w io.WriteCloser, arg2 []byte, appends []keyVal) (err error) {
+	defer w.Close()
+
+	var nh uint16
+	if len(arg2) > 0 {
+		nh = binary.BigEndian.Uint16(arg2[:2])
```
prashantv, line 698: what if `len(arg2) == 1`? this would panic in that case.

i think if the length of arg2 < 2, we should return an error immediately.

relay.go
```diff
@@ -654,6 +664,80 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+	}
+	nh += uint16(len(appends))
+	if err := writeUint16(w, nh); err != nil {
+		return fmt.Errorf("write arg2 nh: %v", err)
+	}
+
+	if len(arg2) > 0 {
```
prashantv, line 705: we're checking `len(arg2) > 0` twice, which is a bit of a code smell.

i think returning an error in the < 2 case might avoid this completely

relay.go
```diff
@@ -654,6 +664,80 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+		if _, err := w.Write(arg2[2:]); err != nil {
```
prashantv, line 706: add comments to explain this code -- it makes sense to someone who has the protocol as top of mind, but will be hard to understand in a few months.

relay.go
```diff
@@ -654,6 +664,80 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+		}
+		if err := writeLen16Data(w, kv.val); err != nil {
+			return fmt.Errorf("append arg2 val: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func writeLen16Data(w io.Writer, b []byte) error {
```
prashantv, line 722: should we create a `typed.Writer` with this logic? we could follow the same pattern and defer error checking till the very end, so the logic is cleaner to follow

relay.go
```diff
@@ -654,6 +664,80 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+	return nil
+}
+
+func writeUint16(w io.Writer, n uint16) error {
+	var sizeBuf [2]byte
+	binary.BigEndian.PutUint16(sizeBuf[:], n)
+	if _, err := w.Write(sizeBuf[:]); err != nil {
+		return err
+	}
+	return nil
```
prashantv, line 738: if we're not wrapping the error, can simplify to:

```suggestion
	_, err := w.Write(sizeBuf[:]); err != nil {
	return err
```

relay.go
```diff
@@ -654,6 +664,80 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+
+	return nil
+}
+
+func writeLen16Data(w io.Writer, b []byte) error {
+	if err := writeUint16(w, uint16(len(b))); err != nil {
+		return fmt.Errorf("write data length: %v", err)
+	}
+	if _, err := w.Write(b); err != nil {
+		return fmt.Errorf("write data: %v", err)
```
prashantv, line 727: should we use `%w`?

relay.go
```diff
@@ -694,3 +778,82 @@ func validateRelayMaxTimeout(d time.Duration, logger Logger) time.Duration {
...
+		checksum:    checksum,
+		contents:    contents,
+		frame:       frame,
+	}, contents.Err()
+}
+
+func (rfs *relayFragmentSender) flushFragment(f *writableFragment) error {
+	frame, ok := f.frame.(*Frame)
+	if !ok {
+		return fmt.Errorf("got unexpected frame type: %t", f.frame)
```
prashantv, line 847: should be `%T` for type.

however, this whole thing is unnecessary, not sure why we had a `interface{}`, cleaned up:
https://github.com/uber/tchannel-go/pull/804

relay_messages.go
```diff
@@ -91,17 +91,22 @@ func (cr lazyCallRes) OK() bool {
...
-	caller, method, delegate, key, as []byte
-
 	checksumTypeOffset             uint16
 	arg2StartOffset, arg2EndOffset uint16
 	arg3StartOffset                uint16
 
-	checksumType     ChecksumType
-	isArg2Fragmented bool
+	caller, method, delegate, key, as []byte
+	arg2appends                       []keyVal
```
prashantv, line 107: nit: rename to `arg2Appends` (camel case)

relay_messages_test.go
```diff
@@ -122,6 +122,8 @@ func (cr testCallReq) frameWithParams(p testCallReqParams) *Frame {
 	}
 	payload.WriteUint16(uint16(arg2Len))
 	payload.WriteBytes(p.arg2Buf)
+	payload.WriteUint16(uint16(len(p.arg3Buf)))
+	payload.WriteBytes(p.arg3Buf)
```
prashantv, line 126: we already write out arg3 below, rebase issue?

testutils/call.go
```diff
@@ -158,6 +161,22 @@ func (f *FakeCallFrame) Arg2Iterator() (arg2.KeyValIterator, error) {
...
+	}
+	kv[string(key)] = string(val)
+
+	iterator, err := arg2.NewKeyValIterator(thriftarg2test.BuildKVBuffer(kv))
+	require.NoError(f.tb, err)
+
+	f.arg2KVIterator = iterator
+
+	return
+}
```
prashantv, line 178: i think we should just keep a list of appends, we don't want to really mutate arg2 (that is also not the behaviour of the real call frame)

testutils/channel.go
```diff
@@ -54,6 +54,11 @@ func NewServerChannel(opts *ChannelOpts) (*tchannel.Channel, error) {
 		return nil, fmt.Errorf("NewChannel failed: %v", err)
 	}
 
+	connOpts := ch.ConnectionOptions()
+	if opts.ConnContext != nil {
+		connOpts.ConnContext = opts.ConnContext
+	}
```
prashantv, line 60: rebase issue?

testutils/thriftarg2test/arg2_kv_builder.go
```diff
@@ -21,3 +21,16 @@ func BuildKVBuffer(kv map[string]string) []byte {
 	}
 	return buf[:wb.BytesWritten()]
 }
+
+// ReadKVBuffer converts an arg2 buffer to a string map
+func ReadKVBuffer(b []byte) map[string]string {
```
prashantv, line 26: can we add unit tests?

relay.go
cinchurge: there are some cases where len(arg2) == 0 (no arg2), so I think we'll still have to accommodate that. I'll change the check to avoid the panic

prashantv: I'm not sure I understand when `len(arg2) == 0`, since that seems like a protocol error. No arg2 should result in a length prefix of 0 (2 bytes).

cinchurge: you're right. I somehow ran into some test failures before where it seemed like it was expecting arg2 to be empty, but apparently it was due to some other bug in prior versions. will drop the check

relay.go
cinchurge: we'd be using a mix of %v and %w in the code base, is that ok? or should we make a separate change to update all %v's to %w?

cinchurge:
bump

relay.go
```diff
@@ -467,12 +474,16 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) error {
 
 	f.Header.ID = destinationID
 	call.SentBytes(f.Frame.Header.FrameSize())
```
prashantv, line 476: we're no longer tracking sent bytes accurately when appends are involved.

can you check if this is easy to fix as part of this change, and if not, add a TODO and fix as a follow up?

relay.go
```diff
@@ -654,6 +665,61 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+
+	if err := NewArgWriter(fragWriter.ArgWriter(true)).Write(f.arg3()); err != nil {
+		return errors.New("arg3 write failed")
+	}
+
+	return nil
+}
+
+func writeArg2WithAppends(w io.WriteCloser, arg2 []byte, appends []keyVal) (err error) {
+	defer w.Close()
```
prashantv, line 696: should we be closing if arg2 is fragmented?

actually, is arg2 is fragmented, i thought we don't support appending arg2, so the check on line 684 probably should happen earlier and return an error?

relay.go
```diff
@@ -654,6 +665,61 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+	if err := writeArg2WithAppends(arg2Writer, f.arg2(), f.arg2Appends); err != nil {
+		return fmt.Errorf("write arg2: %v", err)
+	}
+	if f.isArg2Fragmented {
+		return nil
+	}
+
+	if err := NewArgWriter(fragWriter.ArgWriter(true)).Write(f.arg3()); err != nil {
+		return errors.New("arg3 write failed")
+	}
```
prashantv, line 690: does using `NewArgWriter` have a performance overhead / allocs? It's a little bit more verbose to do it without the arg writer so want to understand if there's a performance impact here

relay.go
prashantv, line 696: we're losing errors on close here which may not be safe?

testutils/call.go
```diff
@@ -95,8 +96,14 @@ func NewIncomingCall(callerName string) tchannel.IncomingCall {
 	return &FakeIncomingCall{CallerNameF: callerName}
 }
 
+// KeyVal is a pair of keyword and values which can be appended to arg2
+type KeyVal struct {
+	Key, Val []byte
```
prashantv, line 101: should this just be `string`? we use `[]byte` for performance but the test versions just store strings, see below for service/method/caller

testutils/thriftarg2test/arg2_kv.go
```diff
@@ -21,3 +21,16 @@ func BuildKVBuffer(kv map[string]string) []byte {
...
+func ReadKVBuffer(b []byte) map[string]string {
+	rbuf := typed.NewReadBuffer(b)
+	nh := rbuf.ReadUint16()
+	retMap := make(map[string]string, nh)
+	for i := uint16(0); i < nh; i++ {
+		key := rbuf.ReadLen16String()
+		val := rbuf.ReadLen16String()
+		retMap[key] = val
+	}
+	return retMap
```
prashantv, line 35: should we verify if there's any data left in the `[]byte`? we should verify that the bytes are fully consumed for testing

typed/writer_test.go
```diff
@@ -0,0 +1,81 @@
+package typed
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
prashantv, line 8: nit: delete blank between imports

typed/writer_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type dummyWriter struct {
+	calls        int
+	bytesWritten []byte
+	retError     []string
```
prashantv, line 15: suggest using a `map` to make this easier to work with in tests

cinchurge:
Difference in relay benchmarks is within the noise level:
```
$ cat benchmark.log | grep allocs
   14970	     77299 ns/op	  13.25 MB/s	     441 B/op	      12 allocs/op
   16476	     64064 ns/op	  15.98 MB/s	     485 B/op	      12 allocs/op
   13135	     97519 ns/op	  42.00 MB/s	     502 B/op	      12 allocs/op
$ cat baseline.log | grep allocs
   14842	     82737 ns/op	  12.38 MB/s	     435 B/op	      12 allocs/op
   17737	     59040 ns/op	  17.34 MB/s	     487 B/op	      12 allocs/op
   11094	    100041 ns/op	  40.94 MB/s	     489 B/op	      12 allocs/op
```
There is a slight increase in bytes allocated for lazyCallReq creation, though that is expected:
```
$ cat benchmark.log
BenchmarkCallReqFrame-8   	 4555321	       238 ns/op	     176 B/op	       1 allocs/op
$ cat baseline.log
BenchmarkCallReqFrame-8   	 4975362	       240 ns/op	     144 B/op	       1 allocs/op
```

typed/writer_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.msg, func(t *testing.T) {
+			w := NewWriter(tt.w)
+			w.err = tt.previousError
+
+			w.WriteUint16(1)
+			w.WriteBytes([]byte{2})
+			w.WriteLen16Bytes([]byte{4, 5, 6})
```
prashantv, line 72: can you declare these at the top, so it's easier to understand the test cases. e.g., 
```
writes := func(w *typed.Writer) {
  w.WriteUint16(1)
  ...
}
```

prashantv:
> Difference in relay benchmarks is within the noise level:
> 
> ```
> $ cat benchmark.log | grep allocs
>    14970	     77299 ns/op	  13.25 MB/s	     441 B/op	      12 allocs/op
>    16476	     64064 ns/op	  15.98 MB/s	     485 B/op	      12 allocs/op
>    13135	     97519 ns/op	  42.00 MB/s	     502 B/op	      12 allocs/op
> $ cat baseline.log | grep allocs
>    14842	     82737 ns/op	  12.38 MB/s	     435 B/op	      12 allocs/op
>    17737	     59040 ns/op	  17.34 MB/s	     487 B/op	      12 allocs/op
>    11094	    100041 ns/op	  40.94 MB/s	     489 B/op	      12 allocs/op
> ```
> 
> There is a slight increase in bytes allocated for lazyCallReq creation, though that is expected:
> 
> ```
> $ cat benchmark.log
> BenchmarkCallReqFrame-8   	 4555321	       238 ns/op	     176 B/op	       1 allocs/op
> $ cat baseline.log
> BenchmarkCallReqFrame-8   	 4975362	       240 ns/op	     144 B/op	       1 allocs/op
> ```

Since we don't modify the relay without modification path, this is expected. What is the performance impact of the appends though?

cinchurge:
gonna add tests for the sentBytes change

testutils/call.go
cinchurge: `CallFrame.Arg2Append()` accepts `[]byte`, so we'd have to convert the inputs to strings when storing them. probably a bit redundant?

relay.go
cinchurge:
```
$ cat withappends.log | grep allocs
   10214	    114906 ns/op	   8.91 MB/s	    1025 B/op	      23 allocs/op
$ cat withappends-noargwriterhelper.log | grep allocs
   10196	    122726 ns/op	   8.34 MB/s	    1025 B/op	      23 allocs/op
```
seems there isn't much of a difference...

cinchurge:
> Since we don't modify the relay without modification path, this is expected. What is the performance impact of the appends though?
```
$ cat withappends.log | grep allocs
   10214	    114906 ns/op	   8.91 MB/s	    1025 B/op	      23 allocs/op
$ cat withoutappends.log | grep allocs
    9610	    124755 ns/op	   8.21 MB/s	     475 B/op	      12 allocs/op
```
there is a rather big hit on allocs, but since we're allocating a completely new frame I think it's still somewhat reasonable. I'm slightly worried about the ns/op, which seems to increase by about 8.5%, not trivial for sure....

prashantv:
> > Since we don't modify the relay without modification path, this is expected. What is the performance impact of the appends though?
> 
> ```
> $ cat withappends.log | grep allocs
>    10214	    114906 ns/op	   8.91 MB/s	    1025 B/op	      23 allocs/op
> $ cat withoutappends.log | grep allocs
>     9610	    124755 ns/op	   8.21 MB/s	     475 B/op	      12 allocs/op
> ```
> 
> there is a rather big hit on allocs, but since we're allocating a completely new frame I think it's still somewhat reasonable. I'm slightly worried about the ns/op, which seems to increase by about 8.5%, not trivial for sure....

The frame is allocated from a pool (and released to a pool) so we actually should have 0 allocs from the new frame allocation. Do you have a sense of where the additional allocs are coming from? The performance impact doesn't seem too bad given the allocs actually (almost 2x allocs, but only a 10% performance hit), but would like to see if the alloc increase is expected, and make note of what we should optimize as follow ups.

benchmark/real_relay.go
```diff
@@ -24,6 +24,8 @@ import (
 	"errors"
 	"os"
 
+	"github.com/uber/tchannel-go/testutils"
```
prashantv, line 27: nit: import order

relay.go
```diff
@@ -466,13 +474,18 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) error {
...
 
 	f.Header.ID = destinationID
-	call.SentBytes(f.Frame.Header.FrameSize())
+
+	if len(f.arg2Appends) > 0 {
+		return r.fragmentingSend(call, f, relayToDest, origID)
+	}
+
+	call.SentBytes(uint(f.Frame.Header.FrameSize()))
+
```
prashantv, line 483: nit: delete blank line, we want the `SentBytes` close to the `Receive` ideally

relay.go
```diff
@@ -654,6 +667,63 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+func (r *Relayer) fragmentingSend(call RelayCall, f *lazyCallReq, relayToDest relayItem, origID uint32) error {
+	// TODO(echung): should we pool the writers?
+	fragWriter := newFragmentingWriter(
+		r.logger, r.newFragmentSender(relayToDest.destination, f, origID, call),
+		f.checksumType.New(),
+	)
+	defer r.conn.opts.FramePool.Release(f.Frame)
+
+	if len(f.arg2Appends) > 0 && f.isArg2Fragmented {
+		return errFragmentedArg2WithAppend
```
prashantv, line 679: `len(f.arg2Appends) > 0` is assumed if we're in this method right?

Do we have any unit tests for this case?

relay.go
```diff
@@ -654,6 +667,63 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+	}
+
+	arg2Writer, err := fragWriter.ArgWriter(false)
```
prashantv, line 682: naked bool,
https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

relay_api.go
```diff
@@ -44,7 +44,7 @@ type RelayCall interface {
 	Destination() (peer *Peer, ok bool)
 
 	// SentBytes is called when a frame is sent to the destination peer.
-	SentBytes(uint16)
+	SentBytes(uint)
```
prashantv, line 47: it's a little odd to have `uint` here, but `uint16` for received bytes. Some alternatives:
 * call `SentBytes` for each frame, so we can keep it `uint16`
 * modify both `SentBytes` and `ReceivedBytes` to `uint`

The first has no impact on Muttley. The latter will need some minor fixes in Muttley since it's technically a breaking change (but since this is an internal API, that's not a big deal)

relay_fragment_sender_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/uber/tchannel-go/testutils/thriftarg2test"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uber/tchannel-go/typed"
```
prashantv, line 11: nit: import ordering. OK either either single group for everything, or break into 2 groups (one for tchannel, one for others). right now it's a mix

relay_test.go
```diff
@@ -1403,6 +1403,72 @@ func TestRelayTransferredBytes(t *testing.T) {
...
+				ctx, cancel := NewContext(testutils.Timeout(300 * time.Millisecond))
+				defer cancel()
+
+				sendArgs := &raw.Args{
+					Arg2: thriftarg2test.BuildKVBuffer(map[string]string{"foo": "bar"}),
+					Arg3: tt.arg3,
+				}
+
+				recvArg2, recvArg3, _, err := raw.Call(ctx, client, rly.PeerInfo().HostPort, ts.ServiceName(), "echo", sendArgs.Arg2, sendArgs.Arg3)
+				if !assert.NoError(t, err, "Call from %v (%v) to %v (%v) failed", client.ServiceName(), client.PeerInfo().HostPort, ts.ServiceName(), rly.PeerInfo().HostPort) {
```
prashantv, line 1451: can we use `require.NoError` instead of return?

testutils/call.go
prashantv: True, but since it's a test-only helper, and we are often dealing with strings, I think that should be ok.

testutils/thriftarg2test/arg2_kv.go
```diff
@@ -0,0 +1,52 @@
...
+	rbuf := typed.NewReadBuffer(b)
+	nh := rbuf.ReadUint16()
+	retMap := make(map[string]string, nh)
+	for i := uint16(0); i < nh; i++ {
+		key := rbuf.ReadLen16String()
+		val := rbuf.ReadLen16String()
+		retMap[key] = val
+	}
+	if rbuf.BytesRemaining() > 0 {
+		return nil, errors.New("kv buffer wasn't fully consumed")
```
prashantv, line 43: can we use `fmt.Errorf` and include the number of bytes not consumed?

relay.go
cinchurge: the assumption was that we'd be reusing this for `Arg2Set()` in the future as well. the name of the method also doesn't explicitly say `append` in it, so I think being explicit about what this check is for is better

relay_api.go
cinchurge: went with the second approach

cinchurge:
> > > Since we don't modify the relay without modification path, this is expected. What is the performance impact of the appends though?
> > 
> > 
> > ```
> > $ cat withappends.log | grep allocs
> >    10214	    114906 ns/op	   8.91 MB/s	    1025 B/op	      23 allocs/op
> > $ cat withoutappends.log | grep allocs
> >     9610	    124755 ns/op	   8.21 MB/s	     475 B/op	      12 allocs/op
> > ```
> > 
> > 
> > there is a rather big hit on allocs, but since we're allocating a completely new frame I think it's still somewhat reasonable. I'm slightly worried about the ns/op, which seems to increase by about 8.5%, not trivial for sure....
> 
> The frame is allocated from a pool (and released to a pool) so we actually should have 0 allocs from the new frame allocation. Do you have a sense of where the additional allocs are coming from? The performance impact doesn't seem too bad given the allocs actually (almost 2x allocs, but only a 10% performance hit), but would like to see if the alloc increase is expected, and make note of what we should optimize as follow ups.

after some pprof'ing, I was able to attribute most of the allocs to two categories. First, the ones that didn't scale with the number of appends:
1. allocation of `relayFragmentingSender`: this can be eliminated with a `sync.Pool`
2. allocation of `writableFragment`: this can also be eliminated with `sync.Pool`

Then the ones that scale with appends, which are problematic:
1. The `append()` that is done on each call to `CallFrame.Arg2Append()`
2. Surprisingly, the `var sizeBuf [2]byte` were also allocations, and allocated on each call to `typed.Writer.WriteUint16()`

After address the above, and changing `CallFrame.Arg2Append(key, val []byte)` to `CallFrame.Arg2Append([]KeyVal)` (which merely assigned that slice instead of appending), I was able to reduce the new allocations down to 19 allocs/op, which also didn't scale with the number fo appends.

cinchurge:
new benchmark results:
```
    8790	    132885 ns/op	   7.71 MB/s	     762 B/op	      19 allocs/op
```

cinchurge:
need to fix a few things :(

relay.go
```diff
@@ -654,6 +666,70 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
+func writeArg2WithAppends(w io.WriteCloser, arg2 []byte, appends []relay.KeyVal) (err error) {
+	if len(arg2) < 2 {
+		return errNoNHInArg2
+	}
+
+	writer := typed.NewWriter(w)
+
+	// nh:2 is the first two bytes of arg2, which should always be present
+	nh := binary.BigEndian.Uint16(arg2[:2]) + uint16(len(appends))
+	writer.WriteUint16(nh)
```
prashantv, line 709: we're using `WriteUint16` here but manually encode to `sizeBuf` and write below? seems like we might be losing the optimization here?

if we end up converting to uint16 manually here, then the typed writer probably isn't adding much value. We may be better off writing to the `[]byte` directly

relay.go
```diff
@@ -694,3 +770,91 @@ func validateRelayMaxTimeout(d time.Duration, logger Logger) time.Duration {
...
+	if !sent {
+		rfs.failRelayItemFunc(rfs.outboundRelayItems, rfs.origID, failure)
+		return nil
+	}
+	return nil
+}
+
+func (rfs *relayFragmentSender) doneSending() {
+	rfs.call.SentBytes(rfs.sentBytes)
+	fragmentSenderPool.Put(rfs)
```
prashantv, line 859: same as above, are we 100% sure that the fragment sender is never used after?

maybe this should be a TODO with a follow up when we're definite, and we may want to put safety checks to ensure it's not reused

relay_messages.go
```diff
@@ -249,6 +250,10 @@ func (f *lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 	return arg2.NewKeyValIterator(f.Payload[f.arg2StartOffset:f.arg2EndOffset])
 }
 
+func (f *lazyCallReq) Arg2Append(keyVals []relay.KeyVal) {
+	f.arg2Appends = keyVals
```
prashantv, line 254: all we've done is shift the allocation from inside of the append method to the caller. i'm not sure if that actually helps. One advantage is that we can now pre-allocate a slice that's sized exactly for the case of > 1 allocation but that's not the common case today

Can you try the array trick? Use a backing array in the `lazyCallReq` and create a slice that's backed by the array and append to that?

This interface is also a little less clear -- since it now replaces any previously added appends, so calling it twice means the last write wins.

relay_fragment_sender_test.go
```diff
@@ -0,0 +1,272 @@
...
+				failRelayItemFunc: func(items *relayItems, id uint32, failure string) {
+					failRelayItemFuncCalled = true
+					assert.Equal(t, uint32(123), id, "got unexpected id")
+					assert.Equal(t, tt.failure, failure, "got unexpected failure string")
+				},
+				origID: 123,
+			}
+
+			err := rfs.flushFragment(wf)
+
```
prashantv, line 76: nit: delete blank

relay_fragment_sender_test.go
```diff
@@ -0,0 +1,272 @@
...
+			arg2Map: exampleArg2Map,
+			appends: []relay.KeyVal{
+				{Key: []byte("foo"), Val: []byte("bar")},
+			},
+		},
+		{
+			msg: "no nh in data",
+			writer: &dummyArgWriter{
+				writeError: []string{
+					"something went wrong", // nh
```
prashantv, line 151: suggest using `assert.AnError.String()` if you just want some error value for tests

relay_messages.go
```diff
@@ -111,7 +112,10 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 		panic(fmt.Errorf("newLazyCallReq called for wrong messageType: %v", msgType))
 	}
 
-	cr := &lazyCallReq{Frame: f}
+	cr := &lazyCallReq{
+		Frame:       f,
+		arg2Appends: make([]relay.KeyVal, 0, 2),
```
prashantv, line 117: What I was thinking was:
```
type lazyCallReq struct {
  arg2AppendInitialBuf [2]relay.KeyVal
  [..]
}

cr := &lazyCallReq {
  Frame: f,
}
cr.arg2Appends = cr.arg2AppendInitialBuf[:]
```

the current code does 2 allocations immediately, while the array approach combines the allocations

I'm curious if the array affects the performance of non-append cases. If so, we may want to skip for now (but we can do it in future if we are appending more frequently than not)

relay_messages.go
```diff
@@ -249,6 +253,16 @@ func (f *lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 	return arg2.NewKeyValIterator(f.Payload[f.arg2StartOffset:f.arg2EndOffset])
 }
 
+func (f *lazyCallReq) Arg2Append(key, val []byte) {
+	// TODO(echung): pool the appends buffers
+	if len(f.arg2Appends) == cap(f.arg2Appends) {
+		newArg2Appends := make([]relay.KeyVal, len(f.arg2Appends), 2*cap(f.arg2Appends))
+		copy(newArg2Appends, f.arg2Appends)
+		f.arg2Appends = newArg2Appends
+	}
```
prashantv, line 262: if we use the array approach, we shouldn't need any of this. all we're doing is reimplementing slice logic here, we'd be better off just pooling the slice completely if we wanted to do this

relay_test.go
```diff
@@ -1403,6 +1403,70 @@ func TestRelayTransferredBytes(t *testing.T) {
...
+		{
+			msg:           "without appends",
+			arg3:          []byte("hello, world"),
+			wantSentBytes: 127,
+		},
+		{
+			msg:           "with appends",
+			arg3:          []byte("hello, world"),
+			appends:       map[string]string{"baz": "qux"},
+			wantSentBytes: 137, // 127 + 2 bytes size + 3 bytes key + 2 byts size + 3 bytes val = 137
```
prashantv, line 1422: can you add a case where we fragment into multiple frames as well

cinchurge:
relay benchmark: (first column is number of appends)
```
  0     8800	    119952 ns/op	   8.54 MB/s	    1075 B/op	      19 allocs/op
  1     9354	    126626 ns/op	   8.09 MB/s	    1294 B/op	      20 allocs/op
  2    10168	    125308 ns/op	   8.17 MB/s	    1291 B/op	      20 allocs/op
  3     9590	    120410 ns/op	   8.50 MB/s	    1718 B/op	      21 allocs/op
  4     9495	    125961 ns/op	   8.13 MB/s	    1740 B/op	      21 allocs/op
  5     9640	    119510 ns/op	   8.57 MB/s	    1766 B/op	      21 allocs/op
  6     9614	    126881 ns/op	   8.07 MB/s	    1731 B/op	      21 allocs/op
  7     8305	    135014 ns/op	   7.58 MB/s	    2503 B/op	      22 allocs/op
  8     9657	    120451 ns/op	   8.50 MB/s	    2551 B/op	      22 allocs/op
  9     9544	    134257 ns/op	   7.63 MB/s	    2575 B/op	      22 allocs/op
 10     9520	    128338 ns/op	   7.98 MB/s	    2623 B/op	      22 allocs/op
 11     9052	    140043 ns/op	   7.31 MB/s	    2490 B/op	      22 allocs/op
 12     8894	    125978 ns/op	   8.13 MB/s	    2543 B/op	      22 allocs/op
 13     8530	    143216 ns/op	   7.15 MB/s	    2569 B/op	      22 allocs/op
 14     9075	    133662 ns/op	   7.66 MB/s	    2524 B/op	      22 allocs/op
 15     9506	    124151 ns/op	   8.25 MB/s	    4216 B/op	      23 allocs/op
 16     9531	    133116 ns/op	   7.69 MB/s	    4196 B/op	      23 allocs/op
 17     8665	    139714 ns/op	   7.33 MB/s	    4184 B/op	      23 allocs/op
 18     8167	    155985 ns/op	   6.56 MB/s	    4146 B/op	      23 allocs/op
 19     8688	    141390 ns/op	   7.24 MB/s	    4223 B/op	      23 allocs/op
 20     9003	    136509 ns/op	   7.50 MB/s	    4180 B/op	      23 allocs/op
 21     9154	    128971 ns/op	   7.94 MB/s	    4168 B/op	      23 allocs/op
 22     7486	    146769 ns/op	   6.98 MB/s	    4145 B/op	      23 allocs/op
 23     8746	    135758 ns/op	   7.54 MB/s	    4212 B/op	      23 allocs/op
 24     8827	    149359 ns/op	   6.86 MB/s	    4254 B/op	      23 allocs/op
 25     8733	    142380 ns/op	   7.19 MB/s	    4242 B/op	      23 allocs/op
 26     8654	    139343 ns/op	   7.35 MB/s	    4138 B/op	      23 allocs/op
 27     9268	    136082 ns/op	   7.52 MB/s	    4254 B/op	      23 allocs/op
 28     9260	    137466 ns/op	   7.45 MB/s	    4254 B/op	      23 allocs/op
 29     8584	    133156 ns/op	   7.69 MB/s	    4234 B/op	      23 allocs/op
 30     7950	    144989 ns/op	   7.06 MB/s	    4195 B/op	      23 allocs/op
```

relay_messages.go
```diff
@@ -94,14 +95,18 @@ func (cr lazyCallRes) OK() bool {
...
-	checksumType     ChecksumType
-	isArg2Fragmented bool
+	caller, method, delegate, key, as []byte
+	arg2Appends                       []relay.KeyVal
+	checksumType                      ChecksumType
+	isArg2Fragmented                  bool
+
+	// Intentionally an array to combine allocations with that of lazyCallReq
+	// TODO(echung): pool the appends buffers
+	arg2InitialBuf [2]relay.KeyVal
```
prashantv, line 109: let's start with `[1]` since we only plan to append a single key-value initially

relay_messages.go
```diff
@@ -94,14 +95,18 @@ func (cr lazyCallRes) OK() bool {
...
 
-	checksumType     ChecksumType
-	isArg2Fragmented bool
+	caller, method, delegate, key, as []byte
+	arg2Appends                       []relay.KeyVal
+	checksumType                      ChecksumType
+	isArg2Fragmented                  bool
+
+	// Intentionally an array to combine allocations with that of lazyCallReq
+	// TODO(echung): pool the appends buffers
```
prashantv, line 108: Just to be clear, this implies that arrays have a performance overhead we'd like to avoid in future -- is that the case?

I think the options are:

1. Don't use array backing because it's too high performance, and instead leave a TODO for pooling
2. Use an array backing, and no TODO for pooling since the array doesn't have a noticeable impact even on the "no append" case
3. Use an array backing since it improves performance, but still has a noticeable impact, so leave a TODO

The TODO implies (3), but I wanted to confirm whether the benchmark profiling shows enough of an impact on the no-append case

relay_benchmark_test.go
```diff
@@ -223,3 +225,20 @@ func BenchmarkRelay2Servers5Clients4k(b *testing.B) {
...
+
+	p := defaultParams()
+	for i := 0; i < n; i++ {
+		p.appends = append(p.appends, relay.KeyVal{Key: []byte("foo"), Val: []byte("bar")})
+	}
+	b.ResetTimer()
+	benchmarkRelay(b, p)
+}
+
+func BenchmarkRelayWithoutAppends(b *testing.B) {
```
prashantv, line 241: I initially had to check why this was different than the 0 case in `BenchmarkRelayWithAppends`, then realized the 0 case was actually 1 append

relay_benchmark_test.go
```diff
@@ -223,3 +225,20 @@ func BenchmarkRelay2Servers5Clients4k(b *testing.B) {
 	p.servers = 2
 	benchmarkRelay(b, p)
 }
+
+func BenchmarkRelayWithAppends(b *testing.B) {
+	n, err := strconv.Atoi(os.Getenv("BENCH_RELAY_APPENDS_N"))
+	require.NoError(b, err)
+
+	p := defaultParams()
+	for i := 0; i < n; i++ {
```
prashantv, line 234: I think the flexibility is probably not needed, we don't really need to test the incremental cost of each append (e.g., 6 --> 7 won't provide much signal). I think the cases we want are:
 - 0 appends (covered below with `WithoutAppends`)
 - 1 append (what we expect to use for token injection)
 - 2 appends (basically the border line for the backing array capacity)
then we probably just need some higher values like 5 and 10 to see if the cost increases linearly for each append or not
so basically just `[1, 2, 5, 10]`

prashantv (commented):
> relay benchmark: (first column is number of appends)
> 
> ```
>   0     8800	    119952 ns/op	   8.54 MB/s	    1075 B/op	      19 allocs/op
>   1     9354	    126626 ns/op	   8.09 MB/s	    1294 B/op	      20 allocs/op
> ```

I think the 0 case is actually 1 append, since no appends should be significantly faster? Can you add `b.Run` and add a message of how many appends to the output.

Change looks good, just want to make sure the benchmark numbers are as expected for no appends vs appends.

relay_benchmark_test.go
cinchurge: really good points. new results:
```
 0    9948	    114703 ns/op	   8.93 MB/s	     617 B/op	      12 allocs/op
 1    9166	    120177 ns/op	   8.52 MB/s	    1122 B/op	      19 allocs/op
 2    9906	    116765 ns/op	   8.77 MB/s	    1101 B/op	      19 allocs/op
 5   10011	    118211 ns/op	   8.66 MB/s	    1692 B/op	      21 allocs/op
10    9634	    134943 ns/op	   7.59 MB/s	    2572 B/op	      22 allocs/op
```

typed/writer.go
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
devastating, line 1: Do we need to update the time?

cinchurge:
new benchmark actually including the 0 appends case:
```
 0    9948	    114703 ns/op	   8.93 MB/s	     617 B/op	      12 allocs/op
 1    9166	    120177 ns/op	   8.52 MB/s	    1122 B/op	      19 allocs/op
 2    9906	    116765 ns/op	   8.77 MB/s	    1101 B/op	      19 allocs/op
 5   10011	    118211 ns/op	   8.66 MB/s	    1692 B/op	      21 allocs/op
10    9634	    134943 ns/op	   7.59 MB/s	    2572 B/op	      22 allocs/op
```

typed/writer.go
```diff
@@ -0,0 +1,88 @@
...
+// WriteUint16 writes a uint16 to the io.Writer
+func (w *Writer) WriteUint16(n uint16) {
+	if w.err != nil {
+		return
+	}
+
+	sizeBuf := intBufferPool.Get().(*intBuffer)
+	defer intBufferPool.Put(sizeBuf)
+
+	binary.BigEndian.PutUint16(sizeBuf[:2], n)
```
devastating, line 68: I don't have the context of why we used sync pool here, but would it be faster than a stack variable here?

like `var [2]byte`

relay.go
```diff
@@ -654,6 +669,63 @@ func (r *Relayer) handleLocalCallReq(cr *lazyCallReq) bool {
...
 
+func (r *Relayer) fragmentingSend(call RelayCall, f *lazyCallReq, relayToDest relayItem, origID uint32) error {
+	// TODO(echung): should we pool the writers?
+	fragWriter := newFragmentingWriter(
+		r.logger, r.newFragmentSender(relayToDest.destination, f, origID, call),
+		f.checksumType.New(),
+	)
+	defer r.conn.opts.FramePool.Release(f.Frame)
+
+	if len(f.arg2Appends) > 0 && f.isArg2Fragmented {
```
devastating, line 681: should this be moved before creating fragmenting writer?

devastating (approved):
I only reviewed the non-test code and left some minor comments.

typed/writer.go
cinchurge: let's do it in a separate diff

typed/writer.go
cinchurge: passing a slice pointing to a `[2]byte` into `io.Writer.Write()` causes that block of memory to escape from the function, resulting in an alloc.

I verified it via:
```golang
package tchannel

import (
	"io/ioutil"
	"sync"
	"testing"
)

var pool = sync.Pool{New: func() interface{} {
	return new([8]byte)
}}

func fooWithPool() error {
	sizeBuf := pool.Get().(*[8]byte)
	defer pool.Put(sizeBuf)

	if _, err := ioutil.Discard.Write(sizeBuf[:]); err != nil {
		return err
	}

	return nil
}

func foo() error {
	var sizeBuf [8]byte

	if _, err := ioutil.Discard.Write(sizeBuf[:]); err != nil {
		return err
	}

	return nil
}

func BenchmarkArrayWithPool(b *testing.B) {
	var err error
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if err = fooWithPool(); err != nil {
			break
		}
	}
}

func BenchmarkArray(b *testing.B) {
	var err error
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if err = foo(); err != nil {
			break
		}
	}
}
```
which gives
```
$ go test -run=NONE -bench=BenchmarkArray -benchmem
goos: darwin
goarch: amd64
pkg: github.com/uber/tchannel-go
BenchmarkArrayWithPool-8   	57540184	        21.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkArray-8           	74787500	        14.8 ns/op	       8 B/op	       1 allocs/op
PASS
ok  	github.com/uber/tchannel-go	2.747s
```

connection.go
```diff
@@ -696,14 +696,16 @@ func (c *Connection) readFrames(_ uint32) {
...
+		var err error
+		if shouldRelease, err = c.relay.Relay(frame); err != nil {
 			c.log.WithFields(
 				ErrField(err),
 				LogField{"header", frame.Header},
 				LogField{"remotePeer", c.remotePeerInfo},
 			).Error("Failed to relay frame.")
 		}
-		return false
+		return shouldRelease
```
prashantv, line 708: the nested if is useful for scoping variables to the if, but in this case, we're not scoping any variables, so it'll be cleaner to avoid it,
```
		shouldRelease, err := c.relay.Relay(frame)
		if err != nil {
			c.log.WithFields(				c.log.WithFields(
				ErrField(err),					ErrField(err),
				LogField{"header", frame.Header},					LogField{"header", frame.Header},
				LogField{"remotePeer", c.remotePeerInfo},					LogField{"remotePeer", c.remotePeerInfo},
			).Error("Failed to relay frame.")				).Error("Failed to relay frame.")
		}			}
		return false			return shouldRelease

```

relay.go
```diff
@@ -230,23 +238,23 @@ func NewRelayer(ch *Channel, conn *Connection) *Relayer {
 }
 
 // Relay is called for each frame that is read on the connection.
-func (r *Relayer) Relay(f *Frame) error {
+func (r *Relayer) Relay(f *Frame) (bool, error) {
```
prashantv, line 241: use named variables to clarify return `shouldRelease bool, _ error`

relay.go
```diff
@@ -394,9 +402,9 @@ func (r *Relayer) getDestination(f *lazyCallReq, call RelayCall) (*Connection, b
 	return remoteConn, true, nil
 }
 
-func (r *Relayer) handleCallReq(f *lazyCallReq) error {
+func (r *Relayer) handleCallReq(f *lazyCallReq) (bool, error) {
```
prashantv, line 405: named return and better documentation on the `false` values, similar to above

relay.go
```diff
@@ -230,23 +238,23 @@ func NewRelayer(ch *Channel, conn *Connection) *Relayer {
...
+func (r *Relayer) Relay(f *Frame) (bool, error) {
 	if f.messageType() != messageTypeCallReq {
 		err := r.handleNonCallReq(f)
 		if err == errUnknownID {
 			// This ID may be owned by an outgoing call, so check the outbound
 			// message exchange, and if it succeeds, then the frame has been
 			// handled successfully.
 			if err := r.conn.outbound.forwardPeerFrame(f); err == nil {
-				return nil
+				return false, nil
```
prashantv, line 249: for each `false`, add a single line of comment about why we don't need to release the frame

or alternatively, 
```
// Comment about why we don't need to release the frame in these paths
const relayNoRelease = false
```

relay.go
```diff
@@ -694,3 +767,87 @@ func validateRelayMaxTimeout(d time.Duration, logger Logger) time.Duration {
...
+	sent, failure := rfs.frameReceiver.Receive(wf.frame, requestFrame)
+	if !sent {
+		rfs.failRelayItemFunc(rfs.outboundRelayItems, rfs.origID, failure)
+		return nil
+	}
+	return nil
+}
+
+func (rfs *relayFragmentSender) doneSending() {
+	rfs.call.SentBytes(rfs.sentBytes)
```
prashantv, line 852: why are we waiting for `doneSending` to increment sent bytes? it would likely be more accurate to track the sent bytes as we sent the frame, so as part of the `frameReceiver.Receive` I think? That way, we increment the counter at the point we're sending, rather than waiting till the end, so we'll have more accurate per-second counters

that would also avoid us changing from `uint16` to `uint` I think

relay_benchmark_test.go
```diff
@@ -223,3 +223,32 @@ func BenchmarkRelay2Servers5Clients4k(b *testing.B) {
...
+	benchmarkRelayWithAppends(b, 2)
+}
+
+func BenchmarkRelayAppends5(b *testing.B) {
+	benchmarkRelayWithAppends(b, 5)
+}
+
+func BenchmarkRelayAppends10(b *testing.B) {
+	benchmarkRelayWithAppends(b, 10)
+}
```
prashantv, line 254: Use `b.Run` instead,
```
func BenchmarkRelayAppends(b *testing.B) {
  for _, appends := range []int{0, 1, 2, 5, 10} {
    b.Run(fmt.Sprintf("%v appends", appends), func(b *testing.B) {

    })
  }
}
```

no need for many functions, and can have the logic of `benchmarkRelayWithAppends` directly in the single function

relay_fragment_sender_test.go
```diff
@@ -0,0 +1,271 @@
...
+func (d *dummyFrameReceiver) Receive(f *Frame, fType frameType) (sent bool, failureReason string) {
+	return d.retSent, d.retFailureReason
+}
+
+func TestRelayFragmentSender(t *testing.T) {
+	f := NewFrame(MaxFramePayloadSize)
+
+	wbuf := typed.NewWriteBuffer(f.Payload)
+
+	wbuf.WriteBytes([]byte("foo"))
```
prashantv, line 28: why do we write something here?

relay_fragment_sender_test.go
```diff
@@ -0,0 +1,271 @@
...
+					failRelayItemFuncCalled = true
+					assert.Equal(t, uint32(123), id, "got unexpected id")
+					assert.Equal(t, tt.failure, failure, "got unexpected failure string")
+				},
+				origID: 123,
+			}
+
+			err := rfs.flushFragment(wf)
+			if tt.wantError != "" {
+				require.EqualError(t, err, tt.wantError)
```
prashantv, line 77: should we be validating something about `wbuf` here?

prashantv (approved):
Few more minor comments, but functionality LGTM

relay.go
cinchurge: 🤦 that was entirely a misunderstanding. somehow I kept thinking that `call.Sent()` sets a per-call value rather than increments a counter. will change.

relay_benchmark_test.go
```diff
@@ -223,3 +223,20 @@ func BenchmarkRelay2Servers5Clients4k(b *testing.B) {
 	p.servers = 2
 	benchmarkRelay(b, p)
 }
+
+func benchmarkRelayWithAppends(b *testing.B, n int) {
```
prashantv, line 227: only used in one place, so suggest inlining this logic into the caller

relay_fragment_sender_test.go
```diff
@@ -0,0 +1,283 @@
...
+		wantPayload                    []byte
+	}{
+		{
+			msg:         "successful send",
+			frame:       f,
+			sent:        true,
+			wantPayload: []byte("hello, world"),
+		},
+		{
+			msg:                            "send falure",
```
prashantv, line 53:
```suggestion
			msg:                            "send failure",
```

prashantv (approved):

PR #795, opened 2020-08-07
Optimize KVIterator and typed.ReadBuffer for 25% improvement

prashantv:
As part of #792, we moved to typed.ReadBuffer which is much more
readable than direct binary translations, but we lost a little
bit of performance on the KV iterator,

Original:
```
BenchmarkKeyValIterator-12      52558426                66.7 ns/op
```

With typed.ReadBuffer:
```
BenchmarkKeyValIterator-12      34918740               101 ns/op
```

WIth these optimizations, performance is close to the original:
```
BenchmarkKeyValIterator-12      46994628                76.1 ns/op
```

This change has the following optimizations:

 * Make ReadBuffer smaller by 2 words (replace []byte with int). This
   gets rid of the Fill functionality, which was never used in any
   production code anyway.
 * Don't create a read buffer for single Uint16 read when creating KeyValueIterator
 * Instead of using io.EOF to check for end of all headers, add a Remaining method
   which returns a bool (vs the Next method which returns > 10 words). We were
   spending time just copying the large struct values on the stack after the last
   header read.
 * Simplify how the KeyValueIterator.Next method creates remaining by adding method
   directly to the ReadBuffer.
 * Use named returns in KeyValueIterator.Next which give an unexpected ~6% performance
   improvement. Filed https://github.com/golang/go/issues/40638 as we'd like to avoid
   named returns here.

### DIFF
--- a/thrift/arg2/kv_iterator.go
+++ b/thrift/arg2/kv_iterator.go
@@ -5,6 +5,7 @@
 package arg2
 
 import (
+	"encoding/binary"
 	"io"
 
 	"github.com/uber/tchannel-go/typed"
@@ -15,7 +16,6 @@ import (
 // NOTE: to be optimized for performance, we try to limit the allocation
 // done in the process of iteration.
 type KeyValIterator struct {
-	arg2Len       int
 	remaining     []byte
 	leftPairCount int
 	key           []byte
@@ -30,14 +30,10 @@ func NewKeyValIterator(arg2Payload []byte) (KeyValIterator, error) {
 		return KeyValIterator{}, io.EOF
 	}
 
-	// We don't hold on to rbuf to avoid the associated alloc
-	rbuf := typed.NewReadBuffer(arg2Payload)
-	leftPairCount := rbuf.ReadUint16()
-
+	leftPairCount := binary.BigEndian.Uint16(arg2Payload[0:2])
 	return KeyValIterator{
-		arg2Len:       len(arg2Payload),
 		leftPairCount: int(leftPairCount),
-		remaining:     arg2Payload[rbuf.BytesRead():],
+		remaining:     arg2Payload[2:],
 	}.Next()
 }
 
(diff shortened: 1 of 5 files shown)


typed/buffer_test.go
```diff
@@ -81,27 +81,44 @@ func TestReadWrite(t *testing.T) {
...
-	require.NoError(t, r.Err())
+func TestReadBufferTracking(t *testing.T) {
+	bs := []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}
+
+	// Run twice, once on the original buffer, and once after wrapping.
+	rbuf := NewReadBuffer(bs)
+	for i := 0; i < 2; i++ {
+		if i == 1 {
+			rbuf.Wrap(bs)
+		}
```
cinchurge, line 94: would be easier to follow to iterate over two `*ReadBuffer`s, e.g. `for _, r := range []*Readbuffer{...}` checking indices in a loop can get a bit confusing

cinchurge:
i like those changes! it also felt a bit redundant to have both `buffer` and `remaining` when i was refactoring in the previous diff. overall looks great!

typed/buffer_test.go
prashantv: yeah, agree, this is a little ugly. However, I want to use the same read buffer wrapped (to make sure that `initialLength` is reset). I've changed the loop to instead be over whether to wrap or not, let me know what you think of the new approach.

thrift/arg2/kv_iterator.go
```diff
@@ -30,14 +30,10 @@ func NewKeyValIterator(arg2Payload []byte) (KeyValIterator, error) {
 		return KeyValIterator{}, io.EOF
 	}
 
-	// We don't hold on to rbuf to avoid the associated alloc
-	rbuf := typed.NewReadBuffer(arg2Payload)
-	leftPairCount := rbuf.ReadUint16()
-
+	leftPairCount := binary.BigEndian.Uint16(arg2Payload[0:2])
```
cinchurge, line 33: since we're not doing anything else with `leftPairCount`, maybe just inline below?

cinchurge:
lgtm :stamp: (don't have the approve option though 😓 )

thrift/arg2/kv_iterator.go
prashantv: Since we're also casting to `int`, and we do the `Next` in that line, thought separate lines would be easier so there's not too much cognitive overhead in that expression.

PR #796, opened 2020-08-07
Switch to pointer receivers for CallFrame implementations

cinchurge:
In order to support arg2 appends in the relayer, we must allow mutations to implementations of the `relay.CallFrame` interface so that we are able to allow mutations in `RelayHost.Start()`.

This change changes the method receivers of `lazyCallReq` and `FakeCallFrame` to pointers, so that we may extend the `CallFrame` interface with methods such as `Arg2Append()` later.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -398,7 +398,7 @@ func (r *Relayer) handleCallReq(f lazyCallReq) error {
 		return nil
 	}
 
-	call, err := r.relayHost.Start(f, r.relayConn)
+	call, err := r.relayHost.Start(&f, r.relayConn)
 	if err != nil {
 		// If we have a RateLimitDropError we record the statistic, but
 		// we *don't* send an error frame back to the client.
--- a/relay.go
+++ b/relay.go
@@ -355,7 +355,7 @@ func (r *Relayer) canHandleNewCall() (bool, connectionState) {
 	return canHandle, curState
 }
 
-func (r *Relayer) getDestination(f lazyCallReq, call RelayCall) (*Connection, bool, error) {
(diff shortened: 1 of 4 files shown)


relay.go
```diff
@@ -398,7 +398,7 @@ func (r *Relayer) handleCallReq(f lazyCallReq) error {
 		return nil
 	}
 
-	call, err := r.relayHost.Start(f, r.relayConn)
+	call, err := r.relayHost.Start(&f, r.relayConn)
```
prashantv, line 401: should we be converting to a pointer when the `lazyCallReq` is creatd? that way it'll be consistent (pointer everywhere) vs right now the `handleCallReq` method is using a value

cinchurge: we only need to mutate the `lazyCallReq` in `Start()`, so is it a good idea to pass pointers everywhere?

prashantv: Generally better to be consistent (e.g., even our read-only methods take a pointer, even though a value is enough).

cinchurge:
no noticeable impact:
```
   44004	     25738 ns/op	    7922 B/op	      21 allocs/op
   12904	    102857 ns/op	   9.96 MB/s	     449 B/op	      12 allocs/op
   16474	     82052 ns/op	  12.48 MB/s	     469 B/op	      12 allocs/op
    8618	    133728 ns/op	  30.63 MB/s	     461 B/op	      12 allocs/op
   36903	     34018 ns/op	    5560 B/op	      18 allocs/op
    8763	    118846 ns/op	   8.62 MB/s	     456 B/op	      12 allocs/op
   13946	     85662 ns/op	  11.95 MB/s	     472 B/op	      12 allocs/op
    7435	    141939 ns/op	  28.86 MB/s	     459 B/op	      12 allocs/op
```

PR #799, opened 2020-08-08
relay: Add test framework to validate arg2 modification

prashantv:
This adds a test that can be used for arg2 modification tests for
different cases such as:
 * Payload sizes
 * Header sizes
 * Different checksum
 * Different header modifications
 * App error bit set

Enable the "no-op" test, which ensures that without modification, the
server/client validations work as expected (the body/payloads are as
expected, the transport headers such as format match, etc).

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -1364,8 +1364,6 @@ func inspectFrames(rh *relaytest.StubRelayHost) chan relay.CallFrame {
 }
 
 func TestRelayModifyArg2(t *testing.T) {
-	// largeKey := testutils.RandString(378)
-	// largeVal := testutils.RandString(65000)
 	largePayload := testutils.RandString(1024)
 
 	modifyTests := []struct {
@@ -1381,7 +1379,8 @@ func TestRelayModifyArg2(t *testing.T) {
 				// no change
 			},
 		},
-		// TODO(echung): Enable once we have frame modification.
+		// Example test showing how modifications should be tested.
+		// TODO(echung): Enable and add more modification tests.
 		// {
 		// 	msg: "add small key/value",
 		// 	modifyFrame: func(cf relay.CallFrame, _ *relay.Conn) {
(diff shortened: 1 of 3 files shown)


relay_test.go
```diff
@@ -1361,3 +1362,207 @@ func inspectFrames(rh *relaytest.StubRelayHost) chan relay.CallFrame {
 	})
 	return frameCh
 }
+
+func TestRelayModifyArg2(t *testing.T) {
+	// largeKey := testutils.RandString(378)
```
cinchurge, line 1367: let's add a TODO here as well so it doesn't look like an accidental debug change

cinchurge: actually on second thought, can we just remove the large key/val code altogether? I already have it in my branch so I can add it back in a follow-up diff. leaving commented code around like this feels weird.

relay_test.go
```diff
@@ -1361,3 +1362,207 @@ func inspectFrames(rh *relaytest.StubRelayHost) chan relay.CallFrame {
...
+					}
+
+					ctx, cancel := NewContextBuilder(testutils.Timeout(time.Second)).
+						SetFormat(format).Build()
+					defer cancel()
+
+					// Make calls with different payloads and expected errors.
+					for _, aet := range appErrTests {
+						for _, tt := range payloadTests {
+							arg2Encoded := encodeThriftHeaders(t, tt.arg2)
```
cinchurge, line 1498: can we use subtests here as well so it's easier to run only specific test cases?

relay_test.go
prashantv: Yeah, I'll delete this, and just leave the "small modification" as an example with a comment, and delete the rest.

relay_test.go
prashantv: Sure, added messages, it looks like:
```
=== RUN   TestRelayModifyArg2/no_change/checksum=crc32/with_relay/app_error_bit,no_payload
=== RUN   TestRelayModifyArg2/no_change/checksum=crc32/with_relay/app_error_bit,small_payloads
=== RUN   TestRelayModifyArg2/no_change/checksum=crc32/with_relay/app_error_bit,large_payloads
```

PR #800, opened 2020-08-10
fragmenting relay 2/n: Update lazyCallReq to return header and args

cinchurge:
In order to mutate a frame and potentially do a fragmenting send, we'll need access to the original content of the frame, specifically the header and arguments. This change refactors `newLazyCallReq` to keep track of the offsets of its contents, and adds some methods for exposing the data.

### DIFF
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -103,7 +103,6 @@ type lazyCallReq struct {
 
 	checksumTypeOffset             uint16
 	arg2StartOffset, arg2EndOffset uint16
-	arg2appends                    []keyVal
 	arg3StartOffset, arg3EndOffset uint16
 
 	checksumType     ChecksumType
@@ -171,7 +170,6 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 	if !cr.isArg2Fragmented {
 		// arg3~2
 		arg3Len := rbuf.ReadUint16()
-		fmt.Println("arg3Len=", arg3Len)
 		cr.arg3StartOffset = rbufOffset()
 
 		if uint16(rbuf.BytesRemaining()) < arg3Len {
@@ -269,10 +267,6 @@ func (f *lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 	}
 
 	return arg2.NewKeyValIterator(f.Payload[f.arg2StartOffset:f.arg2EndOffset])
-}
-
-func (f *lazyCallReq) Arg2Append(key, val []byte) {
-	f.arg2appends = append(f.arg2appends, keyVal{key, val})
 }
 
 // finishesCall checks whether this frame is the last one we should expect for
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -98,7 +98,7 @@ type lazyCallReq struct {
 
 	checksumTypeOffset             uint16
 	arg2StartOffset, arg2EndOffset uint16
-	arg3StartOffset, arg3EndOffset uint16
+	arg3StartOffset                uint16
 
 	checksumType     ChecksumType
 	isArg2Fragmented bool
@@ -155,12 +155,8 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 	cr.arg2EndOffset = cr.arg2StartOffset + arg2Len
 
 	// arg2 is fragmented if we don't see arg3 in this frame.
-	if uint16(rbuf.BytesRemaining()) <= arg2Len {
-		rbuf.SkipBytes(rbuf.BytesRemaining())
-		cr.isArg2Fragmented = cr.HasMoreFragments()
-	} else {
-		rbuf.SkipBytes(int(arg2Len))
-	}
+	rbuf.SkipBytes(int(arg2Len))
+	cr.isArg2Fragmented = rbuf.BytesRemaining() == 0 && cr.HasMoreFragments()
 
 	if !cr.isArg2Fragmented {
 		// arg3~2
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -246,12 +246,6 @@ func (f *lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 	if !bytes.Equal(f.as, _tchanThriftValueBytes) {
 		return arg2.KeyValIterator{}, fmt.Errorf("non thrift scheme %s", f.as)
 	}
-
-	// TODO(echung): drop this since we already do error checks in the constructor
-	if f.arg2EndOffset > f.Header.PayloadSize() {
-		return arg2.KeyValIterator{}, errBadArg2Len
-	}
-
 	return arg2.NewKeyValIterator(f.Payload[f.arg2StartOffset:f.arg2EndOffset])
 }
 
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -101,11 +101,12 @@ type lazyCallReq struct {
 
 	caller, method, delegate, key, as []byte
 
-	headerStartOffset              int
-	headerEndOffset                int
-	checksumType                   ChecksumType
+	checksumType     ChecksumType
+	isArg2Fragmented bool
+
+	transportHeaderStartOffset     int
+	transportHeaderEndOffset       int
 	arg2StartOffset, arg2EndOffset int
-	isArg2Fragmented               bool
 	arg3StartOffset, arg3EndOffset int
 }
 
@@ -126,13 +127,12 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 	rbuf.ReadBytes(serviceLen)
 
 	// nh:1 (hk~1 hv~1){nh}
-	cr.headerStartOffset = rbufOffset()
+	cr.transportHeaderStartOffset = rbufOffset()
 	numHeaders := int(rbuf.ReadSingleByte())
 
 	for i := 0; i < numHeaders; i++ {
 		keyLen := int(rbuf.ReadSingleByte())
 		key := rbuf.ReadBytes(keyLen)
-
 		valLen := int(rbuf.ReadSingleByte())
 		val := rbuf.ReadBytes(valLen)
 
@@ -146,7 +146,7 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 			cr.key = val
 		}
 	}
-	cr.headerEndOffset = rbufOffset()
+	cr.transportHeaderEndOffset = rbufOffset()
 
 	// csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2
 	cr.checksumType = ChecksumType(rbuf.ReadSingleByte())
@@ -249,8 +249,8 @@ func (f *lazyCallReq) Arg2StartOffset() int {
 	return f.arg2StartOffset
 }
 
-func (f *lazyCallReq) header() []byte {
-	return f.Payload[f.headerStartOffset:f.headerEndOffset]
+func (f *lazyCallReq) transportHeader() []byte {
+	return f.Payload[f.transportHeaderStartOffset:f.transportHeaderEndOffset]
 }
 
 func (f *lazyCallReq) arg2() []byte {
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -247,6 +247,7 @@ func (f *lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 		return arg2.KeyValIterator{}, fmt.Errorf("non thrift scheme %s", f.as)
 	}
 
+	// TODO(echung): drop this since we already do error checks in the constructor
 	if f.arg2EndOffset > f.Header.PayloadSize() {
 		return arg2.KeyValIterator{}, errBadArg2Len
 	}
--- a/relay_messages.go
+++ b/relay_messages.go
@@ -113,12 +113,12 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 
 	cr := &lazyCallReq{Frame: f}
 
-	rbuf := typed.NewReadBuffer(f.SizedPayload()[_serviceLenIndex:])
-	rbufOffset := func() uint16 { return _serviceLenIndex + uint16(rbuf.BytesRead()) }
+	rbuf := typed.NewReadBuffer(f.SizedPayload())
+	rbuf.SkipBytes(_serviceLenIndex)
 
 	// service~1
 	serviceLen := rbuf.ReadSingleByte()
-	rbuf.ReadBytes(int(serviceLen))
+	rbuf.SkipBytes(int(serviceLen))
 
 	// nh:1 (hk~1 hv~1){nh}
 	numHeaders := rbuf.ReadSingleByte()
@@ -141,9 +141,9 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 	}
 
 	// csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2
-	cr.checksumTypeOffset = rbufOffset()
+	cr.checksumTypeOffset = uint16(rbuf.BytesRead())
 	cr.checksumType = ChecksumType(rbuf.ReadSingleByte())
-	rbuf.ReadBytes(cr.checksumType.ChecksumSize())
+	rbuf.SkipBytes(cr.checksumType.ChecksumSize())
 
 	// arg1~2
 	arg1Len := int(rbuf.ReadUint16())
@@ -151,21 +151,21 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 
 	// arg2~2
 	arg2Len := rbuf.ReadUint16()
-	cr.arg2StartOffset = rbufOffset()
+	cr.arg2StartOffset = uint16(rbuf.BytesRead())
 	cr.arg2EndOffset = cr.arg2StartOffset + arg2Len
 
 	// arg2 is fragmented if we don't see arg3 in this frame.
 	if uint16(rbuf.BytesRemaining()) <= arg2Len {
-		rbuf.ReadBytes(rbuf.BytesRemaining())
+		rbuf.SkipBytes(rbuf.BytesRemaining())
 		cr.isArg2Fragmented = cr.HasMoreFragments()
 	} else {
-		rbuf.ReadBytes(int(arg2Len))
+		rbuf.SkipBytes(int(arg2Len))
 	}
 
 	if !cr.isArg2Fragmented {
 		// arg3~2
 		arg3Len := rbuf.ReadUint16()
-		cr.arg3StartOffset = rbufOffset()
+		cr.arg3StartOffset = uint16(rbuf.BytesRead())
 
 		if uint16(rbuf.BytesRemaining()) < arg3Len {
 			cr.arg3EndOffset = cr.arg3StartOffset + uint16(rbuf.BytesRemaining())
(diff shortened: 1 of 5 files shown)


relay_messages.go
```diff
@@ -132,24 +146,40 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
...
 
 	// arg1~2
 	arg1Len := int(rbuf.ReadUint16())
 	cr.method = rbuf.ReadBytes(arg1Len)
 
 	// arg2~2
 	arg2Len := int(rbuf.ReadUint16())
-	cr.arg2StartOffset = int(headerStartIndex) + rbuf.BytesRead()
-	// TODO(echung): deprecate the arg2 offsets and store just the arg2 slice instead
-	// Read past arg2
```
cinchurge, line 148: eventually decided against this as the effort doesn't quite justify the return - a lot of tests depend on knowing the arg2 offsets, and switching to a slice may actually slightly hurt performance, so giving up

relay_messages.go
```diff
@@ -91,13 +91,22 @@ func (cr lazyCallRes) OK() bool {
...
+	val []byte
+}
+
 type lazyCallReq struct {
 	*Frame
 
 	caller, method, delegate, key, as []byte
 
+	headerStartOffset              int
+	headerEndOffset                int
```
prashantv, line 105: looking at the code, this refers to the transport headers. If we need to keep these, let's add that to the name, so it's clear this isn't arg2 headers.

It looks like we use this when creating a new frame? Not 100% clear why these are needed, when we need to create a new frame, shouldn't we just copy the frame up to `arg2`?
 *

relay_messages.go
```diff
@@ -91,13 +91,22 @@ func (cr lazyCallRes) OK() bool {
...
+	checksumType                   ChecksumType
```
prashantv, line 106: nit: play with the size of this struct and check whether this is the right place to store this value. I'd guess that putting this next to `isArg2Fragmented` will make the struct smaller (ints are aligned to 8-byte boundaries in 64-bit, so putting a 1 byte "wastes" 7 bytes after it, vs packing the small 1 byte values together)

relay_messages_test.go
```diff
@@ -307,6 +310,68 @@ func TestLazyCallReqSetTTL(t *testing.T) {
 	})
 }
 
+func byteKeyValToMap(tb testing.TB, buffer []byte) map[string]string {
```
prashantv, line 313: we already have code very similar to this:
https://sourcegraph.uberinternal.com/github.com/uber/tchannel-go/-/blob/messages.go#L181

relay_messages_test.go
```diff
@@ -307,6 +310,68 @@ func TestLazyCallReqSetTTL(t *testing.T) {
...
+		key := rbuf.ReadBytes(keyLen)
+		valLen := int(rbuf.ReadSingleByte())
+		val := rbuf.ReadBytes(valLen)
+		retMap[string(key)] = string(val)
+	}
+	require.NoError(tb, rbuf.Err())
+	return retMap
+}
+
+func uint16KeyValToMap(tb testing.TB, buffer []byte) map[string]string {
```
prashantv, line 328: I think this is the same as the `decodeThriftHeaders` function that's in `relay_test.go`? We can move it to a test package to make it easier, but shouldn't need to reimplement it

relay_messages.go
cinchurge: the checksum changes when arg2 is modified (which comes right after the headers) so we can't just blindly copy over everything, but will have to
- write header
- defer checksum
- write arg1 with checksum.Add(arg1)
- write arg2 with checksum.Add(arg2)
- write arg3 with checksum.Add(arg3)
- checksumRef.Update(checksum.Sum())

will rename to transportHeaders

prashantv: Why not:
 * Copy frame up to arg2
 * Store offset of where checksum is located
 * checksum.Add(arg1)
 * Write arg2 and arg3, while also updating checksum
 * Modify checksum using the offset of where checksum is located

relay_messages.go
cinchurge: good call! placed `IsArg2Fragmented` and `ChecksumType` together reduced the struct by 8 bytes.

relay_messages_test.go
cinchurge: turns out there's any import cycle preventing that from being reused: `tchannel` imports `thrift` (`thrift.ReadHeaders()`), which imports `tchannel`. added TODO to change this once we have bandwidth to do a large-ish refactor to resolve this.

cinchurge:
benchmarks show no significant impact:
```
   16287	     65357 ns/op	  15.67 MB/s	     443 B/op	      12 allocs/op
   24693	     49499 ns/op	  20.69 MB/s	     478 B/op	      12 allocs/op
   13987	     75435 ns/op	  54.30 MB/s	     469 B/op	      12 allocs/op
   17516	     67423 ns/op	  15.19 MB/s	     415 B/op	      12 allocs/op
   23822	     60139 ns/op	  17.03 MB/s	     471 B/op	      12 allocs/op
   12994	     93586 ns/op	  43.77 MB/s	     454 B/op	      12 allocs/op
```

cinchurge:
before change: sizeof(lazyCallReq)=152
after change: sizeof(lazyCallReq)=144

relay_messages.go
```diff
@@ -109,16 +113,19 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
 
 	cr := &lazyCallReq{Frame: f}
 
-	serviceLen := f.Payload[_serviceLenIndex]
-	// nh:1 (hk~1 hv~1){nh}
-	headerStartIndex := _serviceLenIndex + 1 /* length byte */ + serviceLen
+	rbuf := typed.NewReadBuffer(f.SizedPayload()[_serviceLenIndex:])
+	rbufOffset := func() uint16 { return _serviceLenIndex + uint16(rbuf.BytesRead()) }
```
prashantv, line 117: do you want to call this `payloadOffset` instead?

relay_messages.go
```diff
@@ -109,16 +113,19 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
...
 
-	rbuf := typed.NewReadBuffer(f.SizedPayload()[headerStartIndex:])
+	// service~1
+	serviceLen := rbuf.ReadSingleByte()
+	rbuf.ReadBytes(int(serviceLen))
```
prashantv, line 121: We're doing a few things which we could cleanup:
 * we offset into the payload to create the rbuf
 * we then add `_serviceLenIndex` (the offset) to `BytesRead` to create the final payload aspect
 * we're calling `ReadBytes` and ignoring the result

I think these all point to a missing method in the read buffer -- a `Skip(n int)` which will skip ahead in the buffer by n bytes. That could be used to skip to the service length, and also skip the service name itself, and avoid the `rbufOffset` function completely.

relay_messages.go
```diff
@@ -134,22 +141,38 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
...
-	// TODO(echung): deprecate the arg2 offsets and store just the arg2 slice instead
-	// Read past arg2
-	rbuf.ReadBytes(arg2Len)
+	arg2Len := rbuf.ReadUint16()
+	cr.arg2StartOffset = rbufOffset()
 	cr.arg2EndOffset = cr.arg2StartOffset + arg2Len
+
 	// arg2 is fragmented if we don't see arg3 in this frame.
-	cr.isArg2Fragmented = rbuf.BytesRemaining() == 0 && cr.HasMoreFragments()
+	if uint16(rbuf.BytesRemaining()) <= arg2Len {
```
prashantv, line 158: if the frame says it contains X bytes for arg2, then it must contain all X bytes, otherwise it's a protocol error. So effectively: `BytesRemaining` is always >= `arg2Len`. Can we simplify this to:
```go
rbuf.Skip(arg2Len)
cr.isArg2Fragmented = rbuf.BytesRemaining() == 0 && cr.HasMoreFragments()
```

relay_messages.go
```diff
@@ -134,22 +141,38 @@ func newLazyCallReq(f *Frame) (*lazyCallReq, error) {
...
+	} else {
+		rbuf.ReadBytes(int(arg2Len))
+	}
+
+	if !cr.isArg2Fragmented {
+		// arg3~2
+		arg3Len := rbuf.ReadUint16()
+		cr.arg3StartOffset = rbufOffset()
+
+		if uint16(rbuf.BytesRemaining()) < arg3Len {
```
prashantv, line 170: as above, don't think this if is necessary. in this case, `BytesRemaining == arg3Len`, otherwise we have a protocol error.

given that, I don't think we need `arg3EndOffset`. It always ends at the frame boundary since there's nothing else after it in the protocol

I think we should do something like:
```go
if arg3Len != rbuf.BytesRemaining() {
  // return error about the mismatched arg3
}

```

relay_messages.go
```diff
@@ -96,8 +96,12 @@ type lazyCallReq struct {
 
 	caller, method, delegate, key, as []byte
 
-	arg2StartOffset, arg2EndOffset int
-	isArg2Fragmented               bool
+	checksumTypeOffset             uint16
```
prashantv, line 99: good idea using `uint16` for the offsets, saves some space!

you might want to play with struct alignment too (although I think what you have looks good already)

prashantv:
> benchmarks show no significant impact:
> 
> ```
> $ cat baseline.log | grep allocs/op
>    16287	     65357 ns/op	  15.67 MB/s	     443 B/op	      12 allocs/op
>    24693	     49499 ns/op	  20.69 MB/s	     478 B/op	      12 allocs/op
>    13987	     75435 ns/op	  54.30 MB/s	     469 B/op	      12 allocs/op
> $ cat benchmark.log | grep allocs/op
>    17516	     67423 ns/op	  15.19 MB/s	     415 B/op	      12 allocs/op
>    23822	     60139 ns/op	  17.03 MB/s	     471 B/op	      12 allocs/op
>    12994	     93586 ns/op	  43.77 MB/s	     454 B/op	      12 allocs/op
> ```

Hmm, every one of those benchmarks has lower (sometimes by a lot) MB/s -- can you run with a longer benchtime `-benchtime 5s` and run it a few times to see if it's consistent. Allocs are good, but seems like CPU is impacted more than I expected from this change.

cinchurge:
> Hmm, every one of those benchmarks has lower (sometimes by a lot) MB/s -- can you run with a longer benchtime `-benchtime 5s` and run it a few times to see if it's consistent. Allocs are good, but seems like CPU is impacted more than I expected from this change.

```
   78856	     72196 ns/op	  14.18 MB/s	     407 B/op	      12 allocs/op
  116932	     53841 ns/op	  19.02 MB/s	     423 B/op	      12 allocs/op
   72776	     87787 ns/op	  46.66 MB/s	     405 B/op	      12 allocs/op

   79749	     76678 ns/op	  13.35 MB/s	     407 B/op	      12 allocs/op
  103214	     59519 ns/op	  17.20 MB/s	     426 B/op	      12 allocs/op
   64903	     89494 ns/op	  45.77 MB/s	     424 B/op	      12 allocs/op

   78600	     68604 ns/op	  14.93 MB/s	     406 B/op	      12 allocs/op
  118072	     50952 ns/op	  20.10 MB/s	     422 B/op	      12 allocs/op
   70366	     84699 ns/op	  48.36 MB/s	     409 B/op	      12 allocs/op

   86704	     68498 ns/op	  14.95 MB/s	     401 B/op	      12 allocs/op
  110632	     56660 ns/op	  18.07 MB/s	     425 B/op	      12 allocs/op
   72034	     81536 ns/op	  50.24 MB/s	     413 B/op	      12 allocs/op
```
yea it does seem that there is a consistent drop in MB/s, though not by a whole lot? should we be concerned?

typed/buffer.go
```diff
@@ -92,6 +92,20 @@ func (r *ReadBuffer) ReadBytes(n int) []byte {
...
+		return
+	}
+
+	if len(r.remaining) < n {
+		r.err = ErrEOF
+		return
+	}
+
+	r.remaining = r.remaining[n:]
+	return
```
prashantv, line 106: no need for return here

relay_messages.go
```diff
@@ -226,7 +247,7 @@ func (f *lazyCallReq) Arg2Iterator() (arg2.KeyValIterator, error) {
 		return arg2.KeyValIterator{}, fmt.Errorf("non thrift scheme %s", f.as)
 	}
 
-	if f.arg2EndOffset > int(f.Header.PayloadSize()) {
+	if f.arg2EndOffset > f.Header.PayloadSize() {
```
prashantv, line 250: might be worth adding a TODO here, but I don't think this can happen anymore since we now have the rbuf + check for error right? we might be able to remove this error check.

relay_messages_test.go
```diff
@@ -524,3 +550,41 @@ func TestLazyErrorCodes(t *testing.T) {
...
+	cr := reqHasAll.reqWithParams(t, testCallReqParams{
+		arg2Buf: thriftarg2test.BuildKVBuffer(map[string]string{
+			"foo": "bar",
+			"baz": "qux",
+		}),
+		arg3Buf: []byte("some arg3 data"),
+	})
+
+	t.Run("checksum", func(t *testing.T) {
+		assert.Equal(t, ChecksumType(0x3), cr.checksumType)
```
prashantv, line 580: can we use the constants that describe the different checksums rather than 0x3 here?

relay_messages_test.go
```diff
@@ -524,3 +550,41 @@ func TestLazyErrorCodes(t *testing.T) {
...
+		assert.Equal(t, uint16(0x80), cr.checksumTypeOffset)
```
prashantv, line 581: can we add a comment on how this offset is 0x80?

relay_messages_test.go
```diff
@@ -524,3 +550,41 @@ func TestLazyErrorCodes(t *testing.T) {
...
+	})
+
+	t.Run(".arg2()", func(t *testing.T) {
+		assert.Equal(t, map[string]string{
```
prashantv, line 585: nit: create a variable called headers with this map, and use it when we create / here?

typed/buffer.go
```diff
@@ -92,6 +92,20 @@ func (r *ReadBuffer) ReadBytes(n int) []byte {
 	return b
 }
 
+func (r *ReadBuffer) SkipBytes(n int) {
```
prashantv, line 95: add unit test coverage to the typed package for this method

cinchurge:
the degradation appears to be related to the use of `byte` as an iteration variable, which was not necessary anyway. the benchmark results now appear to be comparable:
```
baseline:
   13946	     77751 ns/op	  52.68 MB/s	     469 B/op	      12 allocs/op
   13573	     85754 ns/op	  47.76 MB/s	     472 B/op	      12 allocs/op
   13183	     88546 ns/op	  46.26 MB/s	     459 B/op	      12 allocs/op

this change:
   14059	     76475 ns/op	  53.56 MB/s	     461 B/op	      12 allocs/op
   13468	     83500 ns/op	  49.05 MB/s	     446 B/op	      12 allocs/op
   13504	     86154 ns/op	  47.54 MB/s	     436 B/op	      12 allocs/op
```

prashantv:
I think we should be running the benchmark in relay_messages_benchmark_test.go to benchmark the impact of this change.

For comparison,
master:
```
BenchmarkCallReqFrame   35954797               165 ns/op               0 B/op          0 allocs/op
```

dev:
```
BenchmarkCallReqFrame    3472144              1666 ns/op             592 B/op          5 allocs/op
```

This is just a regression caused by use of the `require` package in the benchmark, which allocates like crazy. Removing that,
```
BenchmarkCallReqFrame   20682429               251 ns/op             160 B/op          1 allocs/op
```

The allocation is from when we switched over to using pointers (expected), which likely is causing an increase in the overall time spent.

With this change, there's another slight increase:
```
BenchmarkCallReqFrame   21282790               267 ns/op             144 B/op          1 allocs/op
```

So overall, 165 -> 250 -> 267, which seems reasonable since we're doing more work.

Can you update the benchmark to get rid of require:
```
-               require.NoError(b, err)
+               if err != nil {
+                       b.Fatal("unexpected error", err)
+               }
```

And then post your pre-post benchmark numbers.

cinchurge:
Benchmark results for newLazyCallReq:
dev:
```
BenchmarkCallReqFrame-8   	 5394974	       220 ns/op	     160 B/op	       1 allocs/op
```
this change:
```
BenchmarkCallReqFrame-8   	 5082253	       238 ns/op	     176 B/op	       1 allocs/op
```

prashantv:
Can you fix the lint issue causing one of the builds to fail:

> typed/buffer.go:95:1: exported method ReadBuffer.SkipBytes should have comment or be unexported

Let's also try and add coverage for the new code (ideally the patch coverage should be close to 100%)

cinchurge:
> the degradation appears to be related to the use of `byte` as an iteration variable, which was not necessary anyway. the benchmark results now appear to be comparable:
> 
> ```
> baseline:
>    13946	     77751 ns/op	  52.68 MB/s	     469 B/op	      12 allocs/op
>    13573	     85754 ns/op	  47.76 MB/s	     472 B/op	      12 allocs/op
>    13183	     88546 ns/op	  46.26 MB/s	     459 B/op	      12 allocs/op
> 
> this change:
>    14059	     76475 ns/op	  53.56 MB/s	     461 B/op	      12 allocs/op
>    13468	     83500 ns/op	  49.05 MB/s	     446 B/op	      12 allocs/op
>    13504	     86154 ns/op	  47.54 MB/s	     436 B/op	      12 allocs/op
> ```

turns out this is probably just a red herring - redoing the benchmark now doesn't really show systematic difference:
using `int` to iterate:
```
   13083	     84326 ns/op	  12.14 MB/s	     430 B/op	      12 allocs/op
   17619	     64461 ns/op	  15.89 MB/s	     475 B/op	      12 allocs/op
   12234	    105773 ns/op	  38.72 MB/s	     451 B/op	      12 allocs/op
```
using `byte` to iterate:
```
   11590	    102040 ns/op	  10.04 MB/s	     435 B/op	      12 allocs/op
   17680	     68194 ns/op	  15.02 MB/s	     491 B/op	      12 allocs/op
   10999	    106316 ns/op	  38.53 MB/s	     480 B/op	      12 allocs/op
```

cinchurge:
> Can you fix the lint issue causing one of the builds to fail:
> 
> > typed/buffer.go:95:1: exported method ReadBuffer.SkipBytes should have comment or be unexported
> 
> Let's also try and add coverage for the new code (ideally the patch coverage should be close to 100%)

yea, working on coverage right now

relay_messages_test.go
```diff
@@ -524,3 +550,47 @@ func TestLazyErrorCodes(t *testing.T) {
...
+	return retMap
+}
+
+func TestLazyCallReqContents(t *testing.T) {
+	cr := reqHasAll.reqWithParams(t, testCallReqParams{
+		arg2Buf: thriftarg2test.BuildKVBuffer(map[string]string{
+			"foo": "bar",
+			"baz": "qux",
+		}),
+		arg3Buf: []byte("some arg3 data"),
```
prashantv, line 576: can we move the "some arg3 data" to a constant and use it here and below when validating the value

can do the same for the map of headers too

relay_messages.go
```diff
@@ -210,13 +223,21 @@ func (f *lazyCallReq) HasMoreFragments() bool {
...
-	return f.arg2StartOffset
+	return int(f.arg2StartOffset)
+}
+
+func (f *lazyCallReq) arg2() []byte {
+	return f.Payload[f.arg2StartOffset:f.arg2EndOffset]
+}
+
+func (f *lazyCallReq) arg3() []byte {
+	return f.Payload[f.arg3StartOffset:f.Header.PayloadSize()]
```
prashantv, line 240: optionally, can simplify this: `f.SizedPayload()[f.arg3StartOffset:]`

PR #801, opened 2020-08-12
add "ConnContext" semantics to tchannel inbound, outbound, and relays

cinchurge:


prashantv:
Rather than a very relay-specific feature here, I think this could be a higher-level feature in TChannel:
 * Every connection should have a context (in `connection.go`
 * The function should match the HTTP signature, `ConnContext func(ctx context.Context, c net.Conn) context.Context`
 * The relay.Conn just copies the context from the connection.

preinit_connection.go
```diff
@@ -93,7 +98,12 @@ func (ch *Channel) inboundHandshake(ctx context.Context, c net.Conn, events conn
 		return nil, err
 	}
 
-	return ch.newConnection(c, 0 /* initialID */, "" /* outboundHP */, remotePeer, remotePeerAddress, events), nil
+	baseCtx := context.Background()
+	if p := getTChannelParams(ctx); p != nil {
```
prashantv, line 102: the only caller of this function is in `channel.go` which passes `context.Background()`, so this can never happen. should we delete this (to make it more explicit, we can remove the context argument from this function and just pass `context.Background()` below directly.

relay_test.go
```diff
@@ -516,6 +516,26 @@ func TestRelayMakeOutgoingCall(t *testing.T) {
 	})
 }
 
+func TestConnContext(t *testing.T) {
```
prashantv, line 519: can we also add a test for the outbound connect case, where we create a connection from the relay to some other channel, and when that other channel sends a call, it'll use that relay-created connection and should have the context value set on the outbound side

preinit_connection.go
```diff
@@ -98,12 +98,7 @@ func (ch *Channel) inboundHandshake(ctx context.Context, c net.Conn, events conn
 		return nil, err
 	}
 
-	baseCtx := context.Background()
-	if p := getTChannelParams(ctx); p != nil {
-		baseCtx = p.connectBaseContext
-	}
-
-	return ch.newConnection(baseCtx, c, 0 /* initialID */, "" /* outboundHP */, remotePeer, remotePeerAddress, events), nil
+	return ch.newConnection(context.Background(), c, 0 /* initialID */, "" /* outboundHP */, remotePeer, remotePeerAddress, events), nil
```
prashantv, line 101: Use `ctx`, or remove `ctx` argument from `inboundHandshake`

relay_test.go
```diff
@@ -536,6 +536,34 @@ func TestConnContext(t *testing.T) {
...
+		svr2 := ts.NewServer(svr2Opts)
+		testutils.RegisterEcho(svr2, nil)
+
+		svr3Opts := testutils.NewOpts().SetServiceName("svc3")
+		svr3 := ts.NewServer(svr3Opts)
+		testutils.RegisterEcho(svr3, nil)
+
+		baseCtx := context.WithValue(context.Background(), "foo", "bar")
+		ctx, cancel := NewContextBuilder(time.Second).SetConnectBaseContext(baseCtx).Build()
+		_, err := svr1.Connect(ctx, svr3.PeerInfo().HostPort)
```
prashantv, line 558: can we use `ts.Relay()` or rename `svr1` to `relay` to make this test easier to follow

relay_test.go
```diff
@@ -536,6 +536,34 @@ func TestConnContext(t *testing.T) {
...
+		svr2Opts := testutils.NewOpts().SetServiceName("svc2")
+		svr2 := ts.NewServer(svr2Opts)
+		testutils.RegisterEcho(svr2, nil)
+
+		svr3Opts := testutils.NewOpts().SetServiceName("svc3")
+		svr3 := ts.NewServer(svr3Opts)
+		testutils.RegisterEcho(svr3, nil)
+
+		baseCtx := context.WithValue(context.Background(), "foo", "bar")
+		ctx, cancel := NewContextBuilder(time.Second).SetConnectBaseContext(baseCtx).Build()
```
prashantv, line 557: why not use `defer cancel()` -- there's no need to cancel immediately i think?

relay_test.go
```diff
@@ -516,6 +516,54 @@ func TestRelayMakeOutgoingCall(t *testing.T) {
 	})
 }
 
+func TestConnContextInbound(t *testing.T) {
```
prashantv, line 519: make sure prefix for this function and below is `TestRelay`

testutils/channel.go
```diff
@@ -54,6 +54,11 @@ func NewServerChannel(opts *ChannelOpts) (*tchannel.Channel, error) {
 		return nil, fmt.Errorf("NewChannel failed: %v", err)
 	}
 
+	connOpts := ch.ConnectionOptions()
+	if opts.ConnContext != nil {
+		connOpts.ConnContext = opts.ConnContext
+	}
```
prashantv, line 60: not sure why this logic is here -- is it needed, and if so, can we put it in one of the existing functions

connection_test.go
```diff
@@ -1422,3 +1422,48 @@ func TestCustomDialer(t *testing.T) {
...
+			// Verify that the context passed into the handler inherits from the base context
+			// set by ConnContext
+			assert.Equal(t, "bar", ctx.Value("foo"), "Value unexpectedly different from base context")
+			return &raw.Res{Arg2: args.Arg2, Arg3: args.Arg3}, nil
+		})
+
+		copts := testutils.NewOpts()
+		bob := ts.NewClient(copts)
+		err := testutils.CallEcho(bob, ts.HostPort(), ts.ServiceName(), nil)
+		assert.NoError(t, err, "Echo failed")
```
prashantv, line 1443:
```suggestion
		testutils.AssertEcho(t, bob, ts.HostPort(), ts.ServiceName())
```

same everywhere you do call + check for no error.

connection_test.go
```diff
@@ -1422,3 +1422,48 @@ func TestCustomDialer(t *testing.T) {
...
+
+		bobOpts := testutils.NewOpts().SetServiceName("bob")
+		bob := ts.NewServer(bobOpts)
+		testutils.RegisterEcho(bob, nil)
+
+		baseCtx := context.WithValue(context.Background(), "foo", "bar")
+		ctx, cancel := NewContextBuilder(time.Second).SetConnectBaseContext(baseCtx).Build()
+		defer cancel()
+		_, err := alice.Connect(ctx, bob.PeerInfo().HostPort)
+		require.NoError(t, err)
```
prashantv, line 1464: can you run this test 1000 times, it might be flaky, as there's a chance that bob has received the connection and responded, but hasn't yet added the connection to the internal list of active connections before the call below is made.

might be better to use `Ping` or `AssertEcho` which will trigger the connection, but should also wait for the connection to be registered. it'll also give us validation that the context passed through a method call or ping to connect will not lose the value.

connection_test.go
```diff
@@ -1422,3 +1422,48 @@ func TestCustomDialer(t *testing.T) {
...
+
+		err = testutils.CallEcho(bob, ts.HostPort(), ts.ServiceName(), nil)
+		assert.NoError(t, err, "Echo failed")
```
prashantv, line 1467: can use `AssertEcho` here

relay_test.go
```diff
@@ -516,6 +516,62 @@ func TestRelayMakeOutgoingCall(t *testing.T) {
...
+		svrOpts := testutils.NewOpts().SetServiceName("svc").SetConnContext(func(c context.Context, conn net.Conn) context.Context {
+			return context.WithValue(c, "foo", "bar")
+		})
+		svr := ts.NewServer(svrOpts)
+		testutils.RegisterFunc(svr, "echo", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
+			// Verify that the context passed into the handler inherits from the base context
+			// set by ConnContext
+			assert.Equal(t, "bar", ctx.Value("foo"), "Value unexpectedly different from base context")
+			return &raw.Res{Arg2: args.Arg2, Arg3: args.Arg3}, nil
+		})
```
prashantv, line 540: since we're testing the server side separately, can we replace this with just a normal register echo / call?

relay_test.go
```diff
@@ -516,6 +516,62 @@ func TestRelayMakeOutgoingCall(t *testing.T) {
...
+	rh.SetFrameFn(func(f relay.CallFrame, conn *relay.Conn) {
+		assert.Equal(t, "bar", conn.Context.Value("foo"), "Unexpected value set in base context")
+	})
+	opts := testutils.NewOpts().SetRelayOnly().SetRelayHost(rh)
+
+	testutils.WithTestServer(t, opts, func(t testing.TB, ts *testutils.TestServer) {
+		rly := ts.Relay()
+		calleeOpts := testutils.NewOpts().SetServiceName("svc2")
+		callee := ts.NewServer(calleeOpts)
+		testutils.RegisterEcho(callee, nil)
```
prashantv, line 558: i think we can do `callee := ts.Server()` since a server is created by default when using `WithTestServer`. it'll have a default service name, but you can access that using `ts.Server().ServiceName()` or maybe also `ts.ServiceName()`

PR #802, opened 2020-08-17
Use testutils.RegisterEcho in TestRelayInboundConnContext

cinchurge:


relay_test.go
```diff
@@ -532,12 +532,7 @@ func TestRelayInboundConnContext(t *testing.T) {
 			return context.WithValue(c, "foo", "bar")
 		})
 		svr := ts.NewServer(svrOpts)
-		testutils.RegisterFunc(svr, "echo", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
-			// Verify that the context passed into the handler inherits from the base context
-			// set by ConnContext
-			assert.Equal(t, "bar", ctx.Value("foo"), "Value unexpectedly different from base context")
-			return &raw.Res{Arg2: args.Arg2, Arg3: args.Arg3}, nil
-		})
+		testutils.RegisterEcho(svr, nil)
```
prashantv, line 535: we can also get rid of the usage of the `ConnContext` function when creating the server. We should actually take advantage of `ts.Server()`

relay_test.go
```diff
@@ -519,34 +519,27 @@ func TestRelayMakeOutgoingCall(t *testing.T) {
...
-			// set by ConnContext
-			assert.Equal(t, "bar", ctx.Value("foo"), "Value unexpectedly different from base context")
-			return &raw.Res{Arg2: args.Arg2, Arg3: args.Arg3}, nil
-		})
+		svr := ts.Server()
+		testutils.RegisterEcho(svr, nil)
 
-		err := testutils.CallEcho(rly, ts.HostPort(), "svc", nil)
-		assert.NoError(t, err, "Echo failed")
+		client := testutils.NewClient(t, testutils.NewOpts())
```
prashantv, line 534: nit: we tend to just pass `nil` when there's no options

PR #806, opened 2020-09-24
Fix YAML marshaling for ConnectionOptions

cinchurge:
The addition of `ConnContext` to `ConnectionOptions` resulted in some consumers panicking due to the error
```
panic: cannot marshal type: func(context.Context, net.Conn) context.Context [recovered]
	panic: cannot marshal type: func(context.Context, net.Conn) context.Context [recovered]
	panic: cannot marshal type: func(context.Context, net.Conn) context.Context
```
since `ConnectionOptions` is sometimes embedded in config structs that are serialized or deserialized.

This change moves `ConnContext` to ChannelOptions, which already contains function fields (dial). Since there are no consumers of this feature yet, I don't expect this change to break anything.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -159,10 +159,10 @@ type ConnectionOptions struct {
 	// calls before shutting down. Only used if it is non-zero.
 	MaxCloseTime time.Duration
 
-	// connContext runs when a connection is established, which updates
+	// ConnContext runs when a connection is established, which updates
 	// the per-connection base context. This context is used as the parent context
 	// for incoming calls.
-	connContext func(ctx context.Context, conn net.Conn) context.Context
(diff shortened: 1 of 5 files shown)


connection.go
```diff
@@ -278,11 +273,7 @@ func (co ConnectionOptions) withDefaults() ConnectionOptions {
 		co.SendBufferSize = DefaultConnectionBufferSize
 	}
 	co.HealthChecks = co.HealthChecks.withDefaults()
-	if co.ConnContext == nil {
-		co.ConnContext = func(ctx context.Context, conn net.Conn) context.Context {
-			return ctx
-		}
-	}
+
```
prashantv, line 276: nit: delete blank line

prashantv (approved):
Can you update the summary as well?

PR #807, opened 2020-09-24
Update CHANGELOG and version in prep for 1.20.1 release

cinchurge:
### Fixed
 * Set ConnContext in the channel instead of the connection to avoid serialization
   errors since ConnectionOptions is sometimes embedded in configurations (#806)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ Changelog
 
 ## [1.20.1] - 2020-09-24
 ### Fixed
- * Set ConnContext in the channel instead of the connection (#806)
+ * Set ConnContext in the channel instead of the connection to avoid serialization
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+## [1.20.1] - 2020-09-24
+### Fixed
+ * Set ConnContext in the channel instead of the connection (#806)
```
prashantv, line 6: add some context on why, e.g., "to fix marshalling of connection options, often used in configurations" or something

prashantv (approved):
Technically, this is an API change, but since 1.20.0 was already incompatible with many services, this is fixing the previous breaking change, so seems reasonable.

PR #812, opened 2020-12-09
Update checksum for all frames when calls are mutated by the relayer

cinchurge:
Our previous effort to add arg2 mutation left out an important detail: in TChannel, the checksum for every frame is seeded by the checksum of its previous frame. Meaning that once the `callReq` frame has been mutated, all subsequent `callReqContinue` frames must also have their checksum updated.

This change extends `relayItem` with a reference to the "call state", which tracks
- whether arg3 has been completed (arg2 must fit in `callReq` in order for mutation to be allowed)
- the running checksum for all the frames in the call

once a `callReq` frame has been mutated, all subsequent `callReqContinue` frames with the same ID will have their checksums recalculated and updated before the relayer forwards them to the destination.

The `TestRelayModifyArg2` tests have also been updated to include "large" payloads (64, 128, and 512kB) to ensure that we don't have future regression in this behavior.

## Update
 dropped the argComplete flags since:
- arg1complete and arg2complete are always true, since we don’t support arg2 mutation when it doesn’t fit in the first frame
- if arg2 was enlarged to go over the first frame, it would’ve been already relayed by the time we hit callReqContinue
- arg3complete is never true until the last frame, so we don’t need to check that either

### DIFF
--- a/fragmenting_writer.go
+++ b/fragmenting_writer.go
@@ -54,6 +54,9 @@ type writableFragment struct {
 func (f *writableFragment) finish(hasMoreFragments bool) {
 	f.checksumRef.Update(f.checksum.Sum())
 	if hasMoreFragments {
+		// Important: hasMoreFragmentsFlag is set if there are more fragments, but NOT CLEARED if there aren't.
+		// This allows for callReqContinue frames to follow a fragmented callReq frame e.g. when arg2 is modified
+		// by the relayer
 		f.flagsRef.Update(hasMoreFragmentsFlag)
 	} else if !f.dontReleaseChecksum {
 		f.checksum.Release()
--- a/fragmenting_writer.go
+++ b/fragmenting_writer.go
@@ -42,11 +42,12 @@ const (
 // for contents, a running checksum, and placeholders for the fragment flags
 // and final checksum value
 type writableFragment struct {
-	flagsRef    typed.ByteRef
-	checksumRef typed.BytesRef
-	checksum    Checksum
-	contents    *typed.WriteBuffer
-	frame       *Frame
+	flagsRef            typed.ByteRef
+	checksumRef         typed.BytesRef
+	checksum            Checksum
+	contents            *typed.WriteBuffer
+	frame               *Frame
+	dontReleaseChecksum bool
 }
 
 // finish finishes the fragment, updating the final checksum and fragment flags
@@ -54,7 +55,7 @@ func (f *writableFragment) finish(hasMoreFragments bool) {
 	f.checksumRef.Update(f.checksum.Sum())
 	if hasMoreFragments {
 		f.flagsRef.Update(hasMoreFragmentsFlag)
-	} else {
+	} else if !f.dontReleaseChecksum {
 		f.checksum.Release()
 	}
 }
--- a/fragmenting_writer.go
+++ b/fragmenting_writer.go
@@ -42,12 +42,11 @@ const (
 // for contents, a running checksum, and placeholders for the fragment flags
 // and final checksum value
 type writableFragment struct {
-	flagsRef            typed.ByteRef
-	checksumRef         typed.BytesRef
-	checksum            Checksum
-	contents            *typed.WriteBuffer
-	frame               *Frame
-	dontReleaseChecksum bool
+	flagsRef    typed.ByteRef
+	checksumRef typed.BytesRef
+	checksum    Checksum
+	contents    *typed.WriteBuffer
+	frame       *Frame
 }
 
 // finish finishes the fragment, updating the final checksum and fragment flags
@@ -58,7 +57,7 @@ func (f *writableFragment) finish(hasMoreFragments bool) {
 		// This allows for callReqContinue frames to follow a fragmented callReq frame e.g. when arg2 is modified
 		// by the relayer
 		f.flagsRef.Update(hasMoreFragmentsFlag)
-	} else if !f.dontReleaseChecksum {
+	} else {
 		f.checksum.Release()
 	}
 }
--- a/relay.go
+++ b/relay.go
@@ -208,6 +208,29 @@ const (
 	responseFrame frameType = 1
 )
 
+type mutatedCallState struct {
+	arg2Complete bool
+	arg3Complete bool
+	checksum     Checksum
+}
+
+func (cs *mutatedCallState) updateChecksumFromArg(rbuf *typed.ReadBuffer, complete *bool) {
+	if *complete {
+		return
+	}
+
+	n := rbuf.ReadUint16()
+	if n == 0 {
+		*complete = true
+		return
+	}
+
+	cs.checksum.Add(rbuf.ReadBytes(int(n)))
+	if rbuf.BytesRemaining() > 0 {
+		*complete = true
+	}
+}
+
 // A Relayer forwards frames.
 type Relayer struct {
 	relayHost      RelayHost
@@ -237,6 +260,8 @@ type Relayer struct {
 	relayConn *relay.Conn
 	logger    Logger
 	pending   atomic.Uint32
+
+	mutatedCalls map[uint32]*mutatedCallState
 }
 
 // NewRelayer constructs a Relayer.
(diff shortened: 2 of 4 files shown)


fragmenting_writer.go
```diff
@@ -42,19 +42,20 @@ const (
...
-	checksumRef typed.BytesRef
-	checksum    Checksum
-	contents    *typed.WriteBuffer
-	frame       *Frame
+	flagsRef            typed.ByteRef
+	checksumRef         typed.BytesRef
+	checksum            Checksum
+	contents            *typed.WriteBuffer
+	frame               *Frame
+	dontReleaseChecksum bool
```
prashantv, line 50: This is a bit of an odd behaviour, some questions:
 * is there another way to do this without the if. e.g., can we move the responsibility of releasing the checksum to the caller of the non-relay codepath?
 * alternative to adding a bool + if check, what if we provide a decorator for `Checksum` that passes through all calls but `Reset`()`. Then when we create the `writableFragment`, we just wrap the checksum in our `noreleaseChecksum` type.

relay.go
```diff
@@ -281,6 +282,22 @@ func (r *Relayer) Relay(f *Frame) (shouldRelease bool, _ error) {
...
+func (r *Relayer) updateCallReqContinueChecksum(f *Frame, cs Checksum) {
+	rbuf := typed.NewReadBuffer(f.SizedPayload())
+	rbuf.SkipBytes(1) // flags
+	rbuf.SkipBytes(1) // checksum type
+
+	checksumRef := typed.BytesRef(rbuf.ReadBytes(cs.Size()))
+
+	n := rbuf.ReadUint16()
+	if n == 0 {
+		return
```
prashantv, line 294: can we add a log here?

do we have any way to get coverage here?

relay.go
```diff
@@ -281,6 +282,22 @@ func (r *Relayer) Relay(f *Frame) (shouldRelease bool, _ error) {
 	return r.handleCallReq(cr)
 }
 
+func (r *Relayer) updateCallReqContinueChecksum(f *Frame, cs Checksum) {
```
prashantv, line 285: add comments here on why we only read a single arg (e.g., if we're here, arg1 and arg2 must be complete)

prashantv, line 285: nit: maybe rename `updateMutatedCallReqContinueChecksum`? to indicate this only needs to happen with mutation

relay.go
```diff
@@ -281,6 +282,22 @@ func (r *Relayer) Relay(f *Frame) (shouldRelease bool, _ error) {
...
+	rbuf := typed.NewReadBuffer(f.SizedPayload())
+	rbuf.SkipBytes(1) // flags
+	rbuf.SkipBytes(1) // checksum type
```
prashantv, line 288: do you want to note that this should match the current checksum's type? don't think the check in the code is necessary, but the note may be useful

relay.go
```diff
@@ -530,6 +553,10 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return nil
 	}
 
+	if f.messageType() == messageTypeCallReqContinue && item.mutatedChecksum != nil {
+		r.updateCallReqContinueChecksum(f, item.mutatedChecksum)
+	}
```
prashantv, line 558: can you add a short comment about this case

relay.go
```diff
@@ -562,6 +589,7 @@ func (r *Relayer) addRelayItem(isOriginator bool, id, remapID uint32, destinatio
 	items := r.inbound
 	if isOriginator {
 		items = r.outbound
+		item.mutatedChecksum = mutatedChecksum
```
prashantv, line 592: you can set this in the `relayItem` constructor, since it'll only be set for `isOriginator == true` anyway

relay.go
```diff
@@ -836,11 +871,12 @@ func (rfs *relayFragmentSender) newFragment(initial bool, checksum Checksum) (*w
...
-		checksumRef: checksumRef,
-		checksum:    checksum,
-		contents:    contents,
-		frame:       frame,
+		flagsRef:            flagsRef,
+		checksumRef:         checksumRef,
+		checksum:            checksum,
+		contents:            contents,
+		frame:               frame,
+		dontReleaseChecksum: true,
```
prashantv, line 879: add a comment referencing when we release the checksum

relay_test.go
```diff
@@ -1584,18 +1586,46 @@ func TestRelayModifyArg2(t *testing.T) {
 			arg3: []byte{},
 		},
 		{
-			msg: "small payloads",
+			msg: "1kB payloads",
```
prashantv, line 1589: couple more corner cases just to verify we're not missing other issues: large header values that causes arg2 to end right near the boundary of the first frame (ideally both a few bytes before, 2 bytes before, 1 byte before, and on the exact boundary)

Ideally we'd test that with all sort of arg3 values, so i think we may want to do all combinations of:
 * no arg2, small arg2, large arg2 (3 or 4 cases that are close/on the boundary)
 * no arg3, small arg3,  16kb arg3, 32kb arg3, 64kb arg3, 128kb arg3, 1mb arg3

(I can help with the framework part of the test in a separate PR)

prashantv (commented):
Looks good, just a few minor comments, but think this should work functionally

relay.go
cinchurge: some discoveries/conclusions thanks to some offline pairing with @prashantv:
- The relay code will consider arg2 fragmented when it encounters a call without arg3. Thus it is nearly impossible to successfully relay a call where arg2 fills the entire `calReq` frame
- A `callReqContinue` frame with 0-length arg3 can be deliberately created by writing 0 bytes to arg3, then flushing before closing the writer. Although uncommon, it can still happen in practice
- The code above is incorrect in that it doesn't update the `callReqContinue` frame with the running checksum, which will result in inconsistent checksums for the call and fail that call

Since the following code correctly handles the case where len(arg3) == 0, we've simply removed the condition.

relay_test.go
cinchurge: added TODO for following up afterward

relay_test.go
```diff
@@ -1496,85 +1498,174 @@ func inspectFrames(rh *relaytest.StubRelayHost) chan relay.CallFrame {
...
+	}
+	return m
+}
+
 func TestRelayModifyArg2(t *testing.T) {
+	const kb = 1024
 	largeVal1 := testutils.RandString(16 * 1024)
 	largeVal2 := testutils.RandString(16 * 1024)
 	largeVal3 := testutils.RandString(16 * 1024)
 	largeVal4 := testutils.RandString(16 * 1024)
```
prashantv, line 1574:
```suggestion
	largeVal1 := testutils.RandString(16 * kb)
	largeVal2 := testutils.RandString(16 * kb)
	largeVal3 := testutils.RandString(16 * kb)
	largeVal4 := testutils.RandString(16 * kb)
```

relay_test.go
```diff
@@ -1496,85 +1498,174 @@ func inspectFrames(rh *relaytest.StubRelayHost) chan relay.CallFrame {
...
+			},
+		},
+		{
+			msg: "fill the first frame to 2 bytes over its boundary",
+			modifier: func(t *testing.T, cst ChecksumType, arg1 string, arg2 map[string]string) relayModifier {
+				return fillFrameWithArg2(t, cst, arg1, arg2, 2)
+			},
+		},
+		{
+			msg: "add large key/value",
```
prashantv, line 1653: can we expand this into 2 tests:
 - this one pushes the arg2 into a second frame
 - let's add another one which pushes arg2 into a 2nd and 3rd frame (e.g., we're adding closer to 128kb)

relay_test.go
```diff
@@ -1586,18 +1677,46 @@ func TestRelayModifyArg2(t *testing.T) {
...
+		{
+			msg: "128kB payloads",
+			arg2: map[string]string{
+				"existingKey": "existingValue",
+			},
+			arg3: testutils.RandBytes(128 * kb),
+		},
+		{
+			msg: "1MB payloads",
+			arg2: map[string]string{
```
prashantv, line 1716: let's add a case where arg2 is 60kb as well

prashantv (approved):

prashantv (approved):

PR #814, opened 2020-12-13
relay: Send error frame to caller on fragmented arg2

prashantv:
Currently, if a call fails due to a fragmented arg2 (e.g., arg2 does not
fit in the first frame as it's too large), we drop the frame and log
an error internally, but no error is returned to the caller, so they
see a timeout.

Instead of a timeout, fail the relay item, which triggers an error frame
to be sent to the caller, so they receive mroe information on why the
call failed (and they receive an immediate error instead of waiting for
the timeout to expire).

This also improves other error frames sent back with the same
reason that is used in metrics, although we still have a generic
"frame was not sent to remote side" which could be replaced with
much more specific errors.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -1693,7 +1693,7 @@ func TestRelayModifyArg2OnFrameWithFragmentedArg2ShouldFail(t *testing.T) {
 	opts := testutils.NewOpts().
 		SetRelayOnly().
 		SetRelayHost(rh).
-		AddLogFilter("Failed to relay frame.", 1, "error", "fragmented arg2 not supported for appends")
+		AddLogFilter("Failed to send call with modified arg2.", 1)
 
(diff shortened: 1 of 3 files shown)


relay_test.go
```diff
@@ -1716,7 +1716,8 @@ func TestRelayModifyArg2OnFrameWithFragmentedArg2ShouldFail(t *testing.T) {
 				"fum": testutils.RandString(16 * 1024),
 			}),
 		})
-		assert.EqualError(t, err, "tchannel error ErrCodeTimeout: timeout")
```
cinchurge, line 1719: ah, somehow this didn't stand out when we were testing this....

cinchurge (approved):

relay_test.go
prashantv: yeah I totally missed this too!

PR #815, opened 2020-12-14
relay: Reject arg2 modifications for non-Thrift calls

prashantv:
Arg2 modification assumes the Thrift encoding, and will not work
correctly for any other encoding. To avoid any unexpected corruption
of headers, reject mutating headers for any non-Thrift calls.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -1419,13 +1419,13 @@ func TestRelayAppendArg2SentBytes(t *testing.T) {
 		{
 			msg:           "without appends",
 			arg3:          []byte("hello, world"),
-			wantSentBytes: 127,
+			wantSentBytes: 130,
 		},
 		{
 			msg:           "with appends",
 			arg3:          []byte("hello, world"),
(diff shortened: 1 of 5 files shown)


relay_test.go
```diff
@@ -1685,41 +1686,75 @@ func TestRelayModifyArg2(t *testing.T) {
...
+
+				err := testutils.CallEcho(caller, ts.HostPort(), ts.ServiceName(), &raw.Args{
+					Format: tt.format,
+					Arg2:   tt.arg2,
+				})
+				require.Error(t, err, "should fail to send call with large arg2")
+				assert.Contains(t, err.Error(), tt.wantErr, "unexpected error")
+
+				// Even after a failure, a simple call should still suceed (e.g., connection is left in a safe state).
+				err = testutils.CallEcho(caller, ts.HostPort(), ts.ServiceName(), &raw.Args{
```
cinchurge, line 1749: good idea to test this!

cinchurge (approved):

PR #819, opened 2021-02-13
connections: don't call conncontext on outbound base context

cinchurge:
The `ConnContext()` function was originally meant for connection-specific initialization of the server's base context, but we are now also invoking it on the connection we freshly created for outbound connections, which wouldn't make sense since the outbound connection we created couldn't possibly have any caller information. This would end up overriding the base context meant to be attached to the outbound connection, resulting in unintended behavior.

This change changes `Channel.newConnection()` so that we only invoke `ConnContext()` on inbound connections, while leaving the base context unchanged for outbound connections.


cinchurge:
Abandoning this in favor of just checking in the ConnContext for keys that are already set and avoiding setting them repeatedly.

PR #820, opened 2021-02-26
Drop logs when idle connections are closed

prashantv:
We currently have 2 logs whenever an idle connection is closed:
 * "Closing idle inbound connection"
 * "Connection closing." with reason="Idle connection closed"

The first log doesn't provide much value since the closing log already gives us
a detailed reason. Since it's redundant, drop the first log, and pass any fields
that only the first log had to the `close`, so they are included in the second log.

### DIFF
--- a/idle_sweep.go
+++ b/idle_sweep.go
@@ -119,11 +119,10 @@ func (is *idleSweep) checkIdleConnections() {
 			continue
 		}
 
-		is.ch.log.WithFields(
-			LogField{"remotePeer", conn.remotePeerInfo},
+		conn.close(
+			LogField{"reason", "Idle connection closed"},
 			LogField{"lastActivityTimeRead", conn.getLastActivityReadTime()},
 			LogField{"lastActivityTimeWrite", conn.getLastActivityWriteTime()},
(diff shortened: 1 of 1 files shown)


idle_sweep.go
```diff
@@ -119,11 +119,10 @@ func (is *idleSweep) checkIdleConnections() {
 			continue
 		}
 
-		is.ch.log.WithFields(
-			LogField{"remotePeer", conn.remotePeerInfo},
```
witriew, line 123: Looks like we lose `remotePeerInfo.Version` when we get rid of this since it's not naturally logged as part of the Connection's logger, https://github.com/uber/tchannel-go/blob/dev/connection.go#L310.  Do we care about that?

witriew (approved):

idle_sweep.go
prashantv: We add the remote peer info a little later,
https://github.com/uber/tchannel-go/blob/dev/connection.go#L390

witriew: Ah, okay. 👍

PR #821, opened 2021-03-03
Fixes for compatibility with Go 1.16

cinchurge:
Ref https://golang.org/doc/go1.16#go-command

go modules are now enabled by default, so `go get` would fail
without explicitly turning it off via GO111MODULE=off

additionally, `go build -i` now returns an error during tests. since
that flag has been deprecated, remove it so the tests can pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,19 @@ cache:
   - $HOME/.glide/cache
 
 go:
-- 1.13.x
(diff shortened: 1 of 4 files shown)


prashantv (approved):

prashantv:
Can we also update `.travis.yml` to test against `stable` and `oldstable`?

PR #826, opened 2021-05-19
relay: Prefactor relay_messagest_test for flexible payload sizes

prashantv:
Currently, the test frame generation has some issues:
 * It assumes a fixed frame size (200 bytes)
 * It ignores errors from the typed Writer

Update the test to set the frame payload size based on the contents of
the payload, and to check for errors in the typed Writer.

### DIFF
--- a/relay_messages_test.go
+++ b/relay_messages_test.go
@@ -59,6 +59,7 @@ type testCallReqParams struct {
 	argScheme       Format
 	arg2Buf         []byte
 	overrideArg2Len int
+	skipArg3        bool
 	arg3Buf         []byte
 }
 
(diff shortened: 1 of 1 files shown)


relay_messages_test.go
```diff
@@ -67,20 +68,24 @@ func (cr testCallReq) req(tb testing.TB) lazyCallReq {
...
-	f := NewFrame(200)
+	f := NewFrame(MaxFramePayloadSize)
 	fh := fakeHeader()
-	fh.size = 0xD8 // 200 + 16 bytes of header = 216 (0xD8)
-	f.Header = fh
-	fh.write(typed.NewWriteBuffer(f.headerBuffer))
+
+	// Set the size in the header and write out the header after we know the payload contents.
+	defer func() {
+		fh.size = uint16(len(f.Payload)) + 16 // add 16 bytes for the header header
```
cinchurge, line 85: for the header header?

cinchurge (approved):
lgtm except typo in comment

PR #830, opened 2021-07-29
Extend RelayCall to inspect the call response frame

cinchurge:
In order to support features where inspection of the response frame is required, extend `RelayCall` with an additional `CallResponse()` method which can be used for inspecting the response frame of a call. A new interface `RespFrame` is added which mirrors the `CallFrame` passed into `RelayHost.Start()`.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -554,7 +554,7 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 
 	switch f.messageType() {
 	case messageTypeCallRes:
-		// Invoke call.CallResponse() if we got the response frame
+		// Invoke call.CallResponse() if we get a valid call response frame.
 		cr, err := newLazyCallRes(f)
 		if err == nil {
 			item.call.CallResponse(cr)
--- a/relay.go
+++ b/relay.go
@@ -552,19 +552,24 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return nil
 	}
 
-	if f.messageType() == messageTypeCallRes {
+	switch f.messageType() {
+	case messageTypeCallRes:
+		// Invoke call.CallResponse() if we got the response frame
 		cr, err := newLazyCallRes(f)
-		if err != nil {
-			r.logger.Error(err.Error())
-		} else {
+		if err == nil {
 			item.call.CallResponse(cr)
+		} else {
+			r.logger.WithFields(
+				ErrField(err),
+				LogField{"id", f.Header.ID},
+			).Error("Malformed callRes frame.")
 		}
-	}
-
-	// Recalculate and update the checksum for this frame if it has non-nil item.mutatedChecksum
-	// (meaning the call was mutated) and it is a callReqContinue frame.
-	if f.messageType() == messageTypeCallReqContinue && item.mutatedChecksum != nil {
-		r.updateMutatedCallReqContinueChecksum(f, item.mutatedChecksum)
+	case messageTypeCallReqContinue:
+		// Recalculate and update the checksum for this frame if it has non-nil item.mutatedChecksum
+		// (meaning the call was mutated) and it is a callReqContinue frame.
+		if item.mutatedChecksum != nil {
+			r.updateMutatedCallReqContinueChecksum(f, item.mutatedChecksum)
+		}
 	}
 
 	// Track sent/received bytes. We don't do this before we check
--- a/relay.go
+++ b/relay.go
@@ -552,6 +552,15 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return nil
 	}
 
+	if f.messageType() == messageTypeCallRes {
+		cr, err := newLazyCallRes(f)
+		if err != nil {
+			r.logger.Error(err.Error())
+		} else {
+			item.call.CallResponse(cr)
+		}
+	}
+
 	// Recalculate and update the checksum for this frame if it has non-nil item.mutatedChecksum
 	// (meaning the call was mutated) and it is a callReqContinue frame.
 	if f.messageType() == messageTypeCallReqContinue && item.mutatedChecksum != nil {
@@ -812,7 +821,7 @@ func determinesCallSuccess(f *Frame) (succeeded bool, failMsg string) {
 		msg := newLazyError(f).Code().MetricsKey()
 		return false, msg
 	case messageTypeCallRes:
-		if newLazyCallRes(f).OK() {
+		if isCallResOK(f) {
 			return true, ""
 		}
 		return false, "application-error"
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -67,6 +67,18 @@ type CallFrame interface {
 	Arg2Append(key, val []byte)
 }
 
+// RespFrame is an interface that abstracts access to the CallRes frame
+type RespFrame interface {
+	// OK indicates whether the call was successful
+	OK() bool
+
+	// Arg2 returns the raw arg2 payload
+	Arg2() []byte
+
+	// Arg2Iterator returns an iterator for iterating over arg2 headers
+	Arg2Iterator() (arg2.KeyValIterator, error)
+}
+
 // Conn contains information about the underlying connection.
 type Conn struct {
 	// RemoteAddr is the remote address of the underlying TCP connection.
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -72,11 +72,11 @@ type RespFrame interface {
 	// OK indicates whether the call was successful
 	OK() bool
 
+	// ArgScheme returns the scheme of the arg
+	ArgScheme() []byte
+
 	// Arg2 returns the raw arg2 payload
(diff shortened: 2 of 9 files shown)


relay.go
```diff
@@ -552,6 +552,15 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return nil
 	}
 
+	if f.messageType() == messageTypeCallRes {
+		cr, err := newLazyCallRes(f)
+		if err != nil {
+			r.logger.Error(err.Error())
```
prashantv, line 558: This won't have much useful context if we log errors in this way, can we have more information:
 * a message indicating that a call res frame was not parsed
 * context on the call (e.g., the ID, any info about the item that may be useful)

relay.go
```diff
@@ -552,6 +552,15 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return nil
 	}
 
+	if f.messageType() == messageTypeCallRes {
```
prashantv, line 555: since we hae another `messageType()` check below, can we use a switch for both these cases?

relay/relay.go
```diff
@@ -67,6 +67,18 @@ type CallFrame interface {
 	Arg2Append(key, val []byte)
 }
 
+// RespFrame is an interface that abstracts access to the CallRes frame
+type RespFrame interface {
+	// OK indicates whether the call was successful
+	OK() bool
+
+	// Arg2 returns the raw arg2 payload
+	Arg2() []byte
```
prashantv, line 76: Do we want to expose arg2?

relay_api.go
```diff
@@ -20,7 +20,9 @@
 
 package tchannel
 
-import "github.com/uber/tchannel-go/relay"
+import (
+	"github.com/uber/tchannel-go/relay"
+)
```
prashantv, line 25: nit: can revert this change

relay_messages.go
```diff
@@ -272,3 +342,13 @@ func finishesCall(f *Frame) bool {
 		return false
 	}
 }
+
+// isCallResOK indicates whether the call was successful
+func isCallResOK(f *Frame) bool {
+	return f.Payload[_resCodeIndex] == _resCodeOK
+}
```
prashantv, line 349: should we use this method to implement `lazyCallRes.OK` (so we don't have 2 different ways to read this field)

also means we can drop a few fields from the `lazyCallRes` struct

relay_messages_test.go
```diff
@@ -161,58 +162,73 @@ func withLazyCallReqCombinations(f func(cr testCallReq)) {
...
-		params.headers = map[string]string{
-			"k1":      "v1",
-			"k222222": "",
-			"k3":      "thisisalonglongkey",
-		}
+		params.headers["k1"] = "v1"
+		params.headers["k222222"] = ""
+		params.headers["k3"] = "thisisalonglongkey"
+	}
+	if cr&(resHasArg2|resHasFragmentedArg2) != 0 {
```
prashantv, line 206: why do we check `HasFragmentedArg2` here?

relay_test.go
```diff
@@ -1420,6 +1420,33 @@ func TestRelayTransferredBytes(t *testing.T) {
 	})
 }
 
+func TestRelayCallResponse(t *testing.T) {
```
prashantv, line 1423: Can we add an integration test that sends response headers, and validates them via the `RespFrameFn`?

relay/relay.go
cinchurge: this is primarily for testing since the `thrift` package imports `tchannel`, and we can't really test arg2 iteration with the raw scheme. an alternative is to do a type assertion and inspect `lazyCallRes` directly, but even that would require exporting arg2 somehow. is there any other way we can test without exporting an arg2 accessor?

relay_api.go
cinchurge: not sure why? we're passing in `relay.RespFrame` below

relay_messages_test.go
cinchurge: so the arg scheme is set to `thrift`. we only support arg2 iteration for `thrfit`, and the two types here are meant for adding key/vals to arg2

relay_test.go
cinchurge: not sure if i'm missing something - in this test the response header is set by the echo handler, copied over from the call, and we verify that the bytes in arg2 are exactly what we send. is there something else you had in mind?

relay.go
```diff
@@ -552,19 +552,24 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 		return nil
 	}
 
-	if f.messageType() == messageTypeCallRes {
+	switch f.messageType() {
+	case messageTypeCallRes:
+		// Invoke call.CallResponse() if we got the response frame
```
prashantv, line 557:
```suggestion
		// Invoke call.CallResponse() if we get a valid call response frame.
```

relay.go
```diff
@@ -552,19 +552,24 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
...
 		cr, err := newLazyCallRes(f)
-		if err != nil {
-			r.logger.Error(err.Error())
-		} else {
+		if err == nil {
 			item.call.CallResponse(cr)
+		} else {
+			r.logger.WithFields(
```
prashantv, line 562: can we get coverage for this?

relay/relay.go
prashantv: ahh, I see. What are we doing for testing arg2 on the call req, can we do the same?

One potential follow up to maybe simplify things, we could change both the lazyCallReq and lazyCallRes to only export `Arg2() []byte` and `ArgScheme() []byte`, and then have the caller check the scheme, and if it's thrift, they call `NewKeyValueIterator` on the `Arg2() []byte` directly,
https://pkg.go.dev/github.com/uber/tchannel-go@v1.21.2/thrift/arg2#NewKeyValIterator

relay_api.go
prashantv: Sorry, what I meant is that the import matches the same as what it was previously, but it's grouped (which isn't necessary since there's only one import)

relay_messages_test.go
prashantv: ah, would it be clearer to use a separate orthogonal bit like `resIsThrift` since it is possible to have a fragmented arg2 without thrift (though we may not necessarily find it useful in our tests)

relay_test.go
prashantv: What I was thinking was that instead of verifying the arg2 bytes, we actually response with thift response headers so arg2 is an encoded header map, and we validate the values map.

I think we do something like that on the request side:
https://github.com/uber/tchannel-go/blob/dev/relay_test.go#L1233

relay/relay.go
cinchurge: we do export an accessor for the arg2 offset in the callReq, but it really doesn't feel any better than just exporting the slice directly. I do think though that exposing `Arg2()` and `Scheme()` is a good idea. I'll add a TODO for that.

cinchurge: Added `Scheme()` to `RespFrame`

relay_messages.go
```diff
@@ -80,17 +79,81 @@ func (e lazyError) Code() SystemErrCode {
...
+
+	rbuf := typed.NewReadBuffer(f.SizedPayload())
+	rbuf.SkipBytes(1)           // flags
+	rbuf.SkipBytes(1)           // code
+	rbuf.SkipBytes(_spanLength) // tracing
+
+	var as []byte
+	nh := int(rbuf.ReadSingleByte())
+	for i := 0; i < nh; i++ {
+		hk := int(rbuf.ReadSingleByte())
```
prashantv, line 101: not sure what `hk` stands for, perhaps `keyLen` and `valLen`?

relay_test.go
```diff
@@ -34,6 +34,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/uber/tchannel-go/thrift/arg2"
```
prashantv, line 37: nit: import ordering

relay_test.go
```diff
@@ -931,6 +933,68 @@ func TestRelayStalledClientConnection(t *testing.T) {
...
+		defer client.Close()
+
+		ctx, cancel := NewContext(testutils.Timeout(time.Second))
+		defer cancel()
+
+		// Data to fit one frame fully, but large enough that a number of these frames will fill
+		// all the buffers and cause the relay to drop the response frame. Buffers are:
+		// 1. Relay's sendCh on the connection to the client (set to 10 frames explicitly)
+		// 2. Relay's TCP send buffer for the connection to the client.
+		// 3. Client's TCP receive buffer on the connection to the relay.
```
prashantv, line 981: don't think this is relevant to this test? we don't need any drops to happen for this test right?

prashantv (approved):

relay_test.go
cinchurge: sorry, copypasta

PR #831, opened 2021-07-31
[tchannel-go] Add options to support handlers with ignoring methods

AllenLuUber:
If a user wants to migrate off tchannel-go server on a per method level, we need a way to allow users to specify: `for those specified methods, tchannel-go should handle it as before. But for others, let passed-in handler handle it`. This PR provides an additional option to allow such use case for gradual migration.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -124,10 +124,12 @@ type ChannelOptions struct {
 	// default handler that delegates to a subchannel.
 	Handler Handler
 
-	// HandlerWithIgnore allows users to configure TChannel server such that
-	// requests with specified methods can be ignored by passed-in handler and natively by TChannel.
-	// All other methods will be handled by passed-in handler.
-	HandlerWithIgnore HandlerWithIgnore
+	// IgnoreMethods allow users to configure TChannel server such that
+	// requests with specified methods can be ignored by the above passed-in handler
+	// and handled natively by TChannel.
+	// Requests with other methods will be handled by passed-in handler.
+	// This is useful for the gradual migration purpose.
+	IgnoreMethods []string
 
 	// Dialer is optional factory method which can be used for overriding
 	// outbound connections for things like TLS handshake
@@ -137,16 +139,6 @@ type ChannelOptions struct {
 	// the per-connection base context. This context is used as the parent context
 	// for incoming calls.
 	ConnContext func(ctx context.Context, conn net.Conn) context.Context
-}
-
-// HandlerWithIgnore allows users to configure TChannel server such that
-// requests with method listed in NativeHanderMethods are
-// ignored by passed-in handler and natively by TChannel.
-// All other methods will be handled by passed-in handler.
-type HandlerWithIgnore struct {
-	Handler Handler
-	// should we generalize to a func(request) bool here?
-	IgnoreMethods []string
 }
 
 // ChannelState is the state of a channel.
@@ -319,15 +311,15 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	}
 	ch.peers = newRootPeerList(ch, opts.OnPeerStatusChanged).newChild()
 
-	switch n := opts.HandlerWithIgnore; {
+	switch {
+	case len(opts.IgnoreMethods) > 0 && opts.Handler != nil:
+		ch.handler = userHandlerWithIgnore{
+			localHandler:      channelHandler{ch},
+			ignoreUserHandler: toStringSet(opts.IgnoreMethods),
+			userHandler:       opts.Handler,
+		}
 	case opts.Handler != nil:
 		ch.handler = opts.Handler
-	case len(n.IgnoreMethods) > 0 && n.Handler != nil:
-		ch.handler = userHandlerWithIgnore{
-			ch:          ch,
-			ignore:      toStringSet(n.IgnoreMethods),
-			userHandler: n.Handler,
-		}
 	default:
 		ch.handler = channelHandler{ch}
 	}
--- a/channel.go
+++ b/channel.go
@@ -314,7 +314,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	switch {
 	case len(opts.IgnoreMethods) > 0 && opts.Handler != nil:
 		ch.handler = userHandlerWithIgnore{
-			localHandler:      channelHandler{ch},
+			channelHandler:    channelHandler{ch},
 			ignoreUserHandler: toStringSet(opts.IgnoreMethods),
 			userHandler:       opts.Handler,
 		}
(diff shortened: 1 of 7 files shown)


channel.go
```diff
@@ -134,6 +139,16 @@ type ChannelOptions struct {
 	ConnContext func(ctx context.Context, conn net.Conn) context.Context
 }
 
+// HandlerWithIgnore allows users to configure TChannel server such that
+// requests with method listed in NativeHanderMethods are
+// ignored by passed-in handler and natively by TChannel.
+// All other methods will be handled by passed-in handler.
+type HandlerWithIgnore struct {
+	Handler Handler
```
prashantv, line 147: since we already take a `Handler` above, we could just take `HandlerIgnores` struct, and apply any ignores on top of the existing `Handler` field?

channel.go
```diff
@@ -134,6 +139,16 @@ type ChannelOptions struct {
...
+	// should we generalize to a func(request) bool here?
```
prashantv, line 148: good question, since we have a struct here, we can always add more in the future, OK with starting simple

handlers.go
```diff
@@ -125,3 +126,25 @@ type channelHandler struct{ ch *Channel }
...
+	userHandler Handler
+}
+
+func (u userHandlerWithIgnore) Handle(ctx context.Context, call *InboundCall) {
+	var sb strings.Builder
+	sb.WriteString(call.ServiceName())
+	sb.WriteString("::")
+	sb.Write(call.Method())
+
+	if _, ok := u.ignore[sb.String()]; ok {
```
prashantv, line 145: nice use of `strings.Builder`, can you double-check that it doesn't cause any allocations in doing this check?

handlers.go
```diff
@@ -125,3 +126,25 @@ type channelHandler struct{ ch *Channel }
 func (c channelHandler) Handle(ctx context.Context, call *InboundCall) {
 	c.ch.GetSubChannel(call.ServiceName()).handler.Handle(ctx, call)
 }
+
+// channelHandlerWithNative is a Handler that wraps a Channel and delegates requests
+// to SubChannels if the inbound call's service and method name are configured to
+// ignore passed-in thirdparty handlers.
+type userHandlerWithIgnore struct {
+	ch          *Channel
```
prashantv, line 134: instead of taking the channel, suggest taking a `Handler` for the ignored cases, and that handler can be set to `channelHandler{ch}` at the caller (same as if no ignores were set) vs doing a subchannel lookup ourselves

handlers_test.go
```diff
@@ -57,3 +59,58 @@ func TestHandlers(t *testing.T) {
...
+		method:       []byte(method),
+		methodString: method,
+		conn:         conn,
+	}
+	ignoreCall := &InboundCall{
+		serviceName:  svc,
+		method:       []byte(ignoreMethod),
+		methodString: ignoreMethod,
+		conn:         conn,
+	}
```
prashantv, line 102: instead of creating these objects ourselves, let's create a real channel with the options set, and validate end-to-end

we generally avoid tests that create partial objects in the tchannel codebase, and use more integration-style tests

channel.go
AllenLuUber: My original thought was to not confuse the user. If you provide this `HandlerWithIgore`, you have to provide both Handler and the to-ignore list. But looking at your comment, I think a single `[]string IgnoreMethod` should be enough. If they only "forgot" to provide this list, it will go back to the existing behavior: handle all. This is more concise. I like it.

handlers.go
AllenLuUber: [source code](https://golang.org/src/strings/builder.go) says it all with unsafe package :)
```go
func (b *Builder) String() string {
	return *(*string)(unsafe.Pointer(&b.buf))
}
```

prashantv: True, but we need to confirm that the `[]byte` doesn't escape too, not just the `[]byte` --> `string` conversion.

AllenLuUber: It seems no escape from the `Write` implementation. The `append` should not cause an escape.  Ideally we could pre-allocate a `[]byte` and use unsafe to make a string by ourselves.
```go
// Write appends the contents of p to b's buffer.
// Write always returns len(p), nil.
func (b *Builder) Write(p []byte) (int, error) {
	b.copyCheck()
	b.buf = append(b.buf, p...)
	return len(p), nil
}

// WriteByte appends the byte c to b's buffer.
// The returned error is always nil.
func (b *Builder) WriteByte(c byte) error {
	b.copyCheck()
	b.buf = append(b.buf, c)
	return nil
}

// noescape hides a pointer from escape analysis.  noescape is
// the identity function but escape analysis doesn't think the
// output depends on the input. noescape is inlined and currently
// compiles down to zero instructions.
// USE CAREFULLY!
// This was copied from the runtime; see issues 23382 and 7921.
//go:nosplit
//go:nocheckptr
func noescape(p unsafe.Pointer) unsafe.Pointer {
	x := uintptr(p)
	return unsafe.Pointer(x ^ 0)
}

func (b *Builder) copyCheck() {
	if b.addr == nil {
		// This hack works around a failing of Go's escape analysis
		// that was causing b to escape and be heap allocated.
		// See issue 23382.
		// TODO: once issue 7921 is fixed, this should be reverted to
		// just "b.addr = b".
		b.addr = (*Builder)(noescape(unsafe.Pointer(b)))
	} else if b.addr != b {
		panic("strings: illegal use of non-zero Builder copied by value")
	}
}
```

prashantv: I tried to replicate this code in a simple benchmark, and found that it did allocate,
```go
package main

import (
        "strings"
        "testing"
)

type Call struct{ Service, Method string }

var ignores = map[string]bool{"foo::bar": true}

func canHandle(c Call) bool {
        var sb strings.Builder
        sb.WriteString(c.Service)
        sb.WriteString("::")
        sb.WriteString(c.Method)

        return ignores[sb.String()]
}

func BenchmarkEscape(b *testing.B) {
        for i := 0; i < b.N; i++ {
                if canHandle(Call{"foo", "bar"}) {

                }
        }
}
```

output of benchmark:
```
BenchmarkEscape         33148530                36.5 ns/op             8 B/op          1 allocs/op
```

and pprof shows that the `strings.Builder.WriteString` is leading to the alloc:
```
(pprof) list WriteString
Total: 244MB
ROUTINE ======================== strings.(*Builder).WriteString in /home/prashant/.gimme/versions/go1.15.5.linux.amd64/src/strings/builder.go
     244MB      244MB (flat, cum)   100% of Total
         .          .    118:
         .          .    119:// WriteString appends the contents of s to b's buffer.
         .          .    120:// It returns the length of s and a nil error.
         .          .    121:func (b *Builder) WriteString(s string) (int, error) {
         .          .    122:   b.copyCheck()
     244MB      244MB    123:   b.buf = append(b.buf, s...)
         .          .    124:   return len(s), nil
         .          .    125:}
```

We're allocating a `[]byte` to write to, and I think that's being allocated on the heap rather than the stack.

Can you verify with a similar benchmark here to see if there's allocations -- if so, we may want to do this check differently.

PR #832, opened 2021-08-03
Allow method registration if handler implements Register

prashantv:
Currently, channel/subchannel only allow Register if the type
of the handler is the inbuilt channelHandler or handlerMap types.

This is quite inflexible, e.g., as we add new handler types in #831,
we still want to support registration without extending the type checks.

Instead of looking for a specific type, we can make the logic cleaner
by looking for the Register method. This also makes the channel's
Register method, and the subchannel's Register method more consistent.

This change allows users to specify their own custom handlers that can
support registration (via the same Register method).


AllenLuUber (approved):

AllenLuUber:
I can confirm with change #831 is easier and a panicked test succeeded after rebasing on this.

channel.go
```diff
@@ -425,12 +425,14 @@ type Registrar interface {
...
 //
 // Register panics if the channel was constructed with an alternate root
-// handler.
+// handler that does not support Register.
 func (ch *Channel) Register(h Handler, methodName string) {
-	if _, ok := ch.handler.(channelHandler); !ok {
-		panic("can't register handler when channel configured with alternate root handler")
+	r, ok := ch.handler.(registrar)
+	if !ok {
+		panic("can't register handler when channel configured with alternate root handler without Register method")
```
cinchurge, line 432: seems like this will break existing callers - is this intended? do we want to make this optional rather than mandatory?

prashantv: This should be more permissive than before, previously we panicked if the `handler` wasn't of the type `channelHandler`, but now we check for an interface (that `channelHandler` implements), so any case that worked previously should also work now. Any specific scenario you had in mind that now breaks?

cinchurge: right - somehow i missed the added `Register()` method fo channelHandler. 😓  s'all good

cinchurge (approved):

PR #834, opened 2021-08-04
refactor callRes test frame construction

cinchurge:
Prefactor for #830 

I'd like to refactor the code for constructing `callRes` frames so that we can easily add more scenarios such as malformed args or different arg schemes. This is important since arg2 iteration is supported only for thrift calls, and we will be doing some error checking in `newLazyCallRes` which can be complicated to test with the current code.

### DIFF
--- a/relay_messages_test.go
+++ b/relay_messages_test.go
@@ -167,7 +167,7 @@ type testCallResParams struct {
 	flags    byte
 	code     byte
 	span     [25]byte
-	headers  []byte
+	headers  map[string]string
 	csumType byte
 	arg1     []byte
 	arg2     []byte
@@ -193,17 +193,10 @@ func (cr testCallRes) res(tb testing.TB) lazyCallRes {
 		params.code = 1
 	}
 	if cr&resHasHeaders != 0 {
-		kvMap := map[string]string{
+		params.headers = map[string]string{
 			"k1":      "v1",
 			"k222222": "",
 			"k3":      "thisisalonglongkey",
-		}
-		params.headers = append(params.headers, byte(len(kvMap)))
-		for k, v := range kvMap {
-			params.headers = append(params.headers, byte(len(k)))
-			params.headers = append(params.headers, []byte(k)...)
-			params.headers = append(params.headers, byte(len(v)))
-			params.headers = append(params.headers, []byte(v)...)
 		}
 	}
 	if cr&resHasChecksum != 0 {
@@ -223,10 +216,22 @@ func newCallResFrame(tb testing.TB, p testCallResParams) *Frame {
 	fh := fakeHeader(messageTypeCallRes)
 	payload := typed.NewWriteBuffer(f.Payload)
 
-	payload.WriteSingleByte(p.flags)                                          // flags
-	payload.WriteSingleByte(p.code)                                           // code
-	payload.WriteBytes(p.span[:])                                             // span
-	payload.WriteBytes(p.headers)                                             // headers
+	defer func() {
+		fh.SetPayloadSize(uint16(payload.BytesWritten()))
+		f.Header = fh
+		require.NoError(tb, fh.write(typed.NewWriteBuffer(f.headerBuffer)), "Failed to write header")
+	}()
+
+	payload.WriteSingleByte(p.flags)              // flags
+	payload.WriteSingleByte(p.code)               // code
+	payload.WriteBytes(p.span[:])                 // span
+	payload.WriteSingleByte(byte(len(p.headers))) // headers
+	for k, v := range p.headers {
+		payload.WriteSingleByte(byte(len(k)))
+		payload.WriteBytes([]byte(k))
+		payload.WriteSingleByte(byte(len(v)))
+		payload.WriteBytes([]byte(v))
+	}
 	payload.WriteSingleByte(p.csumType)                                       // checksum type
 	payload.WriteBytes(make([]byte, ChecksumType(p.csumType).ChecksumSize())) // dummy checksum (not used in tests)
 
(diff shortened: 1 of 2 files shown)


ablackmon (approved):

relay_messages_test.go
```diff
@@ -168,45 +186,34 @@ const (
...
 	}
-
-	headers := make(map[string]string)
 	if cr&resHasHeaders != 0 {
-		addRandomHeaders(headers)
+		kvList := []keyVal{
+			{"k1", "v1"},
+			{"k222222", ""},
+			{"k3", "thisisalonglongkey"},
+		}
```
prashantv, line 204: We do something different for call req transport headers (use a `map[string]string`, and then call `writeHeaders`), curious why we treat headers differently for call res.

Remember that these are the transport headers, and the headers returned from services will typically be in the arg2 for the response

relay_messages_test.go
```diff
@@ -215,6 +222,38 @@ func withLazyCallResCombinations(f func(cr testCallRes)) {
 	}
 }
 
+func newCallResFrame(tb testing.TB, p testCallResParams) *Frame {
```
prashantv, line 225: similar to above, the call-req functions are defined a little differently,
```
func (cr testCallReq) frameWithParams(t testing.TB, p testCallReqParams) *Frame {
```

Should we do the same thing for consistency?

relay_messages_test.go
```diff
@@ -215,6 +222,38 @@ func withLazyCallResCombinations(f func(cr testCallRes)) {
...
+	f := NewFrame(p.payloadSize)
+	f.Header = FrameHeader{
+		// size is set below after construction of payload
+		messageType: messageTypeCallRes,
+		ID:          0xDEADBEEF,
+	}
+	require.NoError(tb, f.Header.write(typed.NewWriteBuffer(f.headerBuffer)))
```
prashantv, line 232: We're writing the header here, but then we update the size in the header later. See how this is done for call req,
```
	// Set the size in the header and write out the header after we know the payload contents.
	defer func() {
		fh.size = FrameHeaderSize + uint16(len(f.Payload))
		f.Header = fh
		require.NoError(t, fh.write(typed.NewWriteBuffer(f.headerBuffer)), "failed to write header")
	}()
```

prashantv (commented):
I think we have an issue where the header's size is not written out correctly, it may be better to follow the same pattern as call req (or refactor both to look the same).

relay_messages_test.go
cinchurge: yea, i wanted to avoid the map so that the order of the entries is deterministic and easier for testing. but for the sake of consistency, I guess we should keep it the same.

relay_messages_test.go
cinchurge: i wanted to define this separately since the `.res()` calls are expected to return valid frames, while I'd also like to add tests for corrupted frames. I guess I could also add a `frameWithParams()` that wraps this, but it felt a bit redundant.

relay_messages_test.go
cinchurge: got it. will change for consistency

relay_messages_test.go
```diff
@@ -160,6 +160,20 @@ func withLazyCallReqCombinations(f func(cr testCallReq)) {
 
 type testCallRes int
 
+type testCallResParams struct {
+	payloadSize       int
```
prashantv, line 164: not really sure how `payloadSize` interacts with all the other fields (e.g., what happens if the length of arg1+2+3 > payloadSize).

`testCallReqParams` does not have a payload size, not sure if we're intentionally doing something different for call res construction vs call req construction

relay_messages_test.go
```diff
@@ -168,45 +182,34 @@ const (
...
+			"k222222": "",
+			"k3":      "thisisalonglongkey",
+		}
+		params.headers = append(params.headers, byte(len(kvMap)))
+		for k, v := range kvMap {
+			params.headers = append(params.headers, byte(len(k)))
+			params.headers = append(params.headers, []byte(k)...)
+			params.headers = append(params.headers, byte(len(v)))
+			params.headers = append(params.headers, []byte(v)...)
+		}
```
prashantv, line 207: Unlike the call-req methods, I think I understand the separation here:
 * `testCallRes.res(t)` allows building a frame with some "predefined" values based on the bits set
 * `newCallResFrame` allows building a call res frame with a specific set of parameters

The first uses the latter, and the latter is responsible for all the frame encoding. Given that, should `headers` be a `map[string]string`? that way all the encoding logic is in a single place

relay_messages_test.go
```diff
@@ -215,6 +218,35 @@ func withLazyCallResCombinations(f func(cr testCallRes)) {
...
+	}
+	for _, arg := range args {
+		payload.WriteUint16(uint16(len(arg)))
+		payload.WriteBytes(arg)
+	}
+
+	require.NoError(tb, payload.Err(), "Got unexpected error constructing callRes frame")
+	fh.SetPayloadSize(uint16(payload.BytesWritten()))
+	f.Header = fh
+	require.NoError(tb, fh.write(typed.NewWriteBuffer(f.headerBuffer)), "Failed to write header")
```
prashantv, line 245: optional: should we use a `defer` for this, similar to the call req code?

prashantv (approved):
The call res code is still a little different from the call req code, but I think this is clearer to follow (we probably should make call req work in the same way).

relay_messages_test.go
cinchurge: this is useful for creating corrupted frames that trigger a buffer read error, and is used in a test in the next diff.

prashantv: Perhaps it should be a `if payloadSize > 0` then write it out as the payload size in the header?

with the current implementation,
 - if payloadSize is too small, then I think the typed writer will start to fail (which we check with `require.NoError`
 - if payloadSize is too large, then it has no impact, since we overwrite the frame header's payload size with bytes written.

Perhaps I'm missing something, not sure how it'll work with the next diff

cinchurge: yea i suppose it's not really necessary - it would make the numbers smaller and easier to reason about, so we can trigger an error with e.g.
```
	_, err := newLazyCallRes(newCallResFrame(t, testCallResParams{
		payloadSize: 100,
		arg2Prefix:  []byte{0, 100},
		arg2KeyVals: exampleArg2Map,
	}))
```
but it's not strictly necessary. we should be able to achieve something similar with a larger number in `arg2Prefix`. will change in #830

PR #835, opened 2021-08-11
[tchannel-go] Add options to support handlers with ignoring methods

AllenLuUber:
Redo of #831 after rebasing.

If a user wants to migrate off tchannel-go server on a per method level, we need a way to allow users to specify: for those specified methods, tchannel-go should handle it as before. But for others, let passed-in handler handle it. This PR provides an additional option to allow such use case for gradual migration.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -128,6 +128,7 @@ type ChannelOptions struct {
 	// requests with specified methods can be ignored by the above passed-in handler
 	// and handled natively by TChannel.
 	// Requests with other methods will be handled by passed-in handler.
+	// Methods should be in the format of Service::Method.
 	// This is useful for the gradual migration purpose.
 	SkipHandlerMethods []string
 
@@ -914,7 +915,7 @@ func toServiceMethodSet(sms []string) (map[string]map[string]struct{}, error) {
 	for _, sm := range sms {
 		s := strings.Split(sm, "::")
 		if len(s) != 2 {
-			return nil, fmt.Errorf(`Each "SkipHandlerMethods" value should be of service::Method format but got %q`, sm)
+			return nil, fmt.Errorf("each %q value should be of service::Method format but got %q", "SkipHandlerMethods", sm)
 		}
 		svc, method := s[0], s[1]
 		if _, ok := set[svc]; !ok {
--- a/channel.go
+++ b/channel.go
@@ -907,7 +907,7 @@ func toStringSet(ss []string) map[string]struct{} {
 	return set
 }
 
-// take a lise of service::method formatted string and make
+// take a list of service::method formatted string and make
 // the map[service]map[method]struct{} set
 func toServiceMethodSet(sms []string) (map[string]map[string]struct{}, error) {
 	set := map[string]map[string]struct{}{}
--- a/channel.go
+++ b/channel.go
@@ -124,12 +124,12 @@ type ChannelOptions struct {
 	// default handler that delegates to a subchannel.
 	Handler Handler
 
-	// IgnoreMethods allow users to configure TChannel server such that
+	// SkipHandlerMethods allow users to configure TChannel server such that
 	// requests with specified methods can be ignored by the above passed-in handler
 	// and handled natively by TChannel.
 	// Requests with other methods will be handled by passed-in handler.
 	// This is useful for the gradual migration purpose.
-	IgnoreMethods []string
+	SkipHandlerMethods []string
 
 	// Dialer is optional factory method which can be used for overriding
 	// outbound connections for things like TLS handshake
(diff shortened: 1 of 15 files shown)


handlers.go
```diff
@@ -135,3 +136,31 @@ func (c channelHandler) Handle(ctx context.Context, call *InboundCall) {
...
+	userHandler       Handler
+}
+
+func (u userHandlerWithIgnore) Handle(ctx context.Context, call *InboundCall) {
+	var sb strings.Builder
+	sb.WriteString(call.ServiceName())
+	sb.WriteString("::")
+	sb.Write(call.Method())
+
+	if _, ok := u.ignoreUserHandler[sb.String()]; ok {
```
prashantv, line 157: see https://github.com/uber/tchannel-go/pull/831#discussion_r681340445

To avoid this, you could instead have `map[string]map[string]struct{}` (first map for service name, second for method name)

channel.go
```diff
@@ -124,6 +124,13 @@ type ChannelOptions struct {
 	// default handler that delegates to a subchannel.
 	Handler Handler
 
+	// IgnoreMethods allow users to configure TChannel server such that
+	// requests with specified methods can be ignored by the above passed-in handler
+	// and handled natively by TChannel.
+	// Requests with other methods will be handled by passed-in handler.
+	// This is useful for the gradual migration purpose.
+	IgnoreMethods []string
```
prashantv, line 132: `Ignore` may be a little unclear (is the handler ignoring these?) perhaps `SkipHandlerMethods []string` ?

handlers_test.go
```diff
@@ -57,3 +59,56 @@ func TestHandlers(t *testing.T) {
...
+		IgnoreMethods: []string{procedure(svc, userHandleIgnoreMethod)},
+	}
+	ch, err := NewChannel(svc, opts)
+	require.NoError(t, err, "error creating a TChannel channel")
+
+	// channel should be able to handle user ignored methods
+	ch.Register(recorderHandler(channelCounter), userHandleIgnoreMethod)
+
+	call, ignoreCall := makeInboundCall(svc, userHandleMethod, NullLogger), makeInboundCall(svc, userHandleIgnoreMethod, NullLogger)
+
```
prashantv, line 101: make a real call using the tchannel call using a client, avoid creating partial objects like connection / inbound call

channel.go
```diff
@@ -304,9 +311,20 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	}
 	ch.peers = newRootPeerList(ch, opts.OnPeerStatusChanged).newChild()
 
-	if opts.Handler != nil {
+	switch {
+	case len(opts.SkipHandlerMethods) > 0 && opts.Handler != nil:
+		sm, err := toServiceMethodSet(opts.SkipHandlerMethods)
+		if err != nil {
+			return nil, err
```
prashantv, line 318: can you add coverage for this by trying to construct a channel with an invalid method in `SkipHandlerMethods`?

channel.go
```diff
@@ -124,6 +124,13 @@ type ChannelOptions struct {
 	// default handler that delegates to a subchannel.
 	Handler Handler
 
+	// SkipHandlerMethods allow users to configure TChannel server such that
+	// requests with specified methods can be ignored by the above passed-in handler
+	// and handled natively by TChannel.
+	// Requests with other methods will be handled by passed-in handler.
+	// This is useful for the gradual migration purpose.
+	SkipHandlerMethods []string
```
prashantv, line 132: mention that methods should be in the format `Service::Method`

channel.go
```diff
@@ -888,3 +906,21 @@ func toStringSet(ss []string) map[string]struct{} {
...
 }
+
+// take a list of service::method formatted string and make
+// the map[service]map[method]struct{} set
+func toServiceMethodSet(sms []string) (map[string]map[string]struct{}, error) {
+	set := map[string]map[string]struct{}{}
+	for _, sm := range sms {
+		s := strings.Split(sm, "::")
+		if len(s) != 2 {
+			return nil, fmt.Errorf(`Each "SkipHandlerMethods" value should be of service::Method format but got %q`, sm)
```
prashantv, line 917: nit: error should be lowercase (since they're often concatenated), and use `%q` for quoting.

handlers_with_skip_test.go
```diff
@@ -0,0 +1,87 @@
...
+			ctx, cancel := tchannel.NewContext(testutils.Timeout(300 * time.Millisecond))
+			defer cancel()
+			raw.Call(ctx, client, ts.HostPort(), svc, userHandleSkipMethod, nil, nil)
+		}
+		assert.Equal(t, uint32(handleSkipRuns), channelCounter.c.Load(), "user provided handler not invoked correct amount of times")
+	})
+}
+
+type recorderHandler struct {
+	c *atomic.Uint32
```
prashantv, line 82: instead of a pointer here, suggest:
 * using `c atomic.Uint32` here
 * and then doing `&recordHandler{}` instead (no need to create the atomic counter)

prashantv (approved):
LGTM with some minor comments

PR #841, opened 2021-11-09
Remove service name map in ignore user handler 

jronak:
`IgnoreUserHandler` works on the primary channel where it handles methods belonging to the root service name. Also, the current ignore-user-handler derives the Tchannel service name from the method name, but the method name contains thrift service name which may be different, for example, tchannel service name: `echo` and method name: `Echo::echohi`. To avoid such cases, we can drop the service name map and use the procedure map.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -909,7 +909,7 @@ func toStringSet(ss []string) map[string]struct{} {
 }
 
 // take a list of service::method formatted string and make
-// the map[service]map[service::method]struct{} set
+// the map[service::method]struct{} set
 func toServiceMethodSet(sms []string) (map[string]struct{}, error) {
 	set := map[string]struct{}{}
 	for _, sm := range sms {
--- a/channel.go
+++ b/channel.go
@@ -913,8 +913,7 @@ func toStringSet(ss []string) map[string]struct{} {
 func toServiceMethodSet(sms []string) (map[string]struct{}, error) {
 	set := map[string]struct{}{}
 	for _, sm := range sms {
-		s := strings.Split(sm, "::")
-		if len(s) != 2 {
+		if len(strings.Split(sm, "::")) != 2 {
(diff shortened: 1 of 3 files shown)


channel.go
```diff
@@ -909,19 +909,15 @@ func toStringSet(ss []string) map[string]struct{} {
...
 		s := strings.Split(sm, "::")
 		if len(s) != 2 {
 			return nil, fmt.Errorf("each %q value should be of service::Method format but got %q", "SkipHandlerMethods", sm)
 		}
-		svc := s[0]
-		if _, ok := set[svc]; !ok {
-			set[svc] = map[string]struct{}{}
-		}
-		set[svc][sm] = struct{}{}
+		set[sm] = struct{}{}
```
AllenLuUber, line 920: nit we can `if s:= .. ; len(s) !=2` here now :)

AllenLuUber (approved):

channel.go
bryndin: (optional suggestion) Do we actually care about `sm` validation here, since `svc := s[0]` is out?

Neither `service` or `method` name normalization, nor validation of service/method existence seem to be happening for `sm`.  If they don't have to be correct, why even check for `::` format?

jronak: @bryndin, I would prefer to keep the check to catch incorrectly configured method names, where the method name must always contain the thrift service prefix.

AllenLuUber: yea I agree with @jronak. This is the bare minimal validation we can do to make sure it is X::X format.

AllenLuUber (approved):

channel.go
bryndin: Agree.
Customers asked us to add full validation on service/method spelling, they had outages due to accidental typos. If we ever do that, we'd refactor this code.

PR #843, opened 2021-12-15
remove rpc-service response header to not propagate it upstream

kishoresenji:
removing $rpc$-header from the response headers of a tchannel outbound call to be propagated upstream. this is useful to avoid rpc failures in yarpc framework when one of the intermediaries is a tchannel inbound service. More detailed explanation of the scenario is added as a comment on the client.go.


thrift/client.go
```diff
@@ -155,6 +157,37 @@ func (c *client) Call(ctx Context, thriftService, methodName string, req, resp t
...
+// We have observed this issue while migrating to yarpc from tchannel inbound. For example, let us
+// assume that both C & D were tchannel inbounds earlier and one of them wants to migrate to yarpc
+// and at which point the $rpc$-service header flows upstream and assuming A is already using yarpc outbound
+// the call will fail. Although it is a backward incompatible change, by removing the $rpc$-service header
+// prevents A from seeing the $rpc$-header and marking the rpc as a failure. The alternative is to make
+// B use Child() context while making downstream calls and the net effect is the same and while migrating to yarpc,
+// it is unreasonable to expect the callee service owners to perform changes into the upstream callers
+// to create Child() context especially when the call graph is complex and multiple levels deep. This is
+// a reasonbable backward incompatible change to make as we do not expect any caller to actually depend on
+// or find the $rpc$-service header from a descendant service to be actually useful.
```
rabbbit, line 189: Could you explain what you mean by "actually useful"?

It sounds like $rpc$-service was added there for a reason, and it's working as expected?

thrift/client.go
```diff
@@ -155,6 +157,37 @@ func (c *client) Call(ctx Context, thriftService, methodName string, req, resp t
...
+//
+// We have observed this issue while migrating to yarpc from tchannel inbound. For example, let us
+// assume that both C & D were tchannel inbounds earlier and one of them wants to migrate to yarpc
+// and at which point the $rpc$-service header flows upstream and assuming A is already using yarpc outbound
+// the call will fail. Although it is a backward incompatible change, by removing the $rpc$-service header
+// prevents A from seeing the $rpc$-header and marking the rpc as a failure. The alternative is to make
+// B use Child() context while making downstream calls and the net effect is the same and while migrating to yarpc,
+// it is unreasonable to expect the callee service owners to perform changes into the upstream callers
+// to create Child() context especially when the call graph is complex and multiple levels deep. This is
+// a reasonbable backward incompatible change to make as we do not expect any caller to actually depend on
```
rabbbit, line 188: Is $rpc$-service response header visible to the caller? If so, how do we know that nobody depends on it?

thrift/client.go
kishoresenji: $rpc$-service header is not added by tchannel inbound. It is added by YARPC inbound.

Sample topology: 
```
A (YARPC inbound/YARPC outbound) -> B (tchannel inbound/tchannel outbound) -> C (tchannel inbound)
                                                                          |-> D (tchannel inbound)
```

A makes a yarpc outbound call to B which makes two calls to C & D in parallel. None of C, D & B add $rpc$-service response header and so A will not receive one and so there is no validation from A. 

Now let's say C tries to upgrade to YARPC inbound. It starts emitting $rpc$-service response header.  Because tchannel outbound client in B sets the response headers that it receives from the outbound into its own context. The $rpc$-service header will propagate to A through B. A will see the $rpc$-service response header and finds that the response header is from C and not B as it expects and fails its rpc inbound with an error 

` service name sent from the request does not match the service name received in the response`

another wrinkle: As C & D are made in parallel by B, the response headers from C & D overwrite each other in B's context.

This is what I mean by the "$rpc$-service header from a descendant service" is not useful because it actually causes harm as A would inadvertently think that C is handling its request. 

The recommended solution from tchannel is to use Child() context while making outbound calls. ie. B should use Child() for all the outbound calls. The effect of that is that the downstream response headers are cutoff from being relayed upstream. What this means though is that while upgrading to YARPC, the service needs to find all its callers and modify all the codepaths to use a Child().

What this PR does is to make that process easier for service owner. As we see propagating $rpc$-service upstream from a tchannel outbound call causes unnecessary harm and using Child() anyway removes those headers from propagating upstream, we drop the $rpc$-service header from propagating upstream even when Child() is not used. It is also quite unreasonable to put that burden on the service owner to find out all its callers and modify the caller code. 

This change is only for tchannel outbound calls (to help us prevent outages). The validation in a yarpc only topology is still preserved.  ie A -> B where both A and B are YARPC outbound and inbound respectively, A will get a $rpc$-service response header and will validate.

thrift/client.go
kishoresenji: Sample topology:

```
A (YARPC inbound/YARPC outbound) -> B (YARPC inbound/YARPC outbound) -> C (YARPC inbound)
                                                                    |-> D (YARPC inbound)
```
Let us say B makes C & D in parallel. C & D emit $rpc$-service as C & D respectively. However B emits $rpc$-service response header (as YARPC does not propagate response headers from downstream) as B. 

A receives the $rpc$-service response header and A validates that it to be B. All A wants to ensure is the immediate callee is B. A does not depend or want to know who B depends on to serve that request. 

If we compare that to a sample topology with tchannel as an intermediary or a routing node.

```
A (YARPC inbound/YARPC outbound) -> B (tchannel inbound/tchannel outbound) -> C (YARPC inbound)
                                                                          |-> D (tchannel inbound)
```

C sends the $rpc$-service response header as C, which B relays upstream to A conveying to A that the service which handled the request is C. This is exactly what I mean by A does not depend on a "$rpc$-service header from a descendant service". A depends on the $rpc$-service header from its immediate callee, not a descendant one as propagating descendant headers actually causes harm.

thrift/client.go
rabbbit: Ah, I misunderstood. My (apparently wrong) impression was that yarpc would be re-using tchannel-go, and so this PR would effectively disable `$rpc$-service` for yarpc as well. 

If this is not the case, I was wrong - sorry for the extra noise.

thrift/client.go
rabbbit: Let's talk on the comment above.

thrift/client.go
```diff
@@ -155,6 +157,37 @@ func (c *client) Call(ctx Context, thriftService, methodName string, req, resp t
...
+// or find the $rpc$-service header from a descendant service to be actually useful.
+func removeRPCServiceHeader(headers map[string]string) map[string]string {
```
rabbbit, line 190: separately, can/should we call this `removeSingleHopHeaders`, or is it too early? I imagine there's more of these - one example I can think of already is `load`. 

Also, perhaps unrelated to this change, I believe we hit a very similar issuer with service auth, where a tchannel service was incorrectly copying inbound single-hop tokens to their outbounds.  Similar, but looks like a different code path?

thrift/client.go
kishoresenji: I verified that yarpc does not use thrift.Client to make the outbound call. It uses tchannel [directly](https://github.com/yarpc/yarpc-go/blob/dev/transport/tchannel/channel_outbound.go#L174) and reimplemented most of the thrift.Client logic. 

If we want to be absolutely safe, we can actually make a copy of the map instead of directly modifying it. So what goes into the thrift.Context is a copy of the headers without the $rpc$-service.

rabbbit (approved):
This makes me uncomfortable, for two reasons:
- we're tieing in yarpc behaviour with tchannel-client, where they shouldn't know about each other
- it is a breaking change

However, it seems like the least bad of the options:
- I don't think anyone outside of us would be using both tchannel-go and yarpc (I hope they don't)
- while it's plausible that some tchannel-go user relies on $rpc$-service for some kind of application-level shadowing/caller differentiation, it seems unlikely. @kishoresenji executed an internal scan for this.

AllenLuUber (approved):

thrift/client.go
cinchurge: +1. tchannel sits a layer below yarpc, so coupling in yarpc semantics (e.g. `$rpc$-` headers) breaks the abstraction. prefer to implement this as a configurable list of headers that should be stripped by the client, rather than hard-coding specific headers. since tchannel upgrades are a real pain in the monorepo, this can also help alleviate unnecessary overhead if other headers also need to be stripped.

cinchurge (changes requested):

PR #845, opened 2022-01-21
Fix frame leaks in handleNonCallReq

cinchurge:
Recently it was discovered that a number of code paths in the relayer would result in frame leaks, where after a relay decision (either send to `Connection.writeFrames()` for sending or dropped) the frame isn't returned to the frame pool, resulting in a memory leak.

With the previous PRs #847 and #852, we are now in a position to address the leaks in each code path. This change fixes the leaks in `handleNonCallReq()` as detected by the existing tests.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -209,6 +209,8 @@ const (
 	responseFrame frameType = 1
 )
 
+var _ frameReceiver = (*Relayer)(nil)
+
 // A Relayer forwards frames.
 type Relayer struct {
 	relayHost      RelayHost
@@ -266,7 +268,7 @@ func NewRelayer(ch *Channel, conn *Connection) *Relayer {
 // Relay is called for each frame that is read on the connection.
 func (r *Relayer) Relay(f *Frame) (shouldRelease bool, _ error) {
 	if f.messageType() != messageTypeCallReq {
-		err := r.handleNonCallReq(f)
+		shouldRelease, err := r.handleNonCallReq(f)
 		if err == errUnknownID {
 			// This ID may be owned by an outgoing call, so check the outbound
 			// message exchange, and if it succeeds, then the frame has been
@@ -275,7 +277,7 @@ func (r *Relayer) Relay(f *Frame) (shouldRelease bool, _ error) {
 				return _relayNoRelease, nil
 			}
 		}
-		return _relayNoRelease, err
+		return shouldRelease, err
 	}
 
 	cr, err := newLazyCallReq(f)
@@ -530,7 +532,7 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) (shouldRelease bool, _ error) {
 }
 
 // Handle all frames except messageTypeCallReq.
-func (r *Relayer) handleNonCallReq(f *Frame) error {
+func (r *Relayer) handleNonCallReq(f *Frame) (shouldRelease bool, _ error) {
 	frameType := frameTypeFor(f)
 	finished := finishesCall(f)
 
@@ -544,12 +546,12 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 	// Stop the timeout if the call if finished.
 	item, stopped, ok := items.Get(f.Header.ID, finished /* stopTimeout */)
 	if !ok {
-		return errUnknownID
+		return _relayShouldRelease, errUnknownID
 	}
 	if item.tomb || (finished && !stopped) {
 		// Item has previously timed out, or is in the process of timing out.
 		// TODO: metrics for late-arriving frames.
-		return nil
+		return _relayShouldRelease, nil
 	}
 
 	switch f.messageType() {
@@ -582,13 +584,13 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 	sent, failure := item.destination.Receive(f, frameType)
 	if !sent {
 		r.failRelayItem(items, originalID, failure, errFrameNotSent)
-		return nil
+		return _relayShouldRelease, nil
 	}
 
 	if finished {
 		r.finishRelayItem(items, originalID)
 	}
-	return nil
+	return !sent, nil
 }
 
 // addRelayItem adds a relay item to either outbound or inbound.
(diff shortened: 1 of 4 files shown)


relay_test.go
```diff
@@ -857,6 +861,11 @@ func TestRelayStalledConnection(t *testing.T) {
 			Succeeded().End()
 		ts.AssertRelayStats(calls)
 	})
+
+	if unreleasedCount, isEmpty := pool.CheckEmpty(); isEmpty != "" || unreleasedCount > 0 {
+		t.Errorf("Frame pool has %v unreleased frames, errors:\n%v\n",
+			unreleasedCount, isEmpty)
+	}
```
rabbbit, line 868: Totally new to this code-base but - is there a reason not to do this very check for more tests in this file?

relay.go
```diff
@@ -544,12 +549,12 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 	// Stop the timeout if the call if finished.
 	item, stopped, ok := items.Get(f.Header.ID, finished /* stopTimeout */)
 	if !ok {
-		return errUnknownID
+		return _frameNotDropped, errUnknownID
```
prashantv, line 552: this error case looks like the frame isn't forwarded, and so it is being dropped, shouldn't this be `_frameDropped`?

relay.go
```diff
@@ -544,12 +549,12 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
...
 	}
 	if item.tomb || (finished && !stopped) {
 		// Item has previously timed out, or is in the process of timing out.
 		// TODO: metrics for late-arriving frames.
-		return nil
+		return !finished, nil
```
prashantv, line 557: similar to above, if the frame isn't being forwarded, so it is being dropped right?

relay.go
```diff
@@ -579,16 +584,16 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 	originalID := f.Header.ID
 	f.Header.ID = item.remapID
 
-	sent, failure := item.destination.Receive(f, frameType)
+	sent, failure, dropped := item.destination.Receive(f, frameType)
```
prashantv, line 587: Do we need a third return? A couple of alternative thoughts:
1. Isn't `sent` a sign of whether the frame is handled by the receiver, and so indicates we didn't drop it. is `!sent` == dropped already?
2. Another option is to treat `Receive` as transfer of ownership, regardless of success or failure. So the frame belongs to `item.destination`. In the case where the channel has space, it ends up going to `writeFrames` which then puts it back in the frame pool. However, in the error case, we could call `c.opts.FramePool.Release(frame)`

(2) is simple, but adds one extra caller to release frames outside of read/write frames (it's a little cleaner + easier to follow to have limited interactions with pooling), but it might still be better than changing the Receive API.

Maybe (1) is the best option?

relay_test.go
prashantv: It requires using the RecordingFramePool which doesn't actually pool, and is significantly slower + has a lock which ends up serializing the relay. This causes tests to behave differently (less like production), so it's only used in some cases where we're ensuring that we're pooling correctly + not reusing released frames. There's also tests that don't release frames (rely on GC instead) for the error cases where plumbing through the "drop" added more complexity and didn't seem worth it at the time (since some of these error cases shouldn't happen much).

Ideally, we'd run all relay tests with the normal frame pool + with the recording frame pool. The TestServer test helpers make it easy to run the same test multiple times with different options, so this could be worth doing, so there's both prod-like coverage, as well as ensuring that pooling works correctly. If there's still cases where the frame is intentionally not released, there can be another option to indicate skipping the "is empty" check.

cinchurge:
Given the feedback, I'm taking a different approach of modifying `WithTestServer` to optionally rerun the same test with the frame leak checks. Can we review #847 first?

cinchurge:
Added #852 as a prefactor for this change

cinchurge (commented):
updated

relay.go
```diff
@@ -307,7 +309,7 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
 	if item.tomb || (finished && !stopped) {
 		// Item has previously timed out, or is in the process of timing out.
 		// TODO: metrics for late-arriving frames.
-		return true, ""
+		return false, ""
```
prashantv, line 312: should a `failureReason` be returned here? the current caller of `Receive` uses the failure reason if `!sent`, so I think it's worth double checking how this path should work (what metrics should be emitted, etc)

relay.go
```diff
@@ -323,6 +325,7 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
 	}
 	select {
 	case r.conn.sendCh <- f:
+		sent = true
```
prashantv, line 328: do you need this variable? the `default` case returns `false`, so the only way to get to the `return sent` is if it goes through this path, so seems like line 369 could just be `return true`?

relay.go
```diff
@@ -582,13 +585,13 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
 	sent, failure := item.destination.Receive(f, frameType)
 	if !sent {
 		r.failRelayItem(items, originalID, failure, errFrameNotSent)
```
prashantv, line 587: check whether `failRelayItem` should be called in the cases that previously returned `sent=true`, and now are `sent=false`. You could rely on the `failure` being non-empty instead if you want to keep the calls behave the same (and keep using an empty error) though that is a little subtle

cinchurge: with the changes in `Receive()` reverted, I don't think we have any more cases where the value of `sent` has changed, so I think we should be ok for now?

relay.go
```diff
@@ -582,13 +584,13 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
...
-		return nil
+		return _relayShouldRelease, nil
 	}
 
 	if finished {
 		r.finishRelayItem(items, originalID)
 	}
-	return nil
+	return !sent, nil
```
prashantv, line 593: we handle the `!sent` case earlier, so this must be false at this point, so can simplify and make it explicit.

if we got to this point, the frame was sent to the other side, so no need to release.

```suggestion
	return _relayNoRelease, nil
```

prashantv (approved):
Nice, this is nice and simple now!

rabbbit (approved):
Quick stamp to unblock, will re-read later.

PR #846, opened 2022-01-22
Migrate to go mod

cinchurge:
go mod is the official package manager for go, so switching over from glide.

Tested by running:
```
git clean -fdx
make install_lint
make lint
make all
```
Additionally, compared the test output between `HEAD^` and this change, and found no meaningful deviation.


Makefile
```diff
@@ -40,31 +37,18 @@ setup:
...
 	mkdir -p $(THRIFT_GEN_RELEASE_DARWIN)
 
-# We want to remove `vendor` dir because thrift-gen tests don't work with it.
-# However, glide install even with --cache-gopath option leaves GOPATH at HEAD,
-# not at the desired versions from glide.lock, which are only applied to `vendor`
-# dir. So we move `vendor` to a temp dir and prepend it to GOPATH.
-# Note that glide itself is still executed against the original GOPATH.
 install:
-	GOPATH=$(OLD_GOPATH) glide --debug install --cache --cache-gopath
+	GOPATH=$(OLD_GOPATH) go mod vendor
```
prashantv, line 41: can the `GOPATH` and `OLD_GOPATH` code also be removed?

go.mod
```diff
@@ -0,0 +1,25 @@
+module github.com/uber/tchannel-go
+
+go 1.16
+
+require (
+	github.com/HdrHistogram/hdrhistogram-go v1.1.2 // indirect
+	github.com/apache/thrift v0.0.0-20161221203622-b2a4d4ae21c7
```
prashantv, line 7: Can you use the 0.9.3 version here? (Not sure if the apache tag works with go mod, I remember having issues in the past,
https://issues.apache.org/jira/browse/THRIFT-5205
Didn't end up following up though).

Either way, I'd suggest adding a comment with the specific constraint that previously was in glide.yaml and add comments calling out that it won't work with newer versions (and why not, etc).

cinchurge: pinned to the same sha as glide.lock, was there some other reason why we can't be on a newer version?

prashantv: https://issues.apache.org/jira/browse/THRIFT-5205 is why we can't use a v0.9 in the version for the dep

the reason we don't support newer versions is that Thrift changed the interface to add context to a bunch of interfaces, which then requires breaking changes to tchannel-go. breaking change + major version change resulted in diamond dependency problems (service A wants v1 of tchannel, but a library it depends on wants v2). There wasn't a feasible upgrade path before, but maybe it's more feasible now (if there's only a few repos depending on tchannel)

prashantv (approved):
LGTM

go.mod
cinchurge: ah, gotcha. thanks for the explanation! I'll look into the upgrade issue separately.

mway (approved):

PR #847, opened 2022-01-24
Add option to check frame leaks after each test with WtihTestServer

cinchurge:
With the realization that some code paths aren't properly releasing frames
(e.g. #845), I'd like to add an option to `WithTestServer` to optionally check
for frame leaks along with each test.

As `RecordingFramePool` isn't an actual frame pool, we can't simply replace
the frame pool in all of our tests - we'd miss signals on how the code
is actually run in production. As such, we'll need to run the same test again
with `RecordingFramePool`, which will more than double the time it takes
to run a test.

To avoid an outsized impact to test duration, make the option opt-in so that
we can selectively onboard tests to this check.

### DIFF
--- a/checked_frame_pool.go
+++ b/checked_frame_pool.go
@@ -1,0 +1,97 @@
+package tchannel
+
+import (
+	"fmt"
+	"runtime"
+	"sync"
+)
+
+// CheckedFramePoolForTest tracks gets and releases of frames, verifies that
+// frames aren't double released, and can be used to check for frame leaks.
+// As such, it is not performant, nor is it even a proper frame pool.
+//
+// It is intended to be used ONLY in tests.
+type CheckedFramePoolForTest struct {
+	mu sync.Mutex
+
+	allocations map[*Frame]string
+	badRelease  []string
+}
+
+// NewCheckedFramePoolForTest initializes a new CheckedFramePoolForTest.
+func NewCheckedFramePoolForTest() *CheckedFramePoolForTest {
+	return &CheckedFramePoolForTest{
+		allocations: make(map[*Frame]string),
+	}
+}
+
+// Get implements FramePool
+func (p *CheckedFramePoolForTest) Get() *Frame {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	frame := NewFrame(MaxFramePayloadSize)
+	p.allocations[frame] = recordStack()
+	return frame
+}
+
+// Release implements FramePool
+func (p *CheckedFramePoolForTest) Release(f *Frame) {
+	// Make sure the payload is not used after this point by clearing the frame.
+	zeroOut(f.Payload)
+	f.Payload = nil
+	zeroOut(f.buffer)
+	f.buffer = nil
+	zeroOut(f.headerBuffer)
+	f.headerBuffer = nil
+	f.Header = FrameHeader{}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if _, ok := p.allocations[f]; !ok {
+		p.badRelease = append(p.badRelease, "bad Release at "+recordStack())
+		return
+	}
+
+	delete(p.allocations, f)
+}
+
+// CheckedFramePoolForTestResult contains info on mismatched gets/releases
+type CheckedFramePoolForTestResult struct {
+	BadReleases []string
+	Unreleased  []string
+}
+
+// HasIssues indicates whether there were any issues with gets/releases
+func (r CheckedFramePoolForTestResult) HasIssues() bool {
+	return len(r.BadReleases)+len(r.Unreleased) > 0
+}
+
+// CheckEmpty returns the number of unreleased frames in the pool
+func (p *CheckedFramePoolForTest) CheckEmpty() CheckedFramePoolForTestResult {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	var badCalls []string
+	for f, s := range p.allocations {
+		badCalls = append(badCalls, fmt.Sprintf("frame %p: %v not released, get from: %v", f, f.Header, s))
+	}
+
+	return CheckedFramePoolForTestResult{
+		Unreleased:  badCalls,
+		BadReleases: p.badRelease,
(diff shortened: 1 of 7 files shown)


frame_utils.go
```diff
@@ -0,0 +1,75 @@
+package tchannel
```
prashantv, line 1: nit: since this only contains a single type, suggest naming the file after the type (e.g., `recording_frame_pool.go` or `checked_frame_pool.go` if the type is renamed as suggested below

frame_utils.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+	"sync"
+)
+
+type RecordingFramePool struct {
+	sync.Mutex
```
prashantv, line 11: https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs

especially don't want to embed a mutex

frame_utils.go
```diff
@@ -0,0 +1,75 @@
+package tchannel
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+	"sync"
+)
+
+type RecordingFramePool struct {
```
prashantv, line 10: (I think this is the reason why I skipped this in the past -- was trying to avoid exporting this type)

Missing godoc, can you add comments for all exported symbols, and also explicitly call out that it's only intended for tests.

frame_utils.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if _, ok := p.allocations[f]; !ok {
+		p.badRelease = append(p.badRelease, "bad Release at "+recordStack())
+		return
+	}
+
+	delete(p.allocations, f)
+}
+
+func (p *RecordingFramePool) CheckEmpty() (int, string) {
```
prashantv, line 53: what do you think about returning a struct here that separates into:
 - number of unreleased frames
 - bad allocation messages
 - bad releases

or can combine the last 2, and just do `(unreleasedFrames int, badPooling []string)`

frame_utils.go
```diff
@@ -0,0 +1,75 @@
...
+	badCalls = append(badCalls, p.badRelease...)
+	for f, s := range p.allocations {
+		badCalls = append(badCalls, fmt.Sprintf("frame %p: %v not released, get from: %v", f, f.Header, s))
+	}
+	return len(p.allocations), strings.Join(badCalls, "\n")
+}
+
+func recordStack() string {
+	buf := make([]byte, 4096)
+	runtime.Stack(buf, false)
```
prashantv, line 67:
```suggestion
	runtime.Stack(buf, false /* all */)
```

testutils/channel_opts.go
```diff
@@ -66,6 +66,12 @@ type ChannelOpts struct {
 	// negative values are treated as a single run.
 	RunCount int
 
+	// CheckFrameLeaks indicates whether we should check for frame leaks or not.
+	// This causes the same tests to be run twice, first with the default frame pool,
+	// then with the recording frame pool, which will double the amount of time it takes
+	// for the test.
+	CheckFrameLeaks bool
```
prashantv, line 73: nit: this isn't just checking for frame leaks, but whether the frame pool is being used correctly (e.g., detects multiple release), so maybe `CheckFramePooling`?

relay_test.go
```diff
@@ -398,8 +398,7 @@ func TestLargeTimeoutsAreClamped(t *testing.T) {
 // TestRelayConcurrentCalls makes many concurrent calls and ensures that
 // we don't try to reuse any frames once they've been released.
 func TestRelayConcurrentCalls(t *testing.T) {
-	pool := NewProtectMemFramePool()
-	opts := testutils.NewOpts().SetRelayOnly().SetFramePool(pool)
+	opts := testutils.NewOpts().SetRelayOnly().SetCheckFrameLeaks()
```
prashantv, line 401: optional: what do you think about enabling the recording frame pool for all tests that are relay-only?

frame_utils.go
prashantv, line 10: nit: since this is now exported, suggest naming this more accurately, what do you think about `CheckedFramePool`?

frame_utils.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type RecordingFramePool struct {
+	sync.Mutex
+
+	allocations map[*Frame]string
+	badRelease  []string
+}
+
+func NewRecordingFramePool() *RecordingFramePool {
```
prashantv, line 17: nit: suggest adding `ForTest` suffix to the method name so it's very clear this is only exported for tests (definitely unsafe for prod)

relay_test.go
cinchurge: i think we should, but there will likely be leaks that need to be fixed along with that, so prefer to defer that to the next change

checked_frame_pool.go
```diff
@@ -0,0 +1,97 @@
...
+		return
+	}
+
+	delete(p.allocations, f)
+}
+
+// CheckedFramePoolForTestResult contains info on mismatched gets/releases
+type CheckedFramePoolForTestResult struct {
+	BadReleases           []string
+	UnexpectedAllocations []string
```
prashantv, line 63: I don't think these are unexpected allocs, rather unreleased. Maybe just `Unreleased []string`

checked_frame_pool.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// HasIssues indicates whether there were any issues with gets/releases
+func (r CheckedFramePoolForTestResult) HasIssues() bool {
+	return len(r.BadReleases) != 0 || len(r.UnexpectedAllocations) != 0
```
prashantv, line 68: nit: suggest `> 0` instead of `!= 0` since length is positive. oprionally: `len(r.BadReleases) + len(r.UnexpectedAllocatoins) > 0`

prashantv (approved):

checked_frame_pool.go
cinchurge: done

rabbbit (approved):

PR #848, opened 2022-01-25
Fix race in TestRelayConnection

cinchurge:
In TestRelayConnection, the final test is for the relay to create an inbound connection to a listening Hyperbhan service and verify that a call originating from the HB service uses that connection rather than creating a new outbound connection. The test, however, suffered from a race condition where the call could be made before the inbound handshake completed, resulting in an outbound being created by the HB service.

Add a verification step before `CallEcho` to ensure that the connection is established before the call is made.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -638,6 +638,11 @@ func TestRelayConnection(t *testing.T) {
 		_, err = ts.Relay().Connect(ctx, listeningHBSvc.PeerInfo().HostPort)
 		require.NoError(t, err, "Failed to connect from relay to listening host:port")
 
+		// Wait for inbound connection to be established before making call
+		if !waitForInboundConnection(listeningHBSvc, ts.Relay()) {
+			require.Fail(t, "no inbound connections established from relay to listeningHBSvc")
+		}
+
 		// Now when listeningHBSvc makes a call, it should use the above connection.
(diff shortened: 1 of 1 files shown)


relay_test.go
```diff
@@ -2058,3 +2063,21 @@ func copyHeaders(m map[string]string) map[string]string {
 	}
 	return copied
 }
+
+func waitForInboundConnection(listeningCh, callingCh *Channel) bool {
```
rabbbit, line 2067: I see `testutils.WaitFor` - should we use that?

cinchurge: good call. thanks!

relay_test.go
```diff
@@ -2058,3 +2065,11 @@ func copyHeaders(m map[string]string) map[string]string {
 	}
 	return copied
 }
+
+func getInbounds(listeningCh, callingCh *Channel) int {
+	callingPeer, ok := listeningCh.IntrospectState(nil).RootPeers[callingCh.PeerInfo().HostPort]
```
rabbbit, line 2070: `IntrospectState(nil /*opts*/)` ?

rabbbit (approved):

prashantv (commented):
Nice investigation on the race!

I've run into this before, and have a simpler fix, use `Ping` to ensure the connection is added to both sides
https://github.com/uber/tchannel-go/pull/851

(That also matches previous fixes better).

cinchurge:
Abandoning this in favor of #851

PR #849, opened 2022-01-25
Fix CI testing

cinchurge:
Tests have been failing in CI due to several flaky tests: TestWriteArg3AfterTimeout, TestRelayCorruptedCallResFrame, TestRelayConnection (being addressed in #848). Aside from TestRelayConnection, I have not been able to reproduce the flaky tests, so disabling in CI so we're able to catch other non-flaky failures.

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"math"
 	"net"
+	"os"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


rabbbit (approved):
A bit scary to start by disabling things, but okay.

I would have also really preferred to have the `lint` fix in a separate diff.

cinchurge:
> A bit scary to start by disabling things, but okay.
> 
> I would have also really preferred to have the `lint` fix in a separate diff.

Split off lint change to #850

PR #850, opened 2022-01-25
Update linting in Makefile

cinchurge:
Addressing an issue and adding some improvements in the Makefile:
- Add `install` as a dependency to `lint` so that `vendor` is updated before linting
- Add `lint` as a dependency to `all` so that invoking `make` would also do linting
- Call `go install` for installing the `golint` binary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ TEST_PORT=0
 
 -include crossdock/rules.mk
 
-all: test examples
+all: lint test examples
 
 $(BIN)/thrift:
 	mkdir -p $(BIN)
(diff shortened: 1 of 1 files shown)


rabbbit (approved):

Makefile
```diff
@@ -40,7 +40,7 @@ install:
 
 install_lint:
 	@echo "Installing golint, since we expect to lint"
-	go get -u -f golang.org/x/lint/golint
+	go install golang.org/x/lint/golint@latest
```
prashantv, line 43: If you do `go install <pkg>`, it'll use a version specified in the go.mod, which allows the tools to be version-managed like other deps (more consistent experience that doesn't depend on the host env).

I'd suggest making the tools managed more like zap,
 * Versioned, but separate module (so deps for the tools don't affect deps for prod)
 * Easy to install to a location like ./bin (see `GOBIN`)
 * More consistency (think atomic is the same).

cinchurge: great suggestion. i'll look into that. thanks!

PR #440, opened 2016-07-15
Fix flaky TestPeerScoreOnNewConnection

prashantv:
Test was failing with "Failed to add connection to peer" due to the
connection being in an invalid state.

Similar to #433, the test ended before the connection was added
on both sides, causing the error log. Fix is similar, to use ping
to ensure the connection is added on both sides rather than ending
the test while the connection is still being added.

Fixes #439 
cc @akshayjshah

PR #433, opened 2016-07-07
Fix flaky test TestPeerRemoveClosedConnection

prashantv:
Test was failing with:
```
    <autogenerated>:10: Unexpected log: [Warn]: Failed to add connection
    to peer [{service testService} {process testService-52687} {hostPort
    127.0.0.1:52687} {connID 418} {localAddr 127.0.0.1:52687}
    {remoteAddr 127.0.0.1:54541} {remoteHostPort 127.0.0.1:54541}
    {direction inbound} {error tchannel error ErrCodeNetwork: connection
    is in an invalid state}]
```

The sequence of events was:
p1 creates a connection c1 to p2
once connection is successful, p1 closes c1

On the remote side, p2 has responded to c1 and is about to
mark it active, but the network connection gets closed before
it has been added to the peer's list. When p2 resumes adding
the connection to the peer list, it's now closed due to the underlying
network connection being closed.

We can fix this issue by issuing a ping, which will ensure that the
remote side has processed the connection and has added it to the
peer list before closing the connection.

Verified by setting RunCount = 1000 and ensuring all builds pass:
https://travis-ci.org/uber/tchannel-go/builds/143140663

Fixes #432

PR #851, opened 2022-01-26
Fix flaky TestRelayConnection

prashantv:
See #848 for context. This fixes the same issue but instead of waiting
for the inbound connection, we use Ping, which ensures the connection is
added to the channels on both the inbound/outbound side, and then
returns the pingRes.

This is similar to other fixes in the past, e.g., #440, #433.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -635,7 +635,8 @@ func TestRelayConnection(t *testing.T) {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 		defer cancel()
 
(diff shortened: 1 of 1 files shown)


prashantv:
Verified that this fixes the flake, without the fix, the test would fail within 100 runs. With the fix, I was able to run >1000 times (with & without race) without failures.

cinchurge (commented):
ah, much simpler and no more waits. thanks!

cinchurge (approved):

PR #852, opened 2022-01-27
Enable framepool checks for a subset of relay tests and TODOs for the others

cinchurge:
Ref #845 and the series of related PRs after.

Went through all tests in relay_test.go enabling framepool checks on
each test and checking whether there are frame leaks. Added TODO
items for the tests that showed leaks so we can split up the work
and address them piecemeal.


rabbbit (approved):

rabbbit:
Looks like `TestRelayModifyArg2` failed in go1.16 - this is a test you're modifying.
Given it passed in 1.15 I'm assuming it's just flaky though.

(separately, we should probably upgrade to go1.16 and 1.17?)

prashantv (approved):

PR #853, opened 2022-01-27
Fix frame leaks in handleCallReq

cinchurge:
Followup to #845

Fix various leaks in the Relayer.handleCallReq() code path and
drop TODO items for tests that pass the framepool checks.

For performance reasons, frames are pooled and are mostly
created in `readFrames()` and passed into `handle*` calls, where
a decision must be made whether or not to release that frame back
into the pool. Frames that are successfully passed to `writeFrames()`
cannot be released prematurely or the contents could be clobbered,
and frames not "sent" must be returned to the pool.

This change addresses two main categories of leaks:
- Frames not "marked" for release when they aren't sent
- Frames created outside of `readFrames()`, not sent, but also not released by the creator

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -501,8 +501,16 @@ func (c *Connection) NextMessageID() uint32 {
 }
 
 // SendSystemError sends an error frame for the given system error.
-func (c *Connection) SendSystemError(id uint32, span Span, err error) error {
+func (c *Connection) SendSystemError(id uint32, span Span, err error) (sendErr error) {
+	// Allocate an error frame to be sent over the connection. A nil is
+	// returned if the frame was successfully sent, otherwise an error is
+	// returned, and we must release the error frame back to the pool.
 	frame := c.opts.FramePool.Get()
+	defer func() {
+		if sendErr != nil {
+			c.opts.FramePool.Release(frame)
+		}
+	}()
 
 	if err := frame.write(&errorMessage{
 		id:      id,
@@ -510,7 +518,6 @@ func (c *Connection) SendSystemError(id uint32, span Span, err error) error {
 		tracing: span,
 		message: GetSystemErrorMessage(err),
 	}); err != nil {
-
 		// This shouldn't happen - it means writing the errorMessage is broken.
 		c.log.WithFields(
 			LogField{"remotePeer", c.remotePeerInfo},
--- a/relay.go
+++ b/relay.go
@@ -437,7 +437,7 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) (shouldRelease bool, _ error) {
 				call.Failed("relay-dropped")
 				call.End()
 			}
-			return _relayNoRelease, nil
+			return _relayShouldRelease, nil
 		}
 		if _, ok := err.(SystemError); !ok {
 			err = NewSystemError(ErrCodeDeclined, err.Error())
@@ -450,9 +450,9 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) (shouldRelease bool, _ error) {
 
 		// If the RelayHost returns a protocol error, close the connection.
 		if GetSystemErrorCode(err) == ErrCodeProtocol {
-			return _relayNoRelease, r.conn.close(LogField{"reason", "RelayHost returned protocol error"})
+			return _relayShouldRelease, r.conn.close(LogField{"reason", "RelayHost returned protocol error"})
 		}
-		return _relayNoRelease, nil
+		return _relayShouldRelease, nil
 	}
 
 	// Check that the current connection is in a valid state to handle a new call.
@@ -461,7 +461,7 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) (shouldRelease bool, _ error) {
 		call.End()
 		err := errConnNotActive{"incoming", state}
 		r.conn.SendSystemError(f.Header.ID, f.Span(), NewWrappedSystemError(ErrCodeDeclined, err))
-		return _relayNoRelease, err
+		return _relayShouldRelease, err
 	}
 
 	// Get a remote connection and check whether it can handle this call.
@@ -479,7 +479,7 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) (shouldRelease bool, _ error) {
 		// the current relay, we need to decrement it.
 		r.decrementPending()
 		call.End()
-		return _relayNoRelease, err
+		return _relayShouldRelease, err
 	}
 
 	origID := f.Header.ID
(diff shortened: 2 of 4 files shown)


connection.go
```diff
@@ -510,13 +510,13 @@ func (c *Connection) SendSystemError(id uint32, span Span, err error) error {
...
 		message: GetSystemErrorMessage(err),
 	}); err != nil {
-
 		// This shouldn't happen - it means writing the errorMessage is broken.
 		c.log.WithFields(
 			LogField{"remotePeer", c.remotePeerInfo},
 			LogField{"id", id},
 			ErrField(err),
 		).Warn("Couldn't create outbound frame.")
+		c.opts.FramePool.Release(frame)
```
prashantv, line 519: for each place we call `Release`, having a comment with context for why we release would be useful.
E.g.,
`// Frame allocated is never written to `sendCh`, so we manually release it to the pool.`

connection.go
```diff
@@ -542,6 +543,7 @@ func (c *Connection) SendSystemError(id uint32, span Span, err error) error {
 			LogField{"id", id},
 			ErrField(err),
 		).Warn("Couldn't send outbound frame.")
+		c.opts.FramePool.Release(frame)
 		return fmt.Errorf("failed to send error frame, buffer full")
```
prashantv, line 547: looking at the cases, it seems like the only time we call `Release(frame)` is when there's an error being returned Consider:
 * using a `defer` + a return error variable that releases frames only on errors (1 release vs 3, little simpler to follow)
 * OR moving the logic into a separate function that is given a frame, and returns whether to release

(the fewer places frames are released, the better)

relay.go
```diff
@@ -58,6 +58,9 @@ const (
 	// frame as it won't be relayed.
 	_relayNoRelease     = false
 	_relayShouldRelease = true
+
+	_frameDropped    = true
+	_frameNotDropped = false
```
prashantv, line 63: no longer used

relay.go
```diff
@@ -323,6 +328,7 @@ func (r *Relayer) Receive(f *Frame, fType frameType) (sent bool, failureReason s
 	}
 	select {
 	case r.conn.sendCh <- f:
+		sent = true
```
prashantv, line 331: Looks like this is included in #845 -- not sure if this should be rebased on top of that change?

relay.go
```diff
@@ -524,13 +530,13 @@ func (r *Relayer) handleCallReq(f *lazyCallReq) (shouldRelease bool, _ error) {
 	sent, failure := relayToDest.destination.Receive(f.Frame, requestFrame)
 	if !sent {
 		r.failRelayItem(r.outbound, origID, failure, errFrameNotSent)
-		return _relayNoRelease, nil
+		return _relayShouldRelease, nil
 	}
-	return _relayNoRelease, nil
+	return !sent, nil
```
prashantv, line 535: since we handle `!sent` above, `!sent` will always be false here, so this should just be `return false, nil` or using the constants: `return _relayNoRelease, nil`. simpler to follow

relay.go
```diff
@@ -582,13 +588,13 @@ func (r *Relayer) handleNonCallReq(f *Frame) error {
...
 		r.failRelayItem(items, originalID, failure, errFrameNotSent)
-		return nil
+		return _relayShouldRelease, nil
 	}
 
 	if finished {
 		r.finishRelayItem(items, originalID)
 	}
-	return nil
+	return !sent, nil
```
prashantv, line 597: similar to above, `!sent` is already handled, so to get here, `!sent` must be false, so simplify

prashantv (commented):
Nice work fixing all the returns to track the frame properly.

relay.go
cinchurge: yea, there was a bit of a mixup when pushing this, too many parallel PRs...

connection.go
cinchurge: added comments above the allocation since we now use defer to release, with the code directly under the alloc

connection.go
cinchurge: went with `defer` since the logic is fairly easy to follow

relay.go
cinchurge: updated. thanks!

relay_test.go
```diff
@@ -230,27 +230,30 @@ func TestRelayErrorsOnGetPeer(t *testing.T) {
...
+				SetCheckFramePooling().
+				SetRelayHost(relaytest.HostFunc(f)).
+				SetRelayOnly().
+				DisableLogVerification() // some of the test cases cause warnings.
+			testutils.WithTestServer(t, opts, func(t testing.TB, ts *testutils.TestServer) {
+				client := ts.NewClient(nil)
+				err := testutils.CallEcho(client, ts.HostPort(), "svc", nil)
+				if !assert.Error(t, err, "Call to unknown service should fail") {
+					return
+				}
```
prashantv, line 248: I think this could be `require.Error`

relay_test.go
```diff
@@ -230,27 +230,30 @@ func TestRelayErrorsOnGetPeer(t *testing.T) {
...
+
+				assert.Equal(t, tt.wantErr, err, "%v: unexpected error", tt.desc)
```
prashantv, line 250: nit: drop `tt.desc` in assert messages, since it's now part of `t.Run`

relay_test.go
```diff
@@ -466,7 +471,9 @@ func TestRelayOutgoingConnectionsEphemeral(t *testing.T) {
 
 func TestRelayHandleLocalCall(t *testing.T) {
 	// TODO: enable framepool checks
```
prashantv, line 473: can delete comment

relay_test.go
```diff
@@ -285,8 +288,10 @@ func TestErrorFrameEndsRelay(t *testing.T) {
 // Trigger a race between receiving a new call and a connection closing
 // by closing the relay while a lot of background calls are being made.
 func TestRaceCloseWithNewCall(t *testing.T) {
-	// TODO: enable framepool checks
-	opts := serviceNameOpts("s1").SetRelayOnly().DisableLogVerification()
+	opts := serviceNameOpts("s1").
+		SetRelayOnly().
+		DisableLogVerification().
+		SetCheckFramePooling()
```
prashantv, line 294: optional nit: for consistency, suggest putting `SetRelayOnly()` then `SetCheckFramePooling()` in the options next to each other (since they're mostly related in these tests) and ideally in the same order for all tests that use both.

prashantv (approved):

rabbbit (approved):
Could you update the PR description to explain what's actually changing in this diff?

Stamping because I trust Prashant's stamp, but saying more than just "follow up to X" would be helpful to understand.

cinchurge:
> Could you update the PR description to explain what's actually changing in this diff?
> 
> Stamping because I trust Prashant's stamp, but saying more than just "follow up to X" would be helpful to understand.

That's a good point. Updated the comment with more context.

PR #854, opened 2022-01-27
add rpc-service response header for tchannel inbound

kishoresenji:
Adding $rpc$-service response header with the value call.ServiceName(). This header can be used by the caller to validate that the correct service processed the request. 

This is an alternative to #843.


rabbbit:
Could you expand on the safety of this?

While from the context of tchannel-go this seems safe, we know we have services watching for this header. How can we be sure it doesn't affect the clients negatively?

cinchurge:
Please have a plan for upgrading tchannel-go for all affected services at Uber before making Uber-specific changes. Thanks!

kishoresenji:
> Please have a plan for upgrading tchannel-go for all affected services at Uber before making Uber-specific changes. Thanks!

can you please explain why this is a uber specific change?

cinchurge:
tchannel has no notion of `$rpc$` headers. see specifications here: https://tchannel.readthedocs.io/en/latest/protocol/?q=%24rpc%24&check_keywords=yes&area=default

separately, as we discussed offline and in #843, if we're making ad hoc "hacks" to the library, there must be a way for clients to opt-out as well.

kishoresenji:
> tchannel has no notion of `$rpc$` headers. see specifications here: https://tchannel.readthedocs.io/en/latest/protocol/?q=%24rpc%24&check_keywords=yes&area=default
> 
> separately, as we discussed offline and in #843, if we're making ad hoc "hacks" to the library, there must be a way for clients to opt-out as well.

Let us not discuss about #843 in this PR. Please keep the discussion about #843 in that PR.

thanks for the spec reference. Unless I missed, the spec does not say 1) it is not open for additions 2) It will never add any response headers 3) It will never add a $rpc$-service response header specifically. 

So saying adding a $rpc$-response is a hack is misjudged. It lets clients to validate whether they are talking to the right tchannel server, especially when the request can get routed to through routers. 

If a client breaks because of adding this response header, it is the implementation flaw on the client to assume that the tchannel will never add a $rpc$-service.

cinchurge:
From the specification:
> arg1~2 arg2~2 arg3~2
> The meaning of the three args depends on the systems on each end. The format of arg1, arg2, and arg3 is unspecified at the transport level. These are opaque binary blobs as far as tchannel is concerned.

at the library level, tchannel shouldn't make assumptions about the contents of these args - all transport headers are already specified in the relevant section in the request/response frames.

witriew:
I think @cinchurge's concern here is that `$rpc$-service` is irrelevant to the tchannel protocol itself.  There should be a good reason that a general use library would need to add this header, otherwise it's polluting how the library is used and adding side-effects that aren't completely understood.

Adding this to tchannel is akin to having the HTTP protocol (and all implementations of it) set a header on responses.  Additionally, making a change such as this would effectively constitute as a protocol change, which would mean all implementations of tchannel (Node, Java, etc.) should also adopt this change.

I think that this sort of change in such a general library is a bit too broad of a change - there are alternative ways to resolving failures in YARPC due to the use of this library.

kishoresenji:
> I think @cinchurge's concern here is that `$rpc$-service` is irrelevant to the tchannel protocol itself. There should be a good reason that a general use library would need to add this header, otherwise it's polluting how the library is used and adding side-effects that aren't completely understood.
> 
> Adding this to tchannel is akin to having the HTTP protocol (and all implementations of it) set a header on responses. Additionally, making a change such as this would effectively constitute as a protocol change, which would mean all implementations of tchannel (Node, Java, etc.) should also adopt this change.
> 
> I think that this sort of change in such a general library is a bit too broad and far reaching - there are alternative ways to resolving failures in YARPC due to the use of this library.

@witriew @cinchurge - Your answers to this PR reflect that you are commenting with something else as context (resolving failures in YARPC). But my intention of this PR is not that. The reason is mentioned in the description of this PR. Reiterating - I would like to have a feature from tchannel so that a client can verify that it is talking to the correct tchannel server in a tchannel service mesh. 

1) Do you agree it is a good feature to have or not? If not why not?
2) If you agree, it does not need to be $rpc$-service, this is just my thought process. As tchannel experts please suggest the appropriate hints so that a tchannel client can verify that it connected to the right tchannel server. (for eg. it can be a transport header like server name - akin to `cn` transport header)

kishoresenji:
> I think @cinchurge's concern here is that `$rpc$-service` is irrelevant to the tchannel protocol itself. There should be a good reason that a general use library would need to add this header, otherwise it's polluting how the library is used and adding side-effects that aren't completely understood.
> 
> Adding this to tchannel is akin to having the HTTP protocol (and all implementations of it) set a header on responses. Additionally, making a change such as this would effectively constitute as a protocol change, which would mean all implementations of tchannel (Node, Java, etc.) should also adopt this change.
> 
> I think that this sort of change in such a general library is a bit too broad and far reaching - there are alternative ways to resolving failures in YARPC due to the use of this library.

I do not think this accurately represents the code change that is proposed in this PR. Please correct me if I'm wrong @witriew . The change is in thrift/server.go - which is a server implementation using tchannel protocol. The change is setting an application header using tchannel protocol and so as far as the protocol is concerned the Arg2 is still opaque to the protocol. This is akin to nginx adding Server or X-Powered-By response headers. Those headers are not in the HTTP protocol. Unfortunately the thrift server implementation is in the same repo as the tchannel protocol and that is where I guess the confusion is. I hope this addresses all your and @cinchurge concerns. The change is not in the tchannel protocol.

cinchurge:
closing as we've decided to go with a different approach.

PR #855, opened 2022-01-28
Fix flaky TestIdleSweepIgnoresConnectionsWithCalls

cinchurge:
Due to a race between the idle connection sweeper and message
exchange, the test may end up getting a false negative if the
exchange shutdown happened after the last sweep has been run.

To mitigate this, make the loop run a number of times
so that we can get at least one sweep if the exchange closes
slower than usual.

### DIFF
--- a/idle_sweep_test.go
+++ b/idle_sweep_test.go
@@ -66,6 +66,16 @@ func (pl *peerStatusListener) waitForZeroConnections(t testing.TB, channels ...*
 	}
 }
 
+func (pl *peerStatusListener) waitForZeroExchanges(t testing.TB, channels ...*Channel) {
+	for i := 0; i < 10; i++ {
+		if allExchangesEmpty(channels) {
+			return
+		}
+		<-time.After(100 * time.Millisecond)
+	}
+	t.Fatalf("Some exchanges are still non-empty: %s", exchangeStatus(channels))
+}
+
 func allConnectionsClosed(channels []*Channel) bool {
 	for _, ch := range channels {
 		if numConnections(ch) != 0 {
@@ -74,6 +84,15 @@ func allConnectionsClosed(channels []*Channel) bool {
 	}
 
 	return true
+}
+
+func allExchangesEmpty(channels []*Channel) bool {
+	for _, ch := range channels {
+		if numExchanges(ch) == 0 {
+			return true
+		}
+	}
+	return false
 }
 
 func numConnections(ch *Channel) int {
@@ -88,11 +107,35 @@ func numConnections(ch *Channel) int {
 	return count
 }
 
+func numExchanges(ch *Channel) int {
+	var num int
+	rootPeers := ch.RootPeers().Copy()
+	for _, p := range rootPeers {
+		state := p.IntrospectState(&IntrospectionOptions{IncludeExchanges: true})
+		for _, c := range state.InboundConnections {
+			num += len(c.InboundExchange.Exchanges) + len(c.OutboundExchange.Exchanges)
+		}
+		for _, c := range state.OutboundConnections {
+			num += len(c.InboundExchange.Exchanges) + len(c.OutboundExchange.Exchanges)
+		}
+	}
+	return num
+}
+
 func connectionStatus(channels []*Channel) string {
 	status := make([]string, 0)
 	for _, ch := range channels {
 		status = append(status,
 			fmt.Sprintf("%s: %d open", ch.PeerInfo().ProcessName, numConnections(ch)))
+	}
+	return strings.Join(status, ", ")
+}
+
+func exchangeStatus(channels []*Channel) string {
+	status := make([]string, 0)
+	for _, ch := range channels {
+		status = append(status,
+			fmt.Sprintf("%s: %d open", ch.PeerInfo().ProcessName, numExchanges(ch)))
 	}
 	return strings.Join(status, ", ")
 }
@@ -277,6 +320,8 @@ func TestIdleSweepMisconfiguration(t *testing.T) {
 }
 
 func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
+	const numSweeps = 10
+
 	ctx, cancel := NewContext(time.Second)
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


idle_sweep_test.go
```diff
@@ -362,7 +364,14 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
 		// Unblock the call.
 		close(block)
 		<-c2CallComplete
-		check.tick()
+
```
prashantv, line 367: similar to `waitForZeroConnections`, an alternative would be to add a `waitForZeroExchanges` that uses introspection to check the number of pending message exchanges and only does `check.tick()` once the exchanges are 0?

that way it should avoid the race completely (vs mitigating the race by running more ticks)

rabbbit:
@cinchurge is still a draft still on purpose?

cinchurge:
> @cinchurge is still a draft still on purpose?

no it's not. marked as ready, thanks!

idle_sweep_test.go
cinchurge: added as suggested. thanks!

idle_sweep_test.go
```diff
@@ -66,6 +66,16 @@ func (pl *peerStatusListener) waitForZeroConnections(t testing.TB, channels ...*
 	}
 }
 
+func (pl *peerStatusListener) waitForZeroExchanges(t testing.TB, channels ...*Channel) {
+	for i := 0; i < 10; i++ {
+		if allExchangesEmpty(channels) {
+			return
+		}
+		<-time.After(100 * time.Millisecond)
```
prashantv, line 74: i'd suggest flipping the 10/100, so we run the loop 100 times, and wait 10ms in between checks. that way if the exchange clears right after we check, we won't have to wait 10ms, but 100ms.

also, since we're not using this in a channel, suggest doing `time.Sleep` here, and using `testutils.Timeout` to scale the number (could probably do `time.Sleep(testutils.Timeout(1ms))` and run that 100 times, which on CI is scaled up anyway

idle_sweep_test.go
```diff
@@ -66,6 +66,16 @@ func (pl *peerStatusListener) waitForZeroConnections(t testing.TB, channels ...*
 	}
 }
 
+func (pl *peerStatusListener) waitForZeroExchanges(t testing.TB, channels ...*Channel) {
+	for i := 0; i < 10; i++ {
```
prashantv, line 70: consider `testutils.WaitFor`
https://github.com/uber/tchannel-go/blob/v1.22.2/testutils/wait.go#L30

it handles scaling timeouts by the scale factor, and does increasing wait times between attempts, etc.

idle_sweep_test.go
```diff
@@ -88,6 +107,21 @@ func numConnections(ch *Channel) int {
...
+	var num int
+	rootPeers := ch.RootPeers().Copy()
+	for _, p := range rootPeers {
+		state := p.IntrospectState(&IntrospectionOptions{IncludeExchanges: true})
+		for _, c := range state.InboundConnections {
+			num += len(c.InboundExchange.Exchanges) + len(c.OutboundExchange.Exchanges)
+		}
+		for _, c := range state.OutboundConnections {
+			num += len(c.InboundExchange.Exchanges) + len(c.OutboundExchange.Exchanges)
+		}
```
prashantv, line 120: The `IncludeExchanges` is used to get detailed information about each exchange in `.Exchanges`, but in our case, we only need the count which is always set in `.Count`, so this can be simplified a little,
```suggestion
		state := p.IntrospectState(nil)
		for _, c := range state.InboundConnections {
			num += c.InboundExchange.Count + c.OutboundExchange.Count
		}
		for _, c := range state.OutboundConnections {
			num += c.InboundExchange.Count + c.OutboundExchange.Count
		}
```

idle_sweep_test.go
```diff
@@ -277,6 +320,8 @@ func TestIdleSweepMisconfiguration(t *testing.T) {
 }
 
 func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
+	const numSweeps = 10
```
prashantv, line 323: no longer needed, can revert

idle_sweep_test.go
```diff
@@ -286,7 +331,7 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
 	listener := newPeerStatusListener()
 	// TODO: Log filtering doesn't require the message to be seen.
 	serverOpts := testutils.NewOpts().
-		AddLogFilter("Skip closing idle Connection as it has pending calls.", 2).
+		AddLogFilter("Skip closing idle Connection as it has pending calls.", numSweeps).
```
prashantv, line 334: can revert this

idle_sweep_test.go
```diff
@@ -362,6 +407,12 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
 		// Unblock the call.
 		close(block)
 		<-c2CallComplete
+
+		// Since the idle sweep loop and message exchange run concurrently, there is
+		// a race between the idle sweep and exchange shutdown. To mitigate this,
+		// wait for the exchanges to shut down before triggering the idel sweep.
```
prashantv, line 413: nit: typo
```suggestion
		// wait for the exchanges to shut down before triggering the idle sweep.
```

idle_sweep_test.go
```diff
@@ -97,6 +131,15 @@ func connectionStatus(channels []*Channel) string {
 	return strings.Join(status, ", ")
 }
 
+func exchangeStatus(channels []*Channel) string {
```
prashantv, line 134: optional: rather than a separate method, may be simpler to have `allExchangeEmpty` return an error string if there's non-empty ones, and merge this logic in there.

prashantv (approved):

idle_sweep_test.go
cinchurge: done

idle_sweep_test.go
cinchurge: done

rabbbit (approved):

PR #856, opened 2022-01-28
Fix flaky TestLargeTimeoutsAreClamped

cinchurge:
TestLargeTimeoutsAreClamped tests that timeouts for relayed calls are "clamped"
by:
- Setting up a backend with a blocking handler
- Setting up a relay with a max timeout of `clampTTL`
- Making a call from a client with a timeout of `longTTL`
- Checking that the call returns with a timeout error within 10x `clampTTL`

This worked before, but since the migration to github actions, this test has
become very flaky, likely due to heavier load than TravisCI. The flakiness arises
from the fact that since it's a "blackbox" style test, system noise can easily
factor into the timing, causing e.g. the latency in the relay to increase significantly
under heavy load and breaching the time limit of 10x `clampTTL`.

This change bumps the timeout in the final check to `longTTL/2` to account for
latencies on the order of 100ms. The purpose here becomes checking that we timed out
before `longTTL` is reached, rather than verifying that we're clamping exactly
at `clampTTL`.

### DIFF
--- a/relay_test.go
+++ b/relay_test.go
@@ -404,8 +404,12 @@ func TestLargeTimeoutsAreClamped(t *testing.T) {
 			close(done)
 		}()
 
+		// This test is very sensitive to system noise, where a spike of latency in the relay (e.g. caused by load)
+		// is able to cause the client call to timeout, making this test prone to false positives. As such, we
+		// can't time out too close to clampTTL, but instead check that we don't time out after longTTL/2. This might
+		// be a bit generous, but should be sufficient for our purposes here.
 		select {
-		case <-time.After(testutils.Timeout(10 * clampTTL)):
+		case <-time.After(testutils.Timeout(longTTL / 2)):
 			t.Fatal("Failed to clamp timeout.")
 		case <-done:
 		}


relay_test.go
```diff
@@ -369,7 +369,7 @@ func TestTimeoutCallsThenClose(t *testing.T) {
 func TestLargeTimeoutsAreClamped(t *testing.T) {
 	// TODO: enable frame pool checks
 	const (
-		clampTTL = time.Millisecond
+		clampTTL = 10 * time.Millisecond
 		longTTL  = time.Minute
```
prashantv, line 373: (Another way to think about this test)

Currently, it's written to ensure that `longTTL` isn't used, and `clampTTL` is used (and tries to validate that the timeout happens roughly after `clampTTL`. As you called out in the summary, noise can cause the actual time taken to be significantly higher. I wouldn't be surprised to see O(100ms) in some extreme cases. For the test to run as fast as possible however, we want `clampTTL` to be a low as possible, since it's the minimum time taken for the test.

What we really care about is that `longTTL` is clamped down. We don't necessarily need to validate that it's clamped to `clampTTL` (since it's noisy), just the fact that it's clamped is a useful signal that clamping is working.

To make the test less flaky, I think we can:
 * Rely on the existing check in the server-side echo func to verify the actual timeout is less than `clampTTL` (that should never be flaky)
 * Verify that the time passed till timeout is << `longTTL`. E.g., that time taken is less than `longTTL / 2`.

That allows `clampTTL` to stay low, verifies the rewritten timeout (via the server func), and ensures that the actual timeout used on the relay is clamped down. And at the same time, we keep `clampTTL` low, and likely make the test even less flaky than the current proposed change.

relay_test.go
```diff
@@ -404,6 +404,11 @@ func TestLargeTimeoutsAreClamped(t *testing.T) {
 			close(done)
 		}()
 
+		// This test is very sensitive to system noise, where a spike of latency in the relay (e.g. caused by load)
+		// is able to cause the client call to timeout. As such, testing this with an end-to-end call is prone to
+		// false negatives.
+		// TODO: consider redesigning the test to be more "white-box", e.g. by checking the TTL field of the frame
+		// in a second relay
```
prashantv, line 411: we're effectively doing that with the server-side timeout, so not sure a second relay adds much value?

prashantv (commented):
This looks OK, but I still think it would be flaky in an extreme case where the test gets delayed by 100ms. Suggest an alternative where we're changing the `time.After` to instead use a much higher value like `longTTL / 2` which won't be flaky unless the test takes 30s+

rabbbit (approved):
Stamping if you need it in, but Prashant's description sounds probably better?

relay_test.go
cinchurge: what i had in mind was to check the clamped TTL at the frame level, i.e. what the relay sets in
```
	ttl := f.TTL()
	if ttl > r.maxTimeout {
		ttl = r.maxTimeout
		f.SetTTL(r.maxTimeout)
	}
```
supposedly if we go through a MITM relay, we should be able to inspect the updated TTL there?

cinchurge: that would probably be separate from verifying that the relay correctly times out the call though, so we'd probably still have to deal with the flakiness from and end-to-end test :(

cinchurge: ah, I didn't realize that the deadline check was already doing this, my bad. please ignore 🤦

PR #857, opened 2022-02-01
Bump timeouts to reduce flakiness in CI

cinchurge:
We've seen some tests (e.g. TestFrameRelease, some of the TestRelay* tests)
fail due to timing out, likely also due to the increased load on github actions.
Bump up the timeouts to reduce flakiness in CI.


cinchurge:
Abandoning for now

PR #858, opened 2022-02-03
Disable frame pooling checks when calculating coverage

cinchurge:
Due to the increased time when enabling frame pooling checks,
coverage calculations in CI are often timing out. Disable
said checks during coverage calculation since the impact to
coverage is minimal.


prashantv (approved):

rabbbit (approved):
Interesting.

PR #860, opened 2022-02-03
Fix frame leak due to undrained send buffer

cinchurge:
Follow up to #845

Currently in `Connection.writeFrames()`, we return immediately
upon a connection error. However, at that point there may still
be frames stuck in the send buffer (`Connection.sendCh`) which
haven't been sent due to the connection issue. Leaving it in
such a state results in a frame leak since they will never be
released.

Change `writeFrames()` to always drain the send buffer upon return
to fix this behavior.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -740,14 +740,6 @@ func (c *Connection) handleFrameNoRelay(frame *Frame) bool {
 // writeFrames is the main loop that pulls frames from the send channel and
 // writes them to the connection.
 func (c *Connection) writeFrames(_ uint32) {
-	defer func() {
-		// Remaining frames in sendCh must be drained and released,
-		// or we will leak frames
-		for len(c.sendCh) > 0 {
-			c.opts.FramePool.Release(<-c.sendCh)
-		}
-	}()
-
 	for {
 		select {
 		case f := <-c.sendCh:
@@ -865,6 +857,14 @@ func (c *Connection) checkExchanges() {
 
 func (c *Connection) close(fields ...LogField) error {
 	c.log.WithFields(fields...).Debug("Connection closing.")
+
+	// Remaining frames in sendCh must be drained and released,
+	// or we will leak frames
+	defer func() {
+		for len(c.sendCh) > 0 {
+			c.opts.FramePool.Release(<-c.sendCh)
+		}
+	}()
 
 	// Update the state which will start blocking incoming calls.
 	if err := c.withStateLock(func() error {
--- a/connection.go
+++ b/connection.go
@@ -740,6 +740,14 @@ func (c *Connection) handleFrameNoRelay(frame *Frame) bool {
 // writeFrames is the main loop that pulls frames from the send channel and
 // writes them to the connection.
 func (c *Connection) writeFrames(_ uint32) {
+	defer func() {
+		// Remaining frames in sendCh must be drained and released,
+		// or we will leak frames
+		for len(c.sendCh) > 0 {
+			c.opts.FramePool.Release(<-c.sendCh)
+		}
+	}()
+
 	for {
 		select {
 		case f := <-c.sendCh:
--- a/connection.go
+++ b/connection.go
@@ -900,14 +900,6 @@ func (c *Connection) close(fields ...LogField) error {
 // Close starts a graceful Close which will first reject incoming calls, reject outgoing calls
 // before finally marking the connection state as closed.
 func (c *Connection) Close() error {
-	// Remaining frames in sendCh must be drained and released,
-	// or we will leak frames
-	defer func() {
-		for len(c.sendCh) > 0 {
-			c.opts.FramePool.Release(<-c.sendCh)
-		}
-	}()
-
 	return c.close(LogField{"reason", "user initiated"})
 }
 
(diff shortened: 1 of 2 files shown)


connection.go
```diff
@@ -740,6 +740,14 @@ func (c *Connection) handleFrameNoRelay(frame *Frame) bool {
 // writeFrames is the main loop that pulls frames from the send channel and
 // writes them to the connection.
 func (c *Connection) writeFrames(_ uint32) {
+	defer func() {
+		// Remaining frames in sendCh must be drained and released,
+		// or we will leak frames
+		for len(c.sendCh) > 0 {
+			c.opts.FramePool.Release(<-c.sendCh)
```
rabbbit, line 747: The code below in stopCh looks similar.

Looking at it it looks like the use-case is different (here we drop the frames, there we still send them) - but could you verify?

connection.go
```diff
@@ -740,6 +740,14 @@ func (c *Connection) handleFrameNoRelay(frame *Frame) bool {
 // writeFrames is the main loop that pulls frames from the send channel and
 // writes them to the connection.
 func (c *Connection) writeFrames(_ uint32) {
+	defer func() {
```
rabbbit, line 743: would it make sense to move this defer inside `connectionError` function?

It feels like `writeFrames` is about sending data, while `connectionError` is about "closing" and cleanup?

It looks like `connectionError` is used in a bunch of other places too - so perhaps it would be more generally useful?

rabbbit: Or, perhaps, even in `closeNetwork`? 

This would nicely tie in "destruction" of the object with releasing the frames?

connection.go
```diff
@@ -858,6 +858,14 @@ func (c *Connection) checkExchanges() {
...
 	c.log.WithFields(fields...).Debug("Connection closing.")
 
+	// Remaining frames in sendCh must be drained and released,
+	// or we will leak frames
+	defer func() {
+		for len(c.sendCh) > 0 {
+			c.opts.FramePool.Release(<-c.sendCh)
+		}
+	}()
+
```
prashantv, line 868: This can act as a "best-effort" release frames, but there's many race conditions which can result in frames being written to `sendCh` after the drain here. In the past, TChannel used to `close(c.sendCh)` as part of closing a connection, but it was removed as there was many race conditions that could lead to the close happening while some other goroutine wrote to the channel, see
https://github.com/uber/tchannel-go/pull/388/commits/93ef5c112c8b321367ae52d2bd79396e2e874f31

The same race conditions that would have caused a panic when closing `sendCh` will likely cause this drain to "miss" frames from being released.

It's worth keeping in mind that a frame leak from the pool is *not* a memory leak -- when a connection is closed + removed, the GC will clean up the connection, `sencCh`, and all frames in that channel.

I think having the best-effort is OK, but the comment should indicate that it's not a complete solution.

If we want this logic, it should ideally happen as late in the lifecycle for a connection as possible, when we're as close to being confident that there's no other writers to `sendCh`. E.g., when the connection state changes to closed,
https://cs.github.com/uber/tchannel-go/blob/751f26a5d1d895affe75a1edd208d63fe3b7b109/connection.go?q=repo%3Auber%2Ftchannel-go+%22connectionStartClose%22#L847

(We could also keep the logic in `writeFrames`, and have `writeFrames` block waiting for `stopCh` before draining `sendCh`).

connection.go
```diff
@@ -858,6 +858,14 @@ func (c *Connection) checkExchanges() {
 func (c *Connection) close(fields ...LogField) error {
 	c.log.WithFields(fields...).Debug("Connection closing.")
 
+	// Remaining frames in sendCh must be drained and released,
```
prashantv, line 861: It was nice that `writeFrames` was the only place that consumed from `sendCh`. I don't think this method is the right place since it can be called from many places. More details in the next comment.

connection.go
cinchurge: good pointers! even if it's best-effort I think it's still useful since it'll help improve the signal of the tests since it won't mask leaks from other code paths.

moved the drain back into `writeFrames` and blocked on `<-stopCh`

connection.go
rabbbit: Hah, this was my idea - sorry @cinchurge. Seemed like "draining" as part of general shutdown was a reasonable thing to do. But maybe not :)

cinchurge:
bump, i think i've addressed all the issues, can you guys take another look?

relay_test.go
```diff
@@ -831,8 +834,7 @@ func TestRelayStalledConnection(t *testing.T) {
 		AddLogFilter("Dropping call due to slow connection.", 1, "sendChCapacity", "32").
 		SetSendBufferSize(32). // We want to hit the buffer size earlier, but also ensure we're only dropping once the sendCh is full.
 		SetServiceName("s1").
-		SetRelayOnly().
-		SetCheckFramePooling()
```
rabbbit, line 835: is this removal on purpose?

It's a bit surprising given the diff is expected to make the tests "less leaky", right?

connection.go
```diff
@@ -891,6 +900,14 @@ func (c *Connection) close(fields ...LogField) error {
 // Close starts a graceful Close which will first reject incoming calls, reject outgoing calls
 // before finally marking the connection state as closed.
 func (c *Connection) Close() error {
+	// Remaining frames in sendCh must be drained and released,
```
rabbbit, line 903: do we still want this, or is it a left-over from the previous iteration?

The PR description mentions only changes in `writeFrames`, and your other comment says "as discussed, prefer to keep all code that consume sendCh close together", which seems to contradict this piece of code?

mway (approved):

PR #861, opened 2022-02-04
Fix frame leaks detected in TestRelayStalledConnection

cinchurge:
Due to a race between the message exchange and relayer,
a subtle frame leak would only show up when the message
exchange was shut down (receiver closed the connection)
before the relayer stopped sending frames.

This leak is due to 2 reasons:
- When an exchange is shut down before the relayer has finished relaying,
  the frame can no longer be forwarded. But since we didn't treat that
  as an error, we also didn't release the frame in question.
- When the message exchange shuts down, it doesn't drain the receive
  buffer.

This change fixes the issues above. The test is tested with
100s of repeated runs to ensure it is no longer leaking frames
intermittently.

### DIFF
--- a/mex.go
+++ b/mex.go
@@ -37,6 +37,7 @@ var (
 	errMexSetShutdown      = errors.New("mexset has been shutdown")
 	errMexChannelFull      = NewSystemError(ErrCodeBusy, "cannot send frame to message exchange channel")
 	errUnexpectedFrameType = errors.New("unexpected frame received")
+	errUnknownMex          = errors.New("received frame for unknown message exchange")
 )
 
 const (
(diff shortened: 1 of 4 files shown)


mex.go
```diff
@@ -239,6 +240,11 @@ func (mex *messageExchange) shutdown() {
 		return
 	}
 
+	// Drain recvCh and release any pending frames
+	for len(mex.recvCh) > 0 {
+		mex.framePool.Release(<-mex.recvCh)
+	}
```
prashantv, line 246: Similar to the discussion on draining the connection `sendCh` (https://github.com/uber/tchannel-go/pull/860#discussion_r799108055), this is likely racy.

The comment on this method calls mentions this:
> // receive channel remains open, however, in case there are concurrent
> // goroutines sending to it.

That also implies that while this method is running and even after it returns, some other goroutine may write to `recvCh`. Can still keep this as a best-effort, but just want to make sure you're aware + it's called out as such.

(Since these are races, if we use a best-effort solution and then enable `CheckFramePooling` in tests can trigger the races, that could cause more flaky tests)

cinchurge:
abandoning

PR #865, opened 2022-03-30
Enable frame pool leak checks for more tests

mway:
Additionally, this also:

- Changes `testutils.updateOptsLogger` to more consistently set a logger; this fixes (what I assume was) a bug where it was possible to have a zero-value error logger that would have no filter fields and thus fail logging checks
- Closes the relay (optionally) held by `testutils.TestServer` prior to test finalization to (optimistically) help re-pool frames during teardown
- Fixes some logger returns to return pointers for compatibility with `runtime.SetFinalizer`


cinchurge:
seems like we're still failing due to flaky tests. can you split off the fixes for the flaky tests? (e.g. logger changes)

PR #868, opened 2022-04-18
Remove extraneous PeerList.exists check during PeerList.GetOrAdd

mway:
Minor change just to remove the unnecessary call. `PeerList.GetOrAdd` already calls `PeerList.Add` (which calls `PeerList.exists`), so there's no advantage to calling `exists` explicitly before the call to `GetOrAdd`. Essentially, `GetOrAdd` is functionally the same as `Add`.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -216,9 +216,7 @@ func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool) 
 
 // GetOrAdd returns a peer for the given hostPort, creating one if it doesn't yet exist.
 func (l *PeerList) GetOrAdd(hostPort string) *Peer {
-	if ps, ok := l.exists(hostPort); ok {
-		return ps.Peer
-	}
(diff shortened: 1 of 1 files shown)


peer.go
```diff
@@ -216,9 +216,6 @@ func (l *PeerList) choosePeer(prevSelected map[string]struct{}, avoidHost bool)
 
 // GetOrAdd returns a peer for the given hostPort, creating one if it doesn't yet exist.
 func (l *PeerList) GetOrAdd(hostPort string) *Peer {
-	if ps, ok := l.exists(hostPort); ok {
```
cinchurge, line 219: it's a bit strange how we ended up doing it twice. should we add a TODO to replace all `GetOrAdd` calls?

mway: yeah, same - I feel like it probably happened organically through gradual refactors. will add a TODO

cinchurge (approved):

PR #869, opened 2022-04-18
Increase test timeouts, close test server relay, misc test tweaks

mway:
- Generally speaking, increases all test timeouts by a factor of 2-3
- Ensures that the `TestServer`'s relay `Channel` is closed during server shutdown, if used
- Adds some special handling for cases like `TestRaceExchangesWithClose`:
  - added a new `testutils.CallEchoWithContext` method that optionally respects context deadlines instead of hardcoding timeouts
  - permits various shutdown states, as their transitions are nondeterministic
- Closes channels and connections in parallel during TestServer shutdown due to their (impending) synchronization
- A few other non-intrusive/cosmetic tweaks

### DIFF
--- a/close_test.go
+++ b/close_test.go
@@ -148,15 +148,14 @@ func TestRelayCloseTimeout(t *testing.T) {
 func TestRaceExchangesWithClose(t *testing.T) {
 	var wg sync.WaitGroup
 
-	ctx, cancel := NewContext(testutils.Timeout(70 * time.Millisecond))
-	defer cancel()
-
 	opts := testutils.NewOpts().DisableLogVerification()
 	testutils.WithTestServer(t, opts, func(t testing.TB, ts *testutils.TestServer) {
-		server := ts.Server()
+		var (
+			server       = ts.Server()
+			gotCall      = make(chan struct{})
+			completeCall = make(chan struct{})
+		)
 
-		gotCall := make(chan struct{})
-		completeCall := make(chan struct{})
 		testutils.RegisterFunc(server, "dummy", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
 			return &raw.Res{}, nil
 		})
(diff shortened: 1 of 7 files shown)


testutils/test_server.go
```diff
@@ -252,19 +253,38 @@ func (ts *TestServer) RegisterFunc(name string, f func(context.Context, *raw.Arg
...
-		ts.close(ch)
-		ts.verify(ch)
+
+		closeVerify.Add(1)
+		go func() {
+			defer closeVerify.Done()
+			ch.Logger().Debugf("TEST: TestServer is closing and verifying channel")
+			ts.close(ch)
+			ts.verify(ch)
+		}()
```
cinchurge, line 281: is this just meant to speed up the tests?

close_test.go
```diff
@@ -171,7 +170,23 @@ func TestRaceExchangesWithClose(t *testing.T) {
 
 		callDone := make(chan struct{})
 		go func() {
-			assert.NoError(t, testutils.CallEcho(client, ts.HostPort(), server.ServiceName(), &raw.Args{}), "Echo failed")
+			// n.b. Use a longer context here; server shutdown is inherently
+			//      nondeterministic, and now that it's blocking on channel and
+			//      connection closure, it can take anywhere from 0-2s to fully
+			//      close all of its internals.
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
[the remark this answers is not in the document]
mway: Oops, missed this one. Thanks!

testutils/echo.go
```diff
@@ -32,17 +32,58 @@ import (
...
+	return CallEchoWithContext(
+		nil,
+		src,
+		targetHostPort,
+		targetService,
+		args,
+	)
+}
+
+// CallEchoWithContext calls the "echo" endpoint from the given src to target,
```
cinchurge, line 50: do we have tests for this?

cinchurge (approved):
stamping, but we should look into filling in test coverage for testutils

close_test.go
```diff
@@ -192,22 +211,58 @@ func TestRaceExchangesWithClose(t *testing.T) {
...
 		// Now try to close the channel, it should block since there's active exchanges.
 		server.Close()
 		closed.Store(true)
-		assert.Equal(t, ChannelStartClose, ts.Server().State(), "Server should be in StartClose")
-		closed.Store(true)
 
+		// n.b. As it's shutting down, server state can be in any of the
+		//      outlined states below. It doesn't matter which specific state
+		//      it's in, as long as we're verifying that it's at least in the
+		//      process of shutting down.
```
prashantv, line 237: I don't think this is true -- we explicitly expect the channel state to be in `ChannelStartClose` because there's an inbound call that's not complete.

See:
The channel stays in ChannelStartClose till all connections are past connectionInboundClosed:
https://github.com/aiven/tchannel-go/blob/v1.22.0/channel.go#L794

Connections don't transition to `connectionInboundClosed` till the inbound exchange count is 0:
https://github.com/aiven/tchannel-go/blob/v1.22.0/connection.go#L818

The inbound exchange count is the number of pending calls, and until the below `close(completeCall)`, the server side echo handler will block.

testutils/echo.go
```diff
@@ -32,17 +32,58 @@ import (
...
+	targetService string,
+	args *raw.Args,
+) error {
 	if args == nil {
 		args = &raw.Args{}
 	}
 
-	ctx, cancel := tchannel.NewContextBuilder(Timeout(300 * time.Millisecond)).
+	timeout := time.Second
+	if ctx != nil {
```
prashantv, line 64: why support `ctx == nil` as a valid argument (which isn't common for context usages in go), when `CallEcho` can be used instead?

PR #870, opened 2022-04-18
More explicitly manage Channel and Connection lifetimes

mway:
- Tracks all goroutines
- Ensures that `<-Channel.ClosedChan()` only unblocks when all internal routines are done
- Restructures some routines (e.g. `Connection.readFrames`) for better lifetime control (i.e. better exit control)
- Prevents `Connection.close` from being called during or immediately after `Connection` creation
- Refactors `chan *Frame` drains to be tolerant of slight delays in the case of imminent `chansend`s
- Adds edge case handling for `Handler` delays due to not reading the entire call (e.g. blackhole)
- Releases frames within `messageExchange` when bound for an unknown exchange
- Minor semantic naming tweaks for `Relayer.handleLocalCallReq`

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -190,6 +190,10 @@ type Channel struct {
 	dialer              func(ctx context.Context, hostPort string) (net.Conn, error)
 	connContext         func(ctx context.Context, conn net.Conn) context.Context
 	closed              chan struct{}
+	onClosedCalled      atomic.Bool
+
+	// Tracks the lifetime of all channel-level goroutines.
+	wg sync.WaitGroup
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
@@ -386,7 +390,12 @@ func (ch *Channel) Serve(l net.Listener) error {
 	mutable.peerInfo.IsEphemeral = false
 	ch.log = ch.log.WithFields(LogField{"hostPort", mutable.peerInfo.HostPort})
 	ch.log.Info("Channel is listening.")
-	go ch.serve()
+
+	ch.wg.Add(1)
+	go func() {
+		defer ch.wg.Done()
+		ch.serve()
+	}()
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


cinchurge:
I would be slightly wary of making channel close a blocking call, as this impacts not just muttley, but all other code that consumes tchannel-go. some things I think worth checking:
- investigate if it's possible to make it blocking only for tests in tchannel-go
- if not, try upgrading the monorepo with this change and trigger a CI run so we can make sure this doesn't cause any deadlocks for consumers

prashantv:
> I would be slightly wary of making channel close a blocking call, as this impacts not just muttley, but all other code that consumes tchannel-go. some things I think worth checking:
> 
> * investigate if it's possible to make it blocking only for tests in tchannel-go
> * if not, try upgrading the monorepo with this change and trigger a CI run so we can make sure this doesn't cause any deadlocks for consumers

Just want to +1 this, I'd recommend not making large changes to Close and related lifetime functionality, it has gone through many iterations with a lot of subtle corner cases of leaks, panics, etc. This behaviour also seems like a significant behaviour change: what was previously async, now blocks, and could be considered a breaking change.

If the goal is for tests to wait for close to complete, then I'd suggest relying on [`ClosedChan`](https://pkg.go.dev/github.com/aiven/tchannel-go#Channel.ClosedChan)

mway:
> > I would be slightly wary of making channel close a blocking call, as this impacts not just muttley, but all other code that consumes tchannel-go. some things I think worth checking:
> > 
> > * investigate if it's possible to make it blocking only for tests in tchannel-go
> > * if not, try upgrading the monorepo with this change and trigger a CI run so we can make sure this doesn't cause any deadlocks for consumers
> 
> Just want to +1 this, I'd recommend not making large changes to Close and related lifetime functionality, it has gone through many iterations with a lot of subtle corner cases of leaks, panics, etc. This also seems like a significant behaviour change: what was previously async, now blocks, and could be considered a breaking change.
> 
> If the goal is for tests to wait for close to complete, then I'd suggest relying on [`ClosedChan`](https://pkg.go.dev/github.com/aiven/tchannel-go#Channel.ClosedChan)

Yeah, we discussed a bit offline and arrived at the same place, more or less. The issue is less about tests and more about (1) ambiguous object ownership and lifetimes, and (2) async lifecycle management, IMO. IIUC a lot of it stems from the need to gracefully drain inbounds, then outbounds, the latter being almost entirely arbitrary outside of timeouts.

Anyway, I don't think it'd be very productive to attempt to re-litigate historical decisions/behaviors or rock the boat too much. I'll make this change optional for testing (to ostensibly help track down leaks) and try to address the issues we're seeing in other ways.

PR #872, opened 2022-04-19
Expose net connection in InboundCall struct

jronak:
This PR exposes the underlying `net.Conn` of the request. Handlers can use the connection for extracting TLS connection state (client certs) required for AuthNZ.


channel.go
```diff
@@ -553,6 +554,14 @@ func (ch *Channel) serve() {
 				OnCloseStateChange: ch.connectionCloseStateChange,
 				OnExchangeUpdated:  ch.exchangeUpdated,
 			}
+
+			if tlsConn, ok := netConn.(*tls.Conn); ok {
+				if err := tlsConn.Handshake(); err != nil {
+					netConn.Close()
+					return
```
prashantv, line 561: Why is this needed? From the [`tls.Conn.Handshake` docs](https://pkg.go.dev/crypto/tls#Conn.Handshake):
> Most uses of this package need not call Handshake explicitly: the first Read or Write will call it automatically.

inbound.go
```diff
@@ -267,6 +270,12 @@ func (call *InboundCall) RemotePeer() PeerInfo {
 	return call.conn.RemotePeerInfo()
 }
 
+// TLSConnectionState returns the TLS connection state of the call when the
+// underlying connection is TLS.
+func (call *InboundCall) TLSConnectionState() *tls.ConnectionState {
+	return call.tlsConnectionState
+}
+
```
prashantv, line 278: Instead of tying tchannel to TLS explicitly (which seems like an unnecessary dependency), It seems better to expose just the underlying `net.Conn`, and outside of TChannel, callers can try to cast the `net.Conn` to a `*tls.Conn` and extract the connection state.

prashantv (commented):
My 2c: tchannel shouldn't depend on TLS -- it should provide access to the `net.Conn` and allow TLS to be built on top.

channel.go
jronak: Yup, you are right TLS read/write will take care of it automatically. But when there is a TLS handshake error, read/write will fail with a handshake error. I did this to catch the TLS handshake error upfront, and in the following diff, add log and metrics around it.

inbound.go
jronak: Exposing `net.Conn` seems like a good idea, though exposing it outside could be wrongly used by the callers? (primarily YARPC, so we should be good).

jronak:
Thanks for the review @prashantv! Exposing `net.Conn` seems like a good idea, and I should be to track handshake failures from a wrapped listener.

inbound.go
prashantv: Agree, it could be misused, though it should be possible to do something similar today: wrapping the listener + dialer allows callers to get access to connections and misuse them (just a little more obscure). Since the `InboundCall` isn't accessible to standard clients anyway, it'll still remain relatively obscure!

AllenLuUber (approved):
looks fine to me. Thanks for the suggestion @prashantv

prashantv (approved):
LGTM

cinchurge (approved):

PR #875, opened 2022-05-02
Vendor apache-thrift due to breaking changes with version >= 0.10.0

gandhikrishna:
Apache thrift versions >=0.10.0 contain breaking changes this is primarily due to the introduction of ctx parameter and the difference in generated goTypes.  Today tchannel-go pins to <0.10.0. This conflict causes a lot of pain for projects building with any projects that have Apache thrift dependencies with higher versions.

This PR aims to vendor Apache Thrift inside tchannel go.


abhinav (approved):
LGTM. Please also get a TChannel maintainer to review.

AllenLuUber (approved):

cinchurge (approved):

cinchurge:
the long import name feels a little verbose, but not a big deal. lgtm

PR #879, opened 2022-05-06
Add make target to allow testing without checking frame leaks

cinchurge:
Currently, `make test` bundles "vanilla" testing with frame leak
checks, which re-runs all tests with frame leak checks enabled
for some tests (currently all relay tests).

This makes routine testing pretty painful, as each test run
requires upwards of 10 minutes, slowing down development
significantly.

This change improves the situation by
- Adding a `test_vanilla` target that runs all tests without
  any leak checks
- Adding a `test_relay_frame_leaks` target which tests ONLY
  `relay_test.go` with frame leak checks enabled. Frame leak
  checks are currently enabled only in that file
- Change `make test` to run the above in succession

Except for scoping down the frame leak tests, the behavior of
`make test` is intentionally maintained since we still run all
tests in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,10 +69,18 @@ else
 endif
 
 test: clean setup check_no_test_deps $(BIN)/thrift
+	$(MAKE) test_vanilla
+	$(MAKE) test_relay_frame_leaks
+
+test_vanilla:
 	@echo Testing packages:
-	PATH=$(BIN):$$PATH go test -parallel=4 $(TEST_ARG) $(ALL_PKGS)
+	PATH=$(BIN):$$PATH DISABLE_FRAME_POOLING_CHECKS=1 go test -parallel=4 $(TEST_ARG) $(ALL_PKGS)
 	@echo Running frame pool tests
 	PATH=$(BIN):$$PATH go test -run TestFramesReleased -stressTest $(TEST_ARG)
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,11 +69,19 @@ else
 endif
 
 test: clean setup check_no_test_deps $(BIN)/thrift
+	$(MAKE) test_vanilla
+	$(MAKE) test_relay_frame_leaks
+
+test_vanilla:
 	@echo Testing packages:
```
AllenLuUber, line 76: maybe call out this as `Testing packages without relay frame leaks! To enable, use test_relay_frame_leaks`.

AllenLuUber (approved):

Makefile
cinchurge: i considered that actually, but that would also make the echo show up in `make test`, i.e. something like
```
$ make test
Testing packages without relay frame leaks! To enable, use test_relay_frame_leaks
Running frame pool tests
Testing relay frame leaks
```
which would be pretty confusing. wdyt?

AllenLuUber: rough call hehe. If you read the manual before you run, comment is better. If you don't bother to read and just run, printing out as echo is better. I'm fine with the way it is.

PR #887, opened 2022-08-19
Release unsent frames when flushing fragments

cinchurge:
Currently, when sending mutated frames via `Relayer.fragmentingSend()`
frames that aren't successfully sent (e.g. due to send buffer being
full during fragment flush) aren't returned to the frame pool and
end up piling up until either OOM or the connection gets closed.

This change adds a release call after marking the relay item as failed
to ensure that frames are properly returned to the pool.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -933,6 +933,7 @@ func (rfs *relayFragmentSender) flushFragment(wf *writableFragment) error {
 	sent, failure := rfs.frameReceiver.Receive(wf.frame, requestFrame)
 	if !sent {
 		rfs.failRelayItemFunc(rfs.outboundRelayItems, rfs.origID, failure, errFrameNotSent)
+		rfs.framePool.Release(wf.frame)
 		return nil
 	}
 	return nil
--- a/checked_frame_pool_test.go
+++ b/checked_frame_pool_test.go
@@ -1,9 +1,11 @@
 package tchannel
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/uber/tchannel-go/testutils/goroutines"
 )
 
 func TestCheckedFramePoolForTest(t *testing.T) {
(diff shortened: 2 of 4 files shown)


relay.go
```diff
@@ -933,6 +933,7 @@ func (rfs *relayFragmentSender) flushFragment(wf *writableFragment) error {
 	sent, failure := rfs.frameReceiver.Receive(wf.frame, requestFrame)
 	if !sent {
 		rfs.failRelayItemFunc(rfs.outboundRelayItems, rfs.origID, failure, errFrameNotSent)
+		rfs.framePool.Release(wf.frame)
```
lambdai, line 936: I looked at ho other frameReceiver.Receive are handled. 
My understanding is that this flushFragment maintains the invariant that wf.frame is always released. The release is either by Receive() above or here. Correct?

lambdai (commented):
+1 for the flushFragment mode here. 

However, I am confused by the expectation of calling flushFragment.

Before your change, 
1. The fragment is not released https://github.com/uber/tchannel-go/blob/dev/fragmenting_writer.go#L274-L275
1. All other flushFragment() calls in fragmenting_writer.go *WERE* always followed by a newFragment 
Both occurrences seems to put the curFragment in a to-be-released state.

However, after your change, 
The to-be-released state is no longer hold, but the some leak is fixed.

Can you help me to understand the expectation of the flushFragment()? Or, perhaps there is a trick around fragment is not released https://github.com/uber/tchannel-go/blob/dev/fragmenting_writer.go#L274-L275 ?

DheerendraRathor (approved):
LGTM.

Though I agree with yuchen that we probably need a better modelling of frame management and keeping track of ownership of frames.

DheerendraRathor:
@lambdai 
> All other flushFragment() calls in fragmenting_writer.go WERE always followed by a newFragment
Both occurrences seems to put the curFragment in a to-be-released state.

From what I understand, these calls were followed by newFragment where flushFragment is successful i.e. connection will release them once they are written to network. https://github.com/uber/tchannel-go/blob/dev/connection.go#L687

In case there are errors in flushFragment, Eric's change will take care of releasing them which were earlier contributing to leaks.

lambdai (approved):
@DheerendraRathor Fully agree in the case that no error in flushFragment.

My concern is if we introduce double released. I guess not because this fixes the leak. I think we should merge and rollout to staging

PR #890, opened 2023-03-17
Optionally send cancelled frames when context is canceled

prashantv:
Replaces https://github.com/uber/tchannel-go/pull/817

Add a new option, `SendCancelOnContextCanceled` to send cancel frames when a context is
detected to be canceled on the request sender side.

Cancellations often happen while waiting to read a response, so we can rely on the
reader goroutine waiting on the message exchange to notice the cancellation and
send the cancel message to the remote side vs adding a separate goroutine.

Older servers and relays without this change will log `Received unexpected frame`
but otherwise be unaffected by the unknown frame. However, there's no point enabling
cancellations if the remote side doesn't support them first.


mex.go
```diff
@@ -446,6 +464,31 @@ func (mexset *messageExchangeSet) forwardPeerFrame(frame *Frame) error {
 	return nil
 }
 
+func (mexset *messageExchangeSet) handleCancel(frame *Frame) {
+	if mexset.shutdown {
```
cinchurge, line 468: a lock is acquired when it's set, so a read lock should be acquired here too?

mex.go
```diff
@@ -228,22 +237,29 @@ func (mex *messageExchange) recvPeerFrameOfType(msgType messageType) (*Frame, er
...
+		onCancel(mex.msgID)
+	}
+}
+
 // shutdown shuts down the message exchange, removing it from the message
 // exchange set so  that it cannot receive more messages from the peer.  The
 // receive channel remains open, however, in case there are concurrent
 // goroutines sending to it.
-func (mex *messageExchange) shutdown() {
+func (mex *messageExchange) shutdown() bool {
```
cinchurge, line 250: the return value doesn't seem to be used anywhere

outbound.go
```diff
@@ -134,9 +134,17 @@ func (c *Connection) beginCall(ctx context.Context, serviceName, methodName stri
...
 		return nil, err
 	}
+
 	return call, nil
 }
 
+func (c *Connection) outboundCtxCancel() {
+	// outbound contexts are created by callers, can't cancel them.
+	// However, we shouldn't be trying to cancel them, so log.
+	c.log.Warn("unexpected cancel of outbound context")
```
cinchurge, line 144: worried that this might be overly verbose in practice - should it be a debug log instead?

outbound.go
```diff
@@ -134,9 +134,17 @@ func (c *Connection) beginCall(ctx context.Context, serviceName, methodName stri
 	if err := call.writeMethod([]byte(methodName)); err != nil {
 		return nil, err
 	}
+
```
cinchurge, line 137: intentional extra new line?

mex.go
prashantv: Good catch, this was from the old PR which I started off, no longer need this so reverted.

outbound.go
prashantv: Nope, removed

outbound.go
prashantv: Can change the level, though this should never happen. This is passed to the message exchange's `ctxCancel` func, and that's only used when a cancel frame is received. A cancel frame should only be received for inbound calls, which will cancel the real context. This is basically a no-op used for outbound message exchanges which aren't expected to receive a cancel.

None of the tests trigger this, and so this shouldn't happen unless a bad TChannel implementation sends an unexpected frame.

mex.go
prashantv: Shouldn't need this actually, this method should look a lot like `forwardPeerFrame` above, which doesn't have this, so reverted it.

prashantv (commented):
Thanks for the quick review!

mex.go
```diff
@@ -228,6 +237,12 @@ func (mex *messageExchange) recvPeerFrameOfType(msgType messageType) (*Frame, er
 	}
 }
 
+func (mex *messageExchange) onCtxErr(err error) {
```
cinchurge, line 240: nit: replace `err` with `_` since we don't do anything with it. maybe worth commenting that the error is reported separately so we don't need to e.g. log it here.

mex.go
```diff
@@ -143,6 +144,12 @@ func (mex *messageExchange) forwardPeerFrame(frame *Frame) error {
 	}
 }
 
+func (mex *messageExchange) handleCancel(frame *Frame) {
```
cinchurge, line 147: nit: replace `frame` with `_`

outbound.go
cinchurge: > ...and so this shouldn't happen unless a bad TChannel implementation sends an unexpected frame.

yea, mostly worried that if this is somehow triggered by an edge case there won't be a way to silence it unless we roll back. upvote for changing it to debug level.

messages.go
```diff
@@ -313,6 +314,33 @@ func (m errorMessage) Error() string {
 	return m.AsSystemError().Error()
 }
 
+type cancelMessage struct {
+	id      uint32
+	ttl     uint32 // unused.
```
cinchurge, line 319: nit: add something like `but needed for writes` or something like that

connection_test.go
```diff
@@ -491,6 +491,68 @@ func TestTimeout(t *testing.T) {
...
+}
+
+func TestCancelWithoutSendCancelOnContextCanceled(t *testing.T) {
+	testutils.WithTestServer(t, nil, func(t testing.TB, ts *testutils.TestServer) {
+		callReceived := make(chan struct{})
+		testutils.RegisterFunc(ts.Server(), "ctxWait", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
+			require.NoError(t, ctx.Err(), "context valid before cancellation")
+			close(callReceived)
+
+			<-ctx.Done()
```
cinchurge, line 533: seems like there's room for a race between the wait here for the call to timeout and when the `cancel` is called in the other goroutine. should we wait for at least 300ms before we block on the context canceling?

lambdai (commented):
quick question on this new option

connection.go
```diff
@@ -158,6 +158,10 @@ type ConnectionOptions struct {
 	// MaxCloseTime controls how long we allow a connection to complete pending
 	// calls before shutting down. Only used if it is non-zero.
 	MaxCloseTime time.Duration
+
+	// SendCancelOnContextCanceled will enable sending cancel messages
+	// when a request context is canceled before receiving a response.
```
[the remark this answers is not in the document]
cinchurge: this is a client option, but this question brings up a good point - do we need an option on the server side as well? if there were a bad client somehow spamming cancel frames it could be pretty disruptive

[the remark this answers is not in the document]
prashantv: Yep this option only affects callers -- the inbound was always happening. 

I'm assuming there's no bad clients out there spamming cancel messages (since that would cause a lot of error logs with "Received unexpected frame" which I haven't heard of).

However, it's safer to put the server-side behind an option too so added an option which controls both cancel processing on the server-side, and whether cancels are forwarded by relays.

I'm using a separate option since you generally will want to enable the server-side first, then client-side.

connection_test.go
cinchurge: gotcha, thanks for the explanation!

cinchurge (approved):

mex.go
```diff
@@ -228,6 +237,17 @@ func (mex *messageExchange) recvPeerFrameOfType(msgType messageType) (*Frame, er
 	}
 }
 
+func (mex *messageExchange) onCtxErr(err error) {
+	// On canceled contexts, we may need to send a cancel message.
+	if err != context.Canceled {
```
mway, line 242: nit: consider `errors.Is` instead for sanity

mex.go
```diff
@@ -228,6 +237,17 @@ func (mex *messageExchange) recvPeerFrameOfType(msgType messageType) (*Frame, er
...
+		return
+	}
+
+	if onCancel := mex.mexset.onCancel; onCancel != nil {
```
mway, line 246: nit: I think this condition is unnecessary - unless I'm misreading, we're unconditionally setting `onCancel` on the outbound exchange in `connection.go`, it's just the body of `onCancel` that is conditional

PR #892, opened 2023-03-20
Test against Go 1.19 and 1.20

cinchurge:
Update test spec to test against the two most recent minor releases. Fixed some gofmt changes resulting from the version update.

### DIFF
--- a/tnet/listener.go
+++ b/tnet/listener.go
@@ -41,9 +41,8 @@ func Wrap(l net.Listener) net.Listener {
 // The following can happen:
 // Goroutine 1 is running Accept, and is blocked, waiting for epoll
 // Goroutine 2 calls Close. It sees an extra reference, and so cannot destroy
-//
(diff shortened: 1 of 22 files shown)


tnet/listener.go
```diff
@@ -41,8 +41,10 @@ func Wrap(l net.Listener) net.Listener {
 // The following can happen:
 // Goroutine 1 is running Accept, and is blocked, waiting for epoll
 // Goroutine 2 calls Close. It sees an extra reference, and so cannot destroy
-//  the socket, but instead decrements a reference, marks the connection as closed
-//  and unblocks epoll.
+//
+//	the socket, but instead decrements a reference, marks the connection as closed
+//	and unblocks epoll.
+//
```
mway, line 47: should this be part of the previous line versus a preformatted block?

cinchurge: good catch. fixed.

mway (approved):
lgtm

PR #896, opened 2023-06-12
WIP - Support Span Observer run in the server-side span creation

shivam-jindal:
- The current implementation of starting a new span breaks the Span Observer feature that rely on the tracing baggage, because the lower-level function to start a span creates a span even before reading the application headers from the wire.
- This change fixes that by creating a new span in the higher-level attempt to start a span in ExtractInboundSpan(). This ensures that the observers are run when the baggage data and the procedure name are available.
- This change also ensures that there are very minimal changes to the existing code, given the legacy implementation. So, we still keep the parity with the current behavior by reusing the existing span created by the lower-level function (extractInboundSpan()) when available.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -219,44 +219,53 @@ func (c *Connection) extractInboundSpan(callReq *callReq) opentracing.Span {
 }
 
 // ExtractInboundSpan is a higher level version of extractInboundSpan().
+// A new server-side span is created by deserializing the tracing context
+// from the application headers using the standard OpenTracing API
+// supported by all tracers.
 // If the lower-level attempt to create a span from incoming request was
 // successful (e.g. when then Tracer supports Zipkin-style trace IDs),
-// then the application headers are only used to read the Baggage and add
-// it to the existing span. Otherwise, the standard OpenTracing API supported
-// by all tracers is used to deserialize the tracing context from the
-// application headers and start a new server-side span.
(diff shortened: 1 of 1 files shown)


tracing.go
```diff
@@ -219,44 +219,53 @@ func (c *Connection) extractInboundSpan(callReq *callReq) opentracing.Span {
...
+		// just to capture the Baggage and add it to the existing span. The creation
+		// of new span ensures that the Span Observers are run with all the data.
+		span.Context().ForeachBaggageItem(func(k, v string) bool {
+			existingSpan.SetBaggageItem(k, v)
+			return true
+		})
+
+		// terminate the newly created span, and ensure that it's not sampled.
+		// This span will live only within the scope of the current function.
+		ext.SamplingPriority.Set(span, 0)
```
vprithvi, line 258: Setting the sampling priority to 0, will cause the remainder of the trace to be unsampled. This is not desirable.

deepanshumehndiratta: You mean for subsequent child spans of the span for which we're setting the sampling priority to 0? The intent here is to copy over the context keys from this span and then discard it, i.e. it will not be referenced by any other span. In that case, shouldn't it not impact the sampling priority anywhere else in the trace according to this? https://github.com/jaegertracing/jaeger-client-go/blob/master/tracer.go#L297

PR #898, opened 2023-08-21
Add MPTCP support

zhiyipanuber:
This PR adds [MPTCP](https://www.multipath-tcp.org/) support. It's based on [go 1.21](https://tip.golang.org/doc/go1.21#introduction) net package.

When EnableMPTCP is passed with ChannelOptions to create a new channel, tchannel will use MPTCP instead TCP for network connection.
MPTCP requires underlying system support MPTCP.
tchannel will use normal TCP connection if EnableMPTCP is in one of following cases:
1. sets to false.
2. not passed in ChannelOptions.
3. not supported by underlying system.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -415,7 +415,7 @@ func (ch *Channel) ListenAndServe(hostPort string) error {
 		return errAlreadyListening
 	}
 
-	lc := &net.ListenConfig{}
+	lc := net.ListenConfig{}
 	lc.SetMultipathTCP(ch.enableMPTCP)
 	l, err := lc.Listen(context.Background(), "tcp", hostPort)
 	if err != nil {


channel.go
```diff
@@ -94,6 +94,12 @@ type ChannelOptions struct {
 	// This is an unstable API - breaking changes are likely.
 	RelayTimerVerification bool
 
+	// EnableMPTCP enables MPTCP for TCP network connection to increase reliability.
+	// It requires underlying operating system support MPTCP.
+	// If EnableMPTCP is false or no MPTCP support, the connection will use normal TCP.
+	// It's set to false by default.
```
AllenLuUber, line 100: Let's also call out in the comment passing your own `Dialer` means this value will be ignore on the outbound side. That is what the implementation says down below.
```
// If both a Dialer is passed as option, this value will be ignored.
```

channel.go
```diff
@@ -402,7 +414,15 @@ func (ch *Channel) ListenAndServe(hostPort string) error {
...
-	l, err := net.Listen("tcp", hostPort)
+	var l net.Listener
+	var err error
+	if ch.enableMPTCP {
+		lc := &net.ListenConfig{}
+		lc.SetMultipathTCP(true)
+		l, err = lc.Listen(context.Background(), "tcp", hostPort)
+	} else {
+		l, err = net.Listen("tcp", hostPort)
+	}
```
AllenLuUber, line 425: First, I can confirm the correctness of existing code. The [source code](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/net/dial.go;l=432):
```lang=go
func Listen(network, address string) (Listener, error) {
	var lc ListenConfig
	return lc.Listen(context.Background(), network, address)
}
```
We can simplify this block and skip the if/else check. When the option is false, it is safe to set it as well.
```lang=go
lc := &net.ListenConfig{}
lc.SetMultipathTCP(ch.enableMPTCP) 
l, err := lc.Listen(context.Background(), "tcp", hostPort)
```

channel_test.go
```diff
@@ -64,6 +64,28 @@ func TestNewChannel(t *testing.T) {
...
+			HostPort:    ephemeralHostPort,
+			IsEphemeral: true,
+			Version: PeerVersion{
+				Language:        "go",
+				LanguageVersion: strings.TrimPrefix(runtime.Version(), "go"),
+				TChannelVersion: VersionInfo,
+			},
+		},
+	}, ch.PeerInfo(), "Wrong local peer info")
+}
```
AllenLuUber, line 87: let's merge with the existing test: 
```
func TestNewChannel(t *testing.T) {
        for _, mptcp := range []bool{true, false} {
		ch, err := NewChannel("svc", &ChannelOptions{
			ProcessName: "pname",
			EnableMPTCP: mptcp,
		})
		require.NoError(t, err, "NewChannel failed")
		assert.Equal(t, LocalPeerInfo{
			ServiceName: "svc",
			PeerInfo: PeerInfo{
				ProcessName: "pname",
				HostPort:    ephemeralHostPort,
				IsEphemeral: true,
				Version: PeerVersion{
					Language:        "go",
					LanguageVersion: strings.TrimPrefix(runtime.Version(), "go"),
					TChannelVersion: VersionInfo,
				},
			},
		}, ch.PeerInfo(), "Wrong local peer info")
	}
}
```

AllenLuUber (approved):
Please get approval from another reviewer

channel.go
```diff
@@ -402,7 +415,9 @@ func (ch *Channel) ListenAndServe(hostPort string) error {
 		return errAlreadyListening
 	}
 
-	l, err := net.Listen("tcp", hostPort)
+	lc := &net.ListenConfig{}
```
dastbe, line 418: @AllenLuUber @jronak do you know how we handle configuring ingress mTLS?

dastbe (changes requested):

channel.go
```diff
@@ -402,7 +415,9 @@ func (ch *Channel) ListenAndServe(hostPort string) error {
 		return errAlreadyListening
 	}
 
-	l, err := net.Listen("tcp", hostPort)
+	lc := net.ListenConfig{}
```
jronak, line 418: Note: Yarpc does not hit this method for tchannel servers, it creates a listener of its own and passes into `Serve`. You add this under `EnableMPTCP` flag

channel.go
jronak: We create listener wrapped around tls server around it and pass the wrapped listener into tchannel `Serve`. To enable mptcp, we need to set the flags on listener in yarpc

channel.go
AllenLuUber: yea I'm aware of that. But for the sake of completeness of this feature, I think we need to do this change.

channel.go
```diff
@@ -275,8 +282,12 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
...
-	// Default to dialContext if dialer is not passed in as an option
+	// Default to dialContext or dialMPTCPContex
+	// if dialer is not passed in as an option
 	dialCtx := dialContext
+	if opts.EnableMPTCP {
+		dialCtx = dialMPTCPContext
+	}
 	if opts.Dialer != nil {
 		dialCtx = func(ctx context.Context, hostPort string) (net.Conn, error) {
 			return opts.Dialer(ctx, "tcp", hostPort)
```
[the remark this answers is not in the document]
AllenLuUber: we added a comment regarding [this](https://github.com/uber/tchannel-go/pull/898/files#r1300657045)

channel.go
zhiyipanuber: On my understanding, if a `Channel` is passed in as options to create a `ChannelTransport`, yarpc will use ListenAndServe: https://github.com/yarpc/yarpc-go/blob/dev/transport/tchannel/channel_transport.go#L168

AllenLuUber: we actually use `Transport` in yarpc: https://github.com/yarpc/yarpc-go/blob/dev/transport/tchannel/transport.go#L285. It uses `Serve()`

channel.go
dastbe: given @jronak 's comments about how we pass in listeners/dialers into tchannel explicitly for yarpc, why not just do this consistently rather than adding option here? I don't get why we would add this while not having similar capabilities for other connection-level configuration.

AllenLuUber: for services that are not on yarpc, they are calling `ListenAndServe()` and we should have this parity.

PR #899, opened 2023-08-21
Update go.mod and github test to go 1.21

zhiyipanuber:
This updates go.mod files to use Go 1.21.

It also updates the CI to use 1.21 to test.


AllenLuUber (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/uber/tchannel-go
 
-go 1.17
+go 1.21
```
rabbbit, line 3: this should be 1.20 - see https://github.com/uber-go/fx/pull/1111/files

AllenLuUber: ohh yea. Good callout. @zhiyipanuber can you patch this?

AllenLuUber: actually talked to @rabbbit and @zhiyipanuber offline, this should be fine for now as this future code won't build with version lower than 1.20.0.

PR #903, opened 2023-08-22
Emit the error code and type to the tracing span for inbound calls

deepanshumehndiratta:
### Changelog
1. Emit the error code and type to the tracing span for inbound calls

### Description
This change adds the error type and code (for system errors) to the tracing spans for inbound requests.

### DIFF
--- a/inbound.go
+++ b/inbound.go
@@ -426,7 +426,7 @@ func (response *InboundCallResponse) doneSending() {
 				// if the error is a system error, set the error code in span log
 				span.SetTag("rpc.tchannel.system_error_code", GetSystemErrorCode(response.err))
 			}
-			span.SetTag("rpc.tchannel.app_error_type", errorType)
+			span.SetTag("rpc.tchannel.error_type", errorType)
 		}
 		span.FinishWithOptions(opentracing.FinishOptions{FinishTime: now})
 	}
--- a/inbound.go
+++ b/inbound.go
@@ -424,9 +424,9 @@ func (response *InboundCallResponse) doneSending() {
 			if response.systemError {
 				errorType = systemErrorType
 				// if the error is a system error, set the error code in span log
-				span.LogKV("error-code", GetSystemErrorCode(response.err))
+				span.SetTag("rpc.tchannel.system_error_code", GetSystemErrorCode(response.err))
 			}
-			span.LogKV("error-type", errorType)
+			span.SetTag("rpc.tchannel.app_error_type", errorType)
 		}
 		span.FinishWithOptions(opentracing.FinishOptions{FinishTime: now})
 	}
--- a/inbound.go
+++ b/inbound.go
@@ -31,6 +31,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	systemErrorType = "system"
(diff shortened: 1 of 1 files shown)


inbound.go
```diff
@@ -415,6 +420,13 @@ func (response *InboundCallResponse) doneSending() {
 	if span := response.span; span != nil {
 		if response.applicationError || response.systemError {
 			ext.Error.Set(span, true)
+			errorType := appErrorType
+			if response.systemError {
+				errorType = systemErrorType
+				// if the error is a system error, set the error code in span log
+				span.LogKV("error-code", GetSystemErrorCode(response.err))
```
vprithvi, line 427: Do these system errors map to the errors described here? https://github.com/uber/tchannel/blob/master/docs/protocol.md#error-0xff

Given that these are well defined, short error codes, a span tag with the key (`rpc.tchannel.system_error_code`) would be more appropriate because it is much easier to search span tags compared to logs.

inbound.go
```diff
@@ -415,6 +420,13 @@ func (response *InboundCallResponse) doneSending() {
...
+			}
+			span.LogKV("error-type", errorType)
```
vprithvi, line 429: What values does error type contain? Also, I suggest using the key `rpc.tchannel.app_error_type` instead, and using span tags if the values are well defined and bounded.

inbound.go
ChenX1993: +1

deepanshumehndiratta: Updated in the subsequent commit

inbound.go
deepanshumehndiratta: Updated in the subsequent commit

deepanshumehndiratta: Changed to `rpc.tchannel.error_type`, since the error can be of type "application, or "system"

inbound.go
```diff
@@ -415,6 +420,13 @@ func (response *InboundCallResponse) doneSending() {
 	if span := response.span; span != nil {
 		if response.applicationError || response.systemError {
 			ext.Error.Set(span, true)
+			errorType := appErrorType
+			if response.systemError {
+				errorType = systemErrorType
+				// if the error is a system error, set the error code in span log
+				span.SetTag("rpc.tchannel.system_error_code", GetSystemErrorCode(response.err))
+			}
+			span.SetTag("rpc.tchannel.error_type", errorType)
```
vprithvi, line 429: I feel something like this could be easier to read:
(Also - we can add the error message to the span log)
```
	if span := response.span; span != nil {
		if response.systemError {
			ext.Error.Set(span, true)
			span.SetTag("rpc.tchannel.error_type", "system")
			span.SetTag("rpc.tchannel.system.error_code", GetSystemErrorCode(response.err))
			span.LogKV("rpc.tchannel.system.error_message", GetSystemErrorMessage(response.err))
		} else if response.applicationError {
			ext.Error.Set(span, true)
			span.SetTag("rpc.tchannel.error_type", "application")
		}
```

ChenX1993 (approved):

inbound.go
cinchurge: sorry for coming late to this. why weren't these changes tested?

PR #904, opened 2023-10-11
Add unit tests for emitting error code/type in tracing

deepanshumehndiratta:
Add unit tests as a followup to https://github.com/uber/tchannel-go/pull/903/files

### DIFF
--- a/inbound.go
+++ b/inbound.go
@@ -423,7 +423,7 @@ func (response *InboundCallResponse) doneSending() {
 			errorType := appErrorType
 			if response.systemError {
 				errorType = systemErrorType
-				// if the error is a system error, set the error code in span log
+				// if the error is a system error, set the error code as a span tag
 				span.SetTag("rpc.tchannel.system_error_code", GetSystemErrorCode(response.err).MetricsKey())
 			}
 			span.SetTag("rpc.tchannel.error_type", errorType)
--- a/inbound.go
+++ b/inbound.go
@@ -424,7 +424,7 @@ func (response *InboundCallResponse) doneSending() {
 			if response.systemError {
 				errorType = systemErrorType
 				// if the error is a system error, set the error code in span log
-				span.SetTag("rpc.tchannel.system_error_code", GetSystemErrorCode(response.err))
+				span.SetTag("rpc.tchannel.system_error_code", GetSystemErrorCode(response.err).MetricsKey())
 			}
 			span.SetTag("rpc.tchannel.error_type", errorType)
 		}
--- a/inbound_internal_test.go
+++ b/inbound_internal_test.go
@@ -21,7 +21,6 @@
 package tchannel
 
 import (
-	"context"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


inbound_internal_test.go
```diff
@@ -0,0 +1,98 @@
...
+func (w *statsReporter) IncCounter(name string, tags map[string]string, value int64) {
+}
+
+func (w *statsReporter) UpdateGauge(name string, tags map[string]string, value int64) {
+}
+
+func (w *statsReporter) RecordTimer(name string, tags map[string]string, d time.Duration) {
+}
+
+func TestTracingSpanApplicationError(t *testing.T) {
```
cinchurge, line 44: can you merge this test and the one below as a single table-driven test?

separately, can you add an end-to-end test for this? testing the implementation is great, but end-to-end tests are important since this library is at a very low layer of the stack.

inbound_internal_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	tracer := mocktracer.New()
+
+	_, cancel := context.WithCancel(context.TODO())
```
cinchurge, line 48: tests should have a timeout of 1s to avoid getting stuck

inbound_internal_test.go
deepanshumehndiratta: I was able to address all the comments for the implementation specific tests, however, I could not figure out a reliable way of writing an integration test to cover this flow (nothing similar already exists)

inbound_internal_test.go
```diff
@@ -0,0 +1,117 @@
...
+			applicationError:       false,
+			expectedSpanError:      true,
+			expectedSpanErrorType:  "system",
+			expectedSystemErrorKey: GetSystemErrorCode(systemError).MetricsKey(),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+
```
cinchurge, line 84: nit: redundant newline

inbound_internal_test.go
```diff
@@ -0,0 +1,117 @@
...
+			_, cancel := context.WithTimeout(context.TODO(), time.Second)
```
cinchurge, line 85: `context.Background()` would be more appropriate here since TODO is meant for undetermined situations

> TODO returns a non-nil, empty [Context](https://pkg.go.dev/context#Context). Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

inbound.go
```diff
@@ -424,7 +424,7 @@ func (response *InboundCallResponse) doneSending() {
 			if response.systemError {
 				errorType = systemErrorType
 				// if the error is a system error, set the error code in span log
```
cinchurge, line 426: do you mean "span tag" rather than "span log"?

inbound_internal_test.go
```diff
@@ -0,0 +1,115 @@
...
+				callResp = &InboundCallResponse{
+					span:          tracer.StartSpan("test"),
+					statsReporter: &statsReporter{},
+					reqResWriter: reqResWriter{
+						err: testCase.injectedError,
+					},
+					applicationError: testCase.applicationError,
+					systemError:      testCase.systemError,
+					timeNow:          time.Now,
+					cancel:           cancel,
```
cinchurge, line 98: actually it seems like you don't need an actual cancel function here since the context is never used. can you just replace this with a `func() {}` and add a comment on why it doesn't need to be a real cancel function?

cinchurge (approved):
LGTM, but still a few more nits

PR #907, opened 2023-11-15
Fix unknown error type in span tag rpc.tchannel.system_error_code

deepanshumehndiratta:
This change fixes the issue with an unknown error type being tagged in the tracing span for system errors. It also abstracts out setting the span tag to a different method which is called for both system and application errors

### DIFF
--- a/inbound_test.go
+++ b/inbound_test.go
@@ -27,6 +27,7 @@ import (
 
 	. "github.com/uber/tchannel-go"
 
+	"github.com/uber/jaeger-client-go"
 	"github.com/uber/tchannel-go/raw"
 	"github.com/uber/tchannel-go/testutils"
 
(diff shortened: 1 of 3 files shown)


cinchurge (changes requested):
the change looks ok, but can please add an end-to-end test? probably follow something like TestThriftTracingPropagation

deepanshumehndiratta:
> the change looks ok, but can please add an end-to-end test? probably follow something like TestThriftTracingPropagation

Addressed with e2e tests

inbound_test.go
```diff
@@ -35,6 +36,115 @@ import (
...
+		},
+		{
+			method:         "no-error",
+			systemErr:      false,
+			applicationErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		// We use a jaeger tracer here and not Mocktracer: because jaeger supports
```
cinchurge: use `t.Run(...` here so the results for each test case can be reported separately (also allows the defers to be run per test instead of stacking them in a loop, which is an antipattern)

inbound_test.go
```diff
@@ -35,6 +36,115 @@ import (
...
+			// Register handler that returns no error
+			ts.RegisterFunc("no-error", func(ctx context.Context, args *raw.Args) (*raw.Res, error) {
+				return &raw.Res{}, nil
+			})
+
+			// Set up a relay that can be used to terminate conns
+			// on both sides i.e. client and server
+			relayFunc := func(outgoing bool, f *Frame) *Frame {
+				return f
+			}
```
cinchurge: it looks like you don't plan to modify the frame. i don't think a relay is necessary here? you should be able to send the request directly to the test server

inbound_test.go
```diff
@@ -35,6 +36,115 @@ import (
...
+
+			clientCh := ts.NewClient(opts)
+			defer clientCh.Close()
+
+			// Make a new call, which should fail
+			_, _, resp, err := raw.Call(ctx, clientCh, relayHostPort, ts.ServiceName(), tt.method, []byte("Arg2"), []byte("Arg3"))
+
+			if tt.systemErr {
+				// Providing 'got: %q' is necessary since SystemErrCode is a type alias of byte; testify's
+				// failed test ouput would otherwise print out hex codes.
```
cinchurge:
```suggestion
				// failed test output would otherwise print out hex codes.
```

cinchurge (approved):

PR #915, opened 2024-04-22
Fix skip map value without depth limit

zhiyipanuber:
Fix skip map value without depth limit.

The [vendored apache thrift library](https://github.com/uber/tchannel-go/tree/dev/thirdparty/github.com/apache/thrift/lib/go/thrift) is vulnerable to DoS: incorrect thrift message would cause 100% CPU utilization by the processing goroutine.

The issue [was addressed in the apache/thrift](https://github.com/apache/thrift/pull/2461/commits/63e5b0cf166b458ce52a056a6b20b1b107ac548e) library, but the fix is not available in tchannel-go due to vendored old version. This PR applied this fix to tchannel-go.

Also confirmed other types of collection type (list, map key, and structs) already have the same logic.


AllenLuUber (approved):

cinchurge:
LGTM

PR #916, opened 2024-04-23
Fix protocol and add test

zhiyipanuber:
For `Skip()`, if no type is recognized in switch statement, the function returns nil before. This PR fixed it to return an error.

This PR also adds a test for protocol, which follows [apache thrift library](https://github.com/apache/thrift/blob/master/lib/go/thrift/protocol_test.go). 

The original fix already existed in [upstream](https://github.com/apache/thrift/blob/master/lib/go/thrift/protocol.go#L186)

### DIFF
--- a/thirdparty/github.com/apache/thrift/lib/go/thrift/protocol.go
+++ b/thirdparty/github.com/apache/thrift/lib/go/thrift/protocol.go
@@ -173,6 +173,8 @@ func Skip(self TProtocol, fieldType TType, maxDepth int) (err error) {
 			}
 		}
 		return self.ReadListEnd()
+	default:
+		return NewTProtocolExceptionWithType(INVALID_DATA, fmt.Errorf("Unknown data type %d", fieldType))
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


thirdparty/github.com/apache/thrift/lib/go/thrift/protocol.go
```diff
@@ -88,7 +89,7 @@ func SkipDefaultDepth(prot TProtocol, typeId TType) (err error) {
 
 // Skips over the next data element from the provided input TProtocol object.
 func Skip(self TProtocol, fieldType TType, maxDepth int) (err error) {
-	
+
```
zhiyipanuber, line 92: auto format

cinchurge:
synced offline with @AllenLuUber - looks like we're stuck with this version of apache thrift for the long haul :( can you please verify the correctness with full end-to-end testing? I'd like to reduce back and forth-ing for these changes.

zhiyipanuber:
end to end testing will be done by @vldmit

AllenLuUber (commented):
Confirmed internally that the fix works as expected

AllenLuUber (approved):

PR #918, opened 2024-04-30
fix getSysConn to work with TLS connections

DheerendraRathor:
`tls.Conn` by default doesn't implement `syscall.Conn` and hence tchannel emits `onnection does not implement SyscallConn.` log a lot. 

Since go 1.18, `tls.Conn` exposes method `NetConn()` to expose raw underlying TCP connection. This is used for getting the `syscall.Conn`.

### DIFF
--- a/connection_internal_test.go
+++ b/connection_internal_test.go
@@ -22,7 +22,9 @@ package tchannel
 
 import (
 	"bytes"
+	"crypto/tls"
 	"net"
+	"net/http/httptest"
 	"syscall"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


connection_internal_test.go
```diff
@@ -79,4 +80,20 @@ func TestGetSysConn(t *testing.T) {
 		require.NotNil(t, sysConn)
 		assert.Empty(t, loggerBuf.String(), "expected no logs on success")
 	})
+
+	t.Run("SyscallConn is successful with TLS", func(t *testing.T) {
+		loggerBuf := &bytes.Buffer{}
+		logger := NewLogger(loggerBuf)
```
cinchurge: var block

connection_internal_test.go
```diff
@@ -79,4 +80,20 @@ func TestGetSysConn(t *testing.T) {
 		require.NotNil(t, sysConn)
 		assert.Empty(t, loggerBuf.String(), "expected no logs on success")
 	})
+
+	t.Run("SyscallConn is successful with TLS", func(t *testing.T) {
```
cinchurge, line 85: although unlikely to happen in reality, can we test the case where the underlying `net.Conn` is nil as well? just to ensure it doesn't panic

DheerendraRathor: It wouldn't complain on nil net.Conn

https://go.dev/play/p/JtvaSLR-5xo

DheerendraRathor: Sorry, got the point. Added more tests.

cinchurge (approved):
LGTM

cinchurge:
can you rebase off of #919 and push again?

PR #922, opened 2024-08-26
Fix span leak for inbound & outbound

ChenX1993:
## Problem

Span leak is a issue in observability tracing where a span is started, but is not finished. The span will not be sent to tracing backend if it is not finished, hence causing incomplete traces.
```
span := tracer.StartSpan(...)
...
span.Finish()   // span leak happens when this line is not executed.
```

Internally at Uber, we fond incomplete traces from time to time due to span leak caused in tchannel-go library. In the happy path, the span is properly finished, but the span leak issue in tchannel-go happens in error path when a function early returns with error.   
E.g. 
https://github.com/uber/tchannel-go/blob/af146f5a54ac5787563a6857ed9b6ff26394f0ca/thrift/client.go#L139-L152
In above code snippet, A span is created within `c.startCall`, and is finished within `readResponse` when `reader.Close()` is executed. If there is an error occurred in `writeArgs`, then the span leak happens.

This PR is to fix the span leak issue.

## Change Summary
1. Add span component tag following the [OpenTracing semantic conventions](https://github.com/opentracing/specification/blob/master/semantic_conventions.md) to indicate the software package that created spans.
2. For inbound:
  2.1.  Add `finishSpanWithOptions` method that allows `span.Finish()` to be invoked many times in `InboundCallResponse`.
  2.2 . Use `finishSpanWithOptions` in `dispatchInbound` to always try to finish a span after the handler is called.
3. For outbound:
  3.1.  Add `finishSpanWithOptions` in `OutboundCallResponse` similar to inbound case
  3.2.  Use `finishSpanWithOptions` json, thrift and raw clients.

## Next
At Uber, YARPC ([yarpc-go](https://github.com/yarpc/yarpc-go)) is often used as a umbrella RPC framework with gRPC, TChannel and HTTP as transports.  When YARPC is used with TChannel, the span leak issue does also exist. 

With this PR, the span leak in inbound is expected to be fixed for YARPC as well, but for outbound, we need to add `defer call.Response().Done` [here](https://github.com/yarpc/yarpc-go/blob/1597f032793ab9575e6e9be5c2308264616b4a0b/transport/tchannel/outbound.go#L146) in yarpc-go after this PR is merged.

### DIFF
--- a/inbound.go
+++ b/inbound.go
@@ -27,7 +27,6 @@ import (
 	"time"
 
 	"github.com/opentracing/opentracing-go"
-	"github.com/opentracing/opentracing-go/ext"
 	"golang.org/x/net/context"
 )
 
@@ -173,6 +172,8 @@ func (call *InboundCall) createStatsTags(connectionTags map[string]string) {
 
 // dispatchInbound ispatches an inbound call to the appropriate handler
 func (c *Connection) dispatchInbound(_ uint32, _ uint32, call *InboundCall, frame *Frame) {
+	defer call.Response().finishSpanWithOptions(opentracing.FinishOptions{})
+
 	if call.log.Enabled(LogLevelDebug) {
 		call.log.Debugf("Received incoming call for %s from %s", call.ServiceName(), c.remotePeerInfo)
 	}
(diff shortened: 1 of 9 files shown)


ChenX1993:
Closed

PR #923, opened 2024-10-06
Add component tag to tracing spans

ChenX1993:
This PR is making a minor change to add span component tag to tracing spans. According to the [OpneTracing sematic conventions](https://github.com/opentracing/specification/blob/master/semantic_conventions.md), the component tag indicates the library that created the spans. At Uber, It is quite useful for the tracing team to understand the span creation distribution among libraries in order to optimize the operational cost.
> The software package, framework, library, or module that generated the associated Span. E.g., "grpc", "django", "JDBI".

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -38,6 +38,8 @@ import (
 // NB: the string value is what's actually shared between implementations
 const zipkinSpanFormat = "zipkin-span-format"
 
+const componentName = "tchannel-go"
+
 // Span is an internal representation of Zipkin-compatible OpenTracing Span.
 // It is used as OpenTracing inject/extract Carrier with ZipkinSpanFormat.
 type Span struct {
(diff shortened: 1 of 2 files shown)


tracing.go
```diff
@@ -38,6 +38,8 @@ import (
 // NB: the string value is what's actually shared between implementations
 const zipkinSpanFormat = "zipkin-span-format"
 
+const componentName = "tchannel-go"
```
ChenX1993, line 41: Add `-go` suffix since we have tchannel library for java as well.

cinchurge:
how was this tested end-to-end?

ChenX1993:
> how was this tested end-to-end?

If within the tchannel-go, there is already a tracing testing, and I updated the test with the new tag verification in this PR: 
https://github.com/uber/tchannel-go/blob/af146f5a54ac5787563a6857ed9b6ff26394f0ca/tracing_test.go#L60

If internally in monorepo, once I have the diff to bump the tchannel-go version, I will use our testing service to test end-to-end in staging env - checking the tag from jaeger UI.

cinchurge (approved):
verified that the tag is present in the trace. thanks for testing it end-to-end!

Issue #914, opened 2024-03-28
Fails to compile on FreeBSD 14: missing function getSendQueueLen


hxw:
## Quick Fix

There is a `sockio_unix.go` which shows support for many OSs, but it references a **getSendQueueLen** function.
Currently this function is only defined for Linux (in sockio_linux.go).  A quick fix for FreeBSD is to:

```
cp sockio_linux.go sockio_bsd.go
```

Next patch the sockio_bsd.go with the following diff:

 ```
--- sockio_linux.go	2024-03-28 14:42:39.901422000 +0800
+++ sockio_freebsd.go	2023-04-13 11:42:33.888702000 +0800
@@ -18,14 +18,13 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-//go:build linux
-// +build linux
+//go:build aix || darwin || dragonfly || freebsd || netbsd || openbsd || solaris
+// +build aix || darwin || dragonfly || freebsd || netbsd || openbsd || solaris
 
 package tchannel
 
 import "golang.org/x/sys/unix"
 
 func getSendQueueLen(fd uintptr) (int, error) {
-	// https://linux.die.net/man/7/tcp
-	return unix.IoctlGetInt(int(fd), unix.SIOCOUTQ)
+	return unix.IoctlGetInt(int(fd), unix.TIOCOUTQ)
 }
```

## Better Alternative (perhaps)

Alternatively the function using `unix.TIOCOUTQ` could just be added to `sockio_unix.go` and the `sockio_linux.go` removed. From inspection of the files in ` go120/src/cmd/vendor/golang.org/x/sys/unix/` - All the OSs supported by `sockio_unix.go` have the `TIOCOUTQ` IOCTL and in the various Linux files this has exactly the same as the `SIOCOUTQ` value.

moonetic:
The fix adds a compile issue on MacOS, because of multiple declarations of getSendQueueLen for darwin(in sockio_bsd and sockio_darwin). I opened another PR with a fix by removing darwin tag from sockio_bsd.
Here's the PR: https://github.com/uber/tchannel-go/pull/926

@ibarrajo, can you please assign a reviewer to this PR too?

PR #925, opened 2024-10-31
Fix compile issue on FreeBSD 14

moonetic:
This PR fixes compilation fail on FreeBSD 14, described on [issue by hxw](https://github.com/uber/tchannel-go/issues/914) The problem is, that there's no getSendQueueLen function defined for bsd platform, which causes compile failure. The fix adds new, sockio_bsd file with corresponding function implementation and build tags.


cinchurge (approved):
LGTM

PR #926, opened 2024-12-19
Fix compile issue on MacOS

moonetic:
This PR fixes compilation fail on MacOS(darwin), that occurrs after fixing the FreeBSD 14 compile issue described on [issue by hxw](https://github.com/uber/tchannel-go/issues/914) The problem is, that now there are two getSendQueueLen functions defined for darwin platform, which causes compile failure. The fix removes darwin build tag from sockio_bsd file.

### DIFF
--- a/sockio_bsd.go
+++ b/sockio_bsd.go
@@ -18,8 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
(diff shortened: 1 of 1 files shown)


moonetic:
@cinchurge can you please review this? The compile issue on MacOS is pretty serious, I think))

irfn:
In absence of the Merge, Is there any workaround for this?

kklimov-axcient:
> In absence of the Merge, Is there any workaround for this?

@irfn Only your own fork.

maryis (approved):

PR #928, opened 2025-10-21
Fix serversocket addr race condition

AtharvaChiplunkar12:
Problem
The Addr() method in server_socket.go was accessing the p.listener field without mutex protection, which could cause data races when multiple goroutines access it concurrently (e.g., one goroutine calling Addr() while another calls Close(), Open(), or Listen()).

Solution
Added RLock()/RUnlock() protection around listener access for thread-safety
Changed direct nil check to use IsListening() method for consistency with the codebase
Follows the same locking pattern used in other methods (Accept(), Close(), Listen(), Open())

Sync with upstream fix https://github.com/apache/thrift/commit/2620a12bfa92dfb808fb9668b3d84a2c56526fc3

Also, this PR includes all the necessary upstream fixes to ensure the file is up-to-date https://github.com/apache/thrift/blob/master/lib/go/thrift/server_socket.go

### DIFF
--- a/thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
+++ b/thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
@@ -69,15 +69,12 @@ func (p *TServerSocket) Listen() error {
 func (p *TServerSocket) Accept() (TTransport, error) {
 	p.mu.RLock()
 	interrupted := p.interrupted
+	listener := p.listener
 	p.mu.RUnlock()
 
 	if interrupted {
 		return nil, errTransportInterrupted
 	}
-
-	p.mu.Lock()
-	listener := p.listener
-	p.mu.Unlock()
 	if listener == nil {
 		return nil, NewTTransportException(NOT_OPEN, "No underlying server socket")
 	}
--- a/thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
+++ b/thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
@@ -98,7 +98,7 @@ func (p *TServerSocket) Open() error {
 func (p *TServerSocket) Addr() net.Addr {
 	p.mu.RLock()
 	defer p.mu.RUnlock()
-	if p.IsListening() != nil {
+	if p.IsListening() {
 		return p.listener.Addr()
 	}
 	return p.addr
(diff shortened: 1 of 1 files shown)


pulkit4tech (approved):
Thanks! LGTM!

thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
```diff
@@ -67,10 +74,14 @@ func (p *TServerSocket) Accept() (TTransport, error) {
 	if interrupted {
 		return nil, errTransportInterrupted
 	}
-	if p.listener == nil {
+
+	p.mu.Lock()
+	listener := p.listener
+	p.mu.Unlock()
```
oneporter: why take the write lock?  Can we move line 79 to 71 and only take the read lock once?

oneporter (changes requested):
I guess this was already fixed, can we pull the fixed version? https://github.com/apache/thrift/pull/3220/files

thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
AtharvaChiplunkar12: Yes, good point it will take a read lock. Since these changes were directly pulled from upstream, I assume this was overlooked there

thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
```diff
@@ -62,15 +69,16 @@ func (p *TServerSocket) Listen() error {
 func (p *TServerSocket) Accept() (TTransport, error) {
 	p.mu.RLock()
 	interrupted := p.interrupted
+	listener := p.listener
```
pnndesh, line 72: In https://github.com/apache/thrift/blob/master/lib/go/thrift/server_socket.go, the listener is accessed under exclusive lock. Let's do the same here.

thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
```diff
@@ -26,12 +26,12 @@ import (
 )
 
 type TServerSocket struct {
```
pnndesh, line 28: Let's add more tests in server_socket_test.go and run them with race-detection flag

thirdparty/github.com/apache/thrift/lib/go/thrift/server_socket.go
AtharvaChiplunkar12: it was a bug in upstream. fixed it on upstream https://github.com/apache/thrift/commit/e7ab34eccb272584441f5462104d87f08b217ea7

PR #935, opened 2026-08-04
Fix and modernize CI

justinhwang:
Bump actions/checkout (v2 -> v7) and actions/setup-go (v5 -> v7), which drops the deprecated actions/cache@v1 that was auto-failing every run. Run checkout before setup-go and set cache-dependency-path so setup-go module caching finds go.sum (this repo checks out into $GOPATH/src). Remove the dead glide actions/cache step (no glide.lock or vendor dir in the repo).

Exclude the stdlib internal/synctest package from check_no_test_deps: Go 1.25+ pulls it into production deps via sync, and its name trips the test/mock grep -- the same false positive already handled for internal/testlog.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -28,23 +28,18 @@ jobs:
           LINT: "yes"
 
     steps:
-    - name: Setup Go
-      uses: actions/setup-go@v5
-      with:
-        go-version: ${{ matrix.go }}
-
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v7
       with:
         path: ${{ env.GOPATH }}/src/github.com/${{ github.repository }}
 
-    - name: Load cache
-      uses: actions/cache@v1
+    - name: Setup Go
+      uses: actions/setup-go@v7
       with:
-        path: ~/.glide/cache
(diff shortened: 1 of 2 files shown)


idle_sweep_test.go
```diff
@@ -351,6 +348,11 @@ func TestIdleSweepIgnoresConnectionsWithCalls(t *testing.T) {
 		// Client 1 will just ping, so we create a connection that should be closed.
 		c1 := ts.NewClient(clientOpts)
 		require.True(t, testutils.WaitFor(10*time.Second, func() bool {
+			// Use a fresh, timeout-scaled context per attempt. A single short
+			// context shared across this retry loop would expire partway through
+			// and fail every remaining ping on a slow (e.g. CI) host.
+			ctx, cancel := NewContext(testutils.Timeout(time.Second))
+			defer cancel()
```
rabbbit: I would have to spend too much time to understand if this is by design or not.

Given:
(1) it's orthogonal to your original change
(2) this is going away

wdyt about just removing the test changes? We can ignore test failures for a while (or retry)

justinhwang: Yeah removed the go test file changes, now it's strictly updating the gh actions

rabbbit (approved):

